PR #35, opened 2024-07-28
fix: calculate du on different batch

uniartisan:
du += du_i.sum(0)


yzhangcs:
@uniartisan Thanks for your contributions!

PR #37, opened 2024-07-29
fix: enhance state gradient when bf16

uniartisan:
FIX: https://github.com/sustcsonglin/flash-linear-attention/issues/36


uniartisan:
Greetings maintainers, thank you for your work.

In this PR, I have fixed the issue of inconsistent gradients in the state of chunkrwkv (the original implementation had an absolute deviation that was 200-3000 times the original). The new implementation has an error of 1%-3%, which is simply achieved by introducing fp32 at a critical location.

After benchmarking, the performance is at the same level, with no significant decrease or improvement.

I have also updated the algorithm for measuring gradient error, and now use RMS to evaluate it. There is still a problem of large gradient errors in w and u, but considering the approximate implementation of chunk, this may be acceptable (for example, in my use case, I only need to backpropagate through the state, i.e. h).

PR #39, opened 2024-08-05
Add implementations of Mamba 2 into FLA

DanFosing:
Modified version of https://github.com/huggingface/transformers/tree/add_codestral_mamba2/src/transformers/models/mamba2 (https://github.com/huggingface/transformers/pull/32080) to work with FLA.
I haven't tested if it works yet, but I'm pretty sure it will work. It can probably be made a bit faster by implementing gated RMSNorm utilizing SiLU.

### DIFF
--- a/fla/models/__init__.py
+++ b/fla/models/__init__.py
@@ -11,6 +11,7 @@ from fla.models.linear_attn import (LinearAttentionConfig,
                                     LinearAttentionForCausalLM,
                                     LinearAttentionModel)
 from fla.models.mamba import MambaConfig, MambaForCausalLM, MambaModel
+from fla.models.mamba2 import Mamba2Config, Mamba2ForCausalLM, Mamba2Model
 from fla.models.retnet import RetNetConfig, RetNetForCausalLM, RetNetModel
 from fla.models.rwkv6 import RWKV6Config, RWKV6ForCausalLM, RWKV6Model
 from fla.models.samba import SambaConfig, SambaForCausalLM, SambaModel
@@ -26,6 +27,7 @@ __all__ = [
     'HGRN2Config', 'HGRN2ForCausalLM', 'HGRN2Model',
     'LinearAttentionConfig', 'LinearAttentionForCausalLM', 'LinearAttentionModel',
     'MambaConfig', 'MambaForCausalLM', 'MambaModel',
+    'Mamba2Config', 'Mamba2ForCausalLM', 'Mamba2Model',
     'RetNetConfig', 'RetNetForCausalLM', 'RetNetModel',
     'RWKV6Config', 'RWKV6ForCausalLM', 'RWKV6Model',
     'SambaConfig', 'SambaForCausalLM', 'SambaModel',
--- a/fla/models/mamba2/__init__.py
+++ b/fla/models/mamba2/__init__.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+
+from transformers import AutoConfig, AutoModel, AutoModelForCausalLM
+
+from fla.models.mamba2.configuration_mamba2 import Mamba2Config
+from fla.models.mamba2.modeling_mamba2 import (Mamba2Block, Mamba2ForCausalLM,
+                                             Mamba2Model)
+
+AutoConfig.register(Mamba2Config.model_type, Mamba2Config, True)
+AutoModel.register(Mamba2Config, Mamba2Model, True)
+AutoModelForCausalLM.register(Mamba2Config, Mamba2ForCausalLM, True)
+
+
+__all__ = ['Mamba2Config', 'Mamba2ForCausalLM', 'Mamba2Model', 'Mamba2Block']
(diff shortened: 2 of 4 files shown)


yzhangcs:
@DanFosing Great job! I'll make some tests soon. Thank you for the contributions.

yzhangcs:
Could you add `__init__.py` file like https://github.com/sustcsonglin/flash-linear-attention/blob/main/fla/models/mamba/__init__.py to make the model recongnizable by fla and register it in 
https://github.com/sustcsonglin/flash-linear-attention/blob/main/fla/models/__init__.py

DanFosing:
I think it should work with fla now

DanFosing:
Unfortunately because it's still an ongoing pull request into transformers package, there is a possibility that it may not fully work in some specific cases, that's why it requires some testing.

yzhangcs:
@DanFosing Hi, looks like there are still some errors 
```py
  File "flash-linear-attention/fla/models/mamba2/modeling_mamba2.py", line 607, in forward
    return self.cuda_kernels_forward(hidden_states, cache_params, cache_position, attention_mask)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "flash-linear-attention/fla/models/mamba2/modeling_mamba2.py", line 333, in cuda_kernels_forward
    rmsnorm_weight=self.norm.weight,
                   ^^^^^^^^^^^^^^^^
  File "anaconda3/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1709, in __getattr__
    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
AttributeError: 'GatedRMSNorm' object has no attribute 'weight'
```

yzhangcs:
@DanFosing You can check your impls via running 
```py
python -m benchmarks.benchmark_training_throughput --name mamba2
```

yzhangcs:
You can refer to https://github.com/sustcsonglin/flash-linear-attention/blob/main/fla/layers/gla.py#L219-L228 for our impls of gate norm. I think there is no need to further wrap it with another `GatedNorm`.

yzhangcs:
@DanFosing Appreciate your hard work and quick response.

DanFosing:
Are you sure it works? When I tried it on kaggle I got assertion errors but it often has some weird problems so it may be an issue with kaggle (I can't test it on my pc right now)

DanFosing:
Turns out it was just a problem with T4 not supporting some things from causal_conv1d package, there is the same issue with mamba 1 and 2 from mamba-ssm. I don't remember what it was but there was some workaround that fixed it if you turned off some low memory mode in pytorch or something. Btw torch.compile support was added to mamba-1 but it requires some changes in mambacache and mamba modeling, I will try to implement it when I have some time.

learning-chip:
One question on this PR:

From the [FLA paper](https://arxiv.org/abs/2312.06635), both mamba-1 and mamba-2 can be written as Gated linear attention formulation:

fla_table

So, the SSM part of Mamba-2 (excluding conv1d, normalization, ...) should permit the same "Chunkwise Parallel Form" as advocated by the FLA paper, no?

chunkwise

Then, the `modeling_mamba2.py` implementation here shouldn't need to import `mamba_chunk_scan_combined` from original `mamba_ssm` repository, right? It can use a custom kernel similar to [`ops/gla/chunk_fuse.py`](https://github.com/sustcsonglin/flash-linear-attention/blob/main/fla/ops/gla/chunk_fuse.py) in this repo. 

Otherwise, the current just uses custom `RMSNorm`/`FusedRMSNormSwishGate` kernels, while the SSM part is no different from original [`mamba_ssm` repository](https://github.com/state-spaces/mamba). Then the performance will remain largely the same as original paper, and you cannot tell whether FLA formulation is faster...

learning-chip:
The mamba-2 blog [did mention](https://goombalab.github.io/blog/2024/mamba2-part3-algorithm/#special-cases) that FLA chunkwise parallel "turns out to be essentially equivalent to the SSD algorithm specialized to a restricted case"

special_case

Will the FLA formulation be just identical to the `mamba_chunk_scan_combined` [in original mamba-2 code](https://github.com/state-spaces/mamba/blob/main/mamba_ssm/ops/triton/ssd_combined.py)?

DanFosing:
Indeed I think it can be made faster if a kernel similar to chunk_fuse.py is used (maybe it would be possible to just modify gla one as both GLA and Mamba-2 are extremely similar to each other). Unfortunately I'm not familar with triton so I can't really do it, I think it would be best if you made an issue or something, unless some of FLA devs replies there. Current implementation is a lot like mamba-1 implementation, both are mostly like original mamba-ssm ones, with just some minor speed up thanks to FusedCrossEntropyLoss and custom RMSNorm and those are compatible with huggingface transformers.

DanFosing:
If you take a look at this paper: https://arxiv.org/pdf/2406.06484 you can also see that in terms of recurrence and memory read-out mamba-2 is very similar to RetNet and Linear Attention:
image

learning-chip:
> I think it would be best if you made an issue or something, unless some of FLA devs replies there.

Will look into it!

yzhangcs:
@learning-chip Hi, you may refer to [simple gla](https://github.com/sustcsonglin/flash-linear-attention/tree/main/fla/ops/simple_gla) by @sustcsonglin which provides data-dependent decay upon RetNet.

PR #42, opened 2024-08-08
[DeltaNet] Adds beta as a vector option 

hypnopump:
Adds option to have beta as a headwise vector instead of simply a headwise scalar (for the recurrent kernel)


yzhangcs:
@hypnopump Hi, thank you for your contributions for improving deltanet, could you add some regression tests for the beta vector in https://github.com/sustcsonglin/flash-linear-attention/blob/main/tests/ops/test_delta.py

yzhangcs:
@hypnopump Thanks~

PR #45, opened 2024-08-13
[Attn] fix negative value of seqlen offset during sft

ChaosCodes:
Disable seqlen offset bias during training (pretraining & sft)


yzhangcs:
Thank you for the contributions :-)

PR #48, opened 2024-08-16
[RWKV6] fix backward if h0 not passed

hypnopump:
Currently, if `initial_state` is not passed as 0s, it breaks: 

```python
/fla/ops/rwkv6/chunk.py", line 753, in backward
    dh, dh0 = dh.to(q), dh0.to(q)
AttributeError: 'NoneType' object has no attribute 'to'
```

Fix it by only doing tensor ops if the `initial_state` is passed


yzhangcs:
Thanks~

PR #54, opened 2024-08-25
Update amp custom_fwd, custom_bwd usage for torch 2.4.0 compatibility

mirceamironenco:
In torch 2.4.0 the autocast APIs were unified under torch.amp. Usage of torch.{device}.amp will produce a deprecation warning: FutureWarning: `torch.cuda.amp.custom_fwd(args...)` is deprecated. Please use `torch.amp.custom_fwd(args..., device_type='cuda')` instead.

This PR gates `custom_fwd/custom_bwd` by version, so that < 2.4 torch is still usable. Since these ops are used for a lot of primitives I've moved them to `fla.utils` and renamed them for clarity. This should still allow future support for device types other than cuda.

### DIFF
--- a/fla/ops/abc/recurrent_fuse.py
+++ b/fla/ops/abc/recurrent_fuse.py
@@ -7,9 +7,8 @@ from typing import Optional, Tuple
 import torch
 import triton
 import triton.language as tl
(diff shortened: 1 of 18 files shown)


yzhangcs:
Thank you for the contributions.

PR #55, opened 2024-09-02
Fix syntax error

JulienSiems:
Hey Songlin, 

Thank you for this great work! We are using it quite a bit in our research :) 
When installing the repo, I ran into a syntax error which appears to have sneaked in during the last commit. 

Julien


yzhangcs:
Thanks for the fix!

PR #56, opened 2024-09-03
Add `__init__.py` in `fla/ops/common` for automatic package discovery

zhixuan-lin:
This PR adds an `__init__.py` file to the `fla/ops/common` directory to ensure that it can be recognized by `find_packages` (in `setup.py`).


yzhangcs:
Thank you.

PR #57, opened 2024-09-05
[`Mamba2`] Post Merge Fixes - `norm_before_gate` and generation with `inputs_embeds` 

vasqu:
Since the implementation seems like a mirror to the one of the transformers repo, I thought I'd share some fixes post-merge that have been applied:
1. Generation when passing `inputs_embeds` has been broken, ref. https://github.com/huggingface/transformers/pull/32694
2. Incorrect usage of the `norm_before_gate` flag which causes misaligned training when using the completely fused kernel, ref. https://github.com/huggingface/transformers/pull/32686

Cool repo btw :)


yzhangcs:
Thank you :D

PR #59, opened 2024-09-10
Correctly compute `max_seqlen` when `max_position_embeddings` is `None`

zhixuan-lin:
The original code

```python
q, k = self.rotary(q, k, seqlen_offset, max(max_seqlen, self.max_position_embeddings))
```
does not work when `self.max_position_embeddings` is `None`. This PR fixes it.


yzhangcs:
Thank you.

PR #65, opened 2024-09-19
Add fine-grained warning category for easier supression

mirceamironenco:
Some packages like flash-attn (and causal-conv) are used only for specific primitives, however due to the import strategy in `fla.__init__.py` a user without flash-attn installed will see a warning when using other modules. I've added a category to the warnings emitted in these cases so one can more safely ignore it when importing other layers, i.e. one can use:

```py
import warnings
with warnings.catch_warnings(action="ignore", category=ImportWarning):
    from fla.ops.gla import chunk_gla
    from fla.ops.linear_attn import chunk_linear_attn
```

### DIFF
--- a/fla/layers/attn.py
+++ b/fla/layers/attn.py
@@ -21,7 +21,8 @@ try:
     from flash_attn.bert_padding import (index_first_axis, pad_input,
                                          unpad_input)
 except ImportError:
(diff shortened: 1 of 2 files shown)


yzhangcs (approved):

yzhangcs:
Thx~

PR #78, opened 2024-11-08
Fix RWKV6 Cache Problems

WorldEditors:
Problems:

When the weight to forgetting gate be surfficiently large (>20), backward gives gradient = NAN (forward is still OK)

Solutions:

Clip (might have the risk of zero gradient) 
Add soft bound (with soft sign), might impact the exisiting model


yzhangcs:
@WorldEditors Hi, thank you for the contributions. Unfortunately, I won't be able to accept these commits as it seems that the new code is not compatible with our existing pretrained ckpts, and I'm afraid that the NaNs could stem from some out of boundary issues. I will be working on this recently.Please stay tuned!

yzhangcs:
I really appreciate your work on fixing the cache bugs. Could you please create a separate PR for those fixes? 
Since it's a different subject from this PR, I'd be happy to accept it separately.

WorldEditors:
> I really appreciate your work on fixing the cache bugs. Could you please create a separate PR for those fixes? Since it's a different subject from this PR, I'd be happy to accept it separately.

Remove the softsign bound

yzhangcs:
Thank you for the contributions.

PR #84, opened 2024-11-22
[`Mamba2`] Fix slow path

vasqu:
The slow path has had one wrong contraction - see https://github.com/huggingface/transformers/issues/34817 for details

This PR includes:
- The fix using one permutation less
- Simplified intra-chunk states (1 permutation vs 4)
- Added comments back to follow the original ssd minimal 

Verified myself in my checks over [here](https://github.com/vasqu/mamba2-torch/blob/main/tests/TestSSDMinimal.py) - also tested against the hf versions locally


yzhangcs:
Thank you.

Issue #90, opened 2024-12-09
[Bug]: Bunches of Issues in Mamba and Mamba2


WorldEditors:
### Describe the bug

1. Mamba and Mamba2 There is obvious difference between inference as a whole and chunk-wise inference.

3. In Mamba2, Only Hidden_Size=2048 will work, others will trigger error
4. What's the usage of cache_position ? Quite confusing

### Steps to reproduce the bug

```python
import torch
import fla
print(fla)

from fla.models.mamba2.modeling_mamba2 import Mamba2Mixer, Mamba2Cache
from fla.models.mamba2.configuration_mamba2 import Mamba2Config

from fla.models.mamba.modeling_mamba import MambaMixer, MambaCache
from fla.models.mamba.configuration_mamba import MambaConfig

bsz = 1
hidden_size = 2048  # Will only be ok if hidden_size=2048, others will trigger error
seq_len = 1024
seg_len = 128
lr = 1.0e-3
use_cache = True   # Will be OK if it is False

seg_num = (seq_len - 1) // seg_len + 1

#config = MambaConfig(hidden_size=hidden_size)
config = Mamba2Config(hidden_size=hidden_size)

#encoder = MambaMixer(
#    config,
#    layer_idx=0)
encoder = Mamba2Mixer(
    config,
    layer_idx=0)

encoder = encoder.to('cuda')
inputs = torch.randn(bsz, seq_len, hidden_size, device='cuda')
outputs = torch.randn(bsz, seq_len, hidden_size, device='cuda')

#cache = MambaCache(config, bsz, dtype=inputs.dtype, device=inputs.device)
cache = Mamba2Cache(config, bsz, dtype=inputs.dtype, device=inputs.device)
cache.reset()

cache_position = torch.arange(0, config.conv_kernel, device=inputs.device)
y_full = encoder(inputs, cache_params=cache, cache_position=cache_position)

cache.reset()

b = 0
for seg_id in range(seg_num):
    b = seg_id * seg_len
    e = min(b + seg_len, seq_len)
    cache_position = torch.arange(b, e, device=inputs.device)
    y = encoder(inputs[:, b:(b+seg_len)], cache_params=cache, cache_position=cache_position)
    err = torch.sum((y_full[:, b:(b+seg_len)] - y)**2)
    print(seg_id, err)
    b += seg_len
```

Got the following Output:

0 tensor(2.8154e-07, device='cuda:0', grad_fn=<SumBackward0>)
[...]

PR #129, opened 2025-01-20
[`Mamba2`] Fixes for caching and multiple other small issues

vasqu:
Already mentioned these fixes in #90 based on https://github.com/huggingface/transformers/pull/35154 - so it's just porting code (got a bit later to it). Some relevant issues fixed with this:
- Caching on multiple counts (fast and slow path)
- Dt clamping
- Some multi gpu stuff
- Maybe more it's been a while :)


sustcsonglin:
Thanks!

PR #130, opened 2025-01-21
[LinAttn] Fix handling of None scale in chunk_linear_attn for output normalization

HallerPatrick:
This is handled in `chunk_simple_gla`, but scale is still `None` for post output normalization


yzhangcs:
@HallerPatrick Thank you!

PR #131, opened 2025-01-21
RWKV-7 weight conversion

Triang-jyed-driung:


yzhangcs:
@Triang-jyed-driung Thank you for you hard work!
minor suggestions:
Could move `test_rwvk7.py` into `tests/utils/test_rwkv7_conversion.py` and name the functions like in https://github.com/fla-org/flash-linear-attention/blob/main/tests/ops/test_delta.py#L35

Also it would be better to do some code checks first using flake8.

Triang-jyed-driung:
> Could you move test_rwvk7.py into tests/utils/test_rwkv7_conversion.py and name the functions like in https://github.com/fla-org/flash-linear-attention/blob/main/tests/ops/test_delta.py#L35
My code has different requirements and tests different things. 
> Also it would be better to do some code checks first using flake8.
Never heard before, I need some time to get around with it

PR #134, opened 2025-01-22
Fix incorrect kwarg name in `fused_recurrent`

fffffgggg54:
Models that depend on this layer have been updated to use `cu_seqlens` instead of `offsets` as the kwarg, but the operator was not updated. This resulted in GLA and RetNet models not working, possibly others if they use them. kwarg name changes allow HF GLA and RetNet models to be loaded and run without issue. Haven't tested anything else beyond that.


yzhangcs:
Thank you

PR #137, opened 2025-01-23
Fixed dtype mismatch of mamba & mamba2 under residual_in_fp32 setting

chengshuang18:


fla/models/mamba/modeling_mamba.py
```diff
@@ -453,6 +453,8 @@ def forward(
             hidden_states, cache_params=cache_params, cache_position=cache_position, attention_mask=attention_mask
         )
         hidden_states = residual + hidden_states
+        if self.residual_in_fp32:
```
yzhangcs, line 456: It feels like if converted, the dtype of `hidden_states` will be fp32 across the layers?
This might be not expected.

yzhangcs:
@chengshuang18 Thank you!

PR #141, opened 2025-01-24
RWKV-7 fixes

Triang-jyed-driung:
1. Implemented attention mask for Channel-mixing module
2. Fixed returning `None` for logits
3. Always launch `chunk` kernel if training


yzhangcs:
Cool! Thank you.

PR #161, opened 2025-02-04
modify to enable tp for transformer model using the flame framework

xffxff:
This PR updated the transformer model to support tp using `apply_tp` from the Flame framework. The key modifications include:
1. TP splits attention heads across tp group, so the attention module has to account for it.
2. Make all the fused kernels configurable, as they are not incompatible with dtensor, allowing them to be disabled when tp is enabled.

### DIFF
--- a/fla/layers/attn.py
+++ b/fla/layers/attn.py
@@ -94,9 +94,9 @@ class Attention(nn.Module):
         if self.norm_first:
             hidden_states = self.norm(hidden_states)
 
-        q = rearrange(self.q_proj(hidden_states), '... (h d) -> ... h d', h=self.num_heads)
-        k = rearrange(self.k_proj(hidden_states), '... (h d) -> ... h d', h=self.num_kv_heads)
-        v = rearrange(self.v_proj(hidden_states), '... (h d) -> ... h d', h=self.num_kv_heads)
+        q = rearrange(self.q_proj(hidden_states), '... (h d) -> ... h d', d=self.head_dim)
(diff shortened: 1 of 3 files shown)


fla/layers/attn.py
```diff
@@ -94,9 +94,9 @@ def forward(
         if self.norm_first:
             hidden_states = self.norm(hidden_states)
 
-        q = rearrange(self.q_proj(hidden_states), '... (h d) -> ... h d', h=self.num_heads)
-        k = rearrange(self.k_proj(hidden_states), '... (h d) -> ... h d', h=self.num_kv_heads)
-        v = rearrange(self.v_proj(hidden_states), '... (h d) -> ... h d', h=self.num_kv_heads)
+        q = rearrange(self.q_proj(hidden_states), '... (h d) -> ... h d', d=self.head_dim)
+        k = rearrange(self.k_proj(hidden_states), '... (h d) -> ... h d', d=self.head_dim)
+        v = rearrange(self.v_proj(hidden_states), '... (h d) -> ... h d', d=self.head_dim)
```
xffxff, line 99: The head_dim remains unchanged when TP is enabled, but the num_heads can change.

fla/layers/attn.py
```diff
@@ -159,7 +159,7 @@ def forward(
                 causal=True,
                 window_size=(-1, -1) if self.window_size is None else (self.window_size-1, 0)
             )
-        o = o.reshape(batch_size, q_len, self.hidden_size)
+        o = o.reshape(batch_size, q_len, -1)
```
xffxff, line 162: The hidden_size can change when tp is enabled, but we can leave it inferred by torch

Issue #157, opened 2025-02-02
[Bug] Broken RWKV7 generation


Triang-jyed-driung:
### Checklist

- [x] I have checked [FAQs](https://github.com/fla-org/flash-linear-attention/blob/main/FAQs.md) and existing issues for similar problems
- [x] Please report this bug in English to ensure wider understanding and support

### Describe the Bug

RWKV-7's generation seems to have broken.

### Steps to Reproduce the Bug

```
import fla
from transformers import AutoModelForCausalLM, AutoTokenizer
name = 'fla-hub/rwkv7-191M-world'
tokenizer = AutoTokenizer.from_pretrained(name, trust_remote_code=True)
model = AutoModelForCausalLM.from_pretrained(name, trust_remote_code=True).cuda()

input_prompt = "Power goes with permanence. Impermanence is impotence. And rotation is castration."
input_ids = tokenizer(input_prompt, return_tensors="pt").input_ids.cuda()
outputs = model.generate(input_ids, max_length=64)

print(outputs)
```

### Expected Behavior

```
  File "~/flash-linear-attention/fla/models/rwkv7/modeling_rwkv7.py", line 72, in forward
    shifted = state[self.layer_idx]['ffn_state'].unsqueeze(1)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'unsqueeze'
```

### Environment Information

1. Torch: 2.3.0 and 2.5.1
2. Triton: 3.0.0 nigntly

This bug is newly introduced: around Jan 28 the code works fine.

Triang-jyed-driung:
I suspect this change introduced the bug: 
https://github.com/fla-org/flash-linear-attention/issues/152

SmerkyG:
Hmm I think the cause is a bit more subtle than this - the rwkv7 model config specifies use_cache=True, but the model code does not create an appropriate Cache object if one is not handed to it. So now, because generate became aware that the model uses a cache object, instead of being passed None, it is being passed an improper cache object type by generate(). Which is fine, if we replaced it with the correct type (our definition of "Cache").

SmerkyG:
On a separate note related to decay, I would recommend putting in an assertion to check that log_w is in float32 for the fla rwkv7 kernels (maybe all of them!). If you pass it in bfloat16 performance becomes very bad at any context length past around 512.

SmerkyG:
[...]

PR #168, opened 2025-02-05
[RWKV] Follow-up to fix cache management

jannalulu:
*added the fixes that I made to rwkv7 to rwkv6

*add check in legacy cache

(cont. implement @SmerkyG's suggestions in issue #157--they are all implemented now)


yzhangcs:
@jannalulu Hey, could you split this PR into 2 parts? one for code formats and one for RWKV cache.

PR #192, opened 2025-02-21
[RWKV] fix logits handling

jannalulu:
This fix handles the case when `num_logits_to_keep` is `None` and thus cannot be sliced.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the model’s prediction processing logic for improved adaptability. The update allows for more flexible handling of the `hidden_states` input based on the `num_logits_to_keep` parameter, enabling the use of the entire tensor when no specific number of logits is requested. This results in more robust and versatile performance when generating outputs for varied inputs.

### DIFF
--- a/fla/models/abc/modeling_abc.py
+++ b/fla/models/abc/modeling_abc.py
@@ -353,7 +353,7 @@ class ABCForCausalLM(ABCPreTrainedModel, GenerationMixin):
 
         loss, logits = None, None
         if not fuse_linear_and_cross_entropy or labels is None:
-            logits = self.lm_head(hidden_states[:, -num_logits_to_keep:])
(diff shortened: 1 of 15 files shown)


yzhangcs:
@jannalulu Great job! Could you fix other modeling files as well :D

jannalulu:
> @jannalulu Great job! Could you fix other modeling files as well :D
> 
> Also it would be better if checking the dtype of num_logits_to_keep.

Will do! Should I implement it for all the models that have this logic?

mutiann:
Hi,

It seems that the case when `num_logits_to_keep is None` is now fixed in all models in commit c4697f0, except delta_net...

https://github.com/fla-org/flash-linear-attention/blob/af3fda2018247f1d97b9db8cef41c8a70294e7e7/fla/models/delta_net/modeling_delta_net.py#L378

jannalulu:
oh you're right @mutiann, will fix

Issue #198, opened 2025-02-23
[Bug] RWKV-7 conversion results in subtly degraded performance, possibly due to numerical precision


Triang-jyed-driung:
### Checklist

- [x] I have checked [FAQs](https://github.com/fla-org/flash-linear-attention/blob/main/FAQs.md) and existing issues for similar problems
- [x] Please report this bug in English to ensure wider understanding and support

### Describe the Bug

RWKV-7 in FLA format:
|    Tasks     |Version|Filter|n-shot|  Metric  |   |Value |   |Stderr|
|--------------|------:|------|-----:|----------|---|-----:|---|-----:|
|hellaswag     |      1|none  |     0|acc       |↑  |0.5318|±  |0.0050|
|              |       |none  |     0|acc_norm  |↑  |0.7079|±  |0.0045|
|lambada_openai|      1|none  |     0|acc       |↑  |0.6883|±  |0.0065|
|              |       |none  |     0|perplexity|↓  |4.2631|±  |0.0918|

In rwkv-pip:
```
{
  "lambada_openai": {
    "perplexity,none": 4.1329989138930125,
    "perplexity_stderr,none": 0.08878972964627665,
    "acc,none": 0.6931884339219871,
    "acc_stderr,none": 0.006425006782127345,
    "alias": "lambada_openai"
  }
}
```
The perplexity degradation is close to 1.47 standard derivations (p=0.07), but worth investigating.

This is a subtle degradation, but matters for data integrity and reproducibility for our upcoming RWKV-7 paper.

### Steps to Reproduce the Bug

Reference:
https://github.com/howard-hou/VisualRWKV/blob/main/VisualRWKV-v6/v6.0/eval/run_lm_eval.py
(change to `RWKV_PAD = [0]` to align with rwkv fla)

lm-eval command:
```
lm_eval --model hf \
    --model_args pretrained=fla-hub/rwkv7-1.5B-world,dtype="float",trust_remote_code=True \
    --tasks lambada_openai,hellaswag \
    --device cuda:0 \
    --batch_size auto:4
```

### Expected Behavior

Evaluation results should align.

### Environment Information

1. Torch: 2.5.1+cu118
2. Triton: 3.0.0 nightly
3. rwkv: 0.8.28
4. lm-eval: 0.4.3

SmerkyG:
[...]

PR #199, opened 2025-02-25
[RWKV] Fix conversion imprecision

jannalulu:
Uses `nn.functional.group_norm()` and casts up the floats to try to fix issue #198. Performs the same as original fla code for now

## Summary by CodeRabbit

- **Refactor**
  - Improved the internal output normalization process to ensure consistent shape and data type handling, enhancing the overall stability of the attention mechanism.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,6 +1,8 @@
 name: pr
 
 on:
+  push:
+    branches: [ main ]
   pull_request:
     branches: [ main ]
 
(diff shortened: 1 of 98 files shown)


uniartisan:
I would prefer to use a custom Triton kernel and modify the Triton kernel to use fp32 inside the kernel, which will reduce the number of unnecessary writes.

uniartisan:
https://github.com/fla-org/flash-linear-attention/pull/200

Could you please check this PR?

jannalulu:
> Hi, is this different from loading the weight into kernels and then converting it to fp32?

Not sure; the `lambada.openai` score is still the same as described in the issue

PR #238, opened 2025-03-24
[RWKV7] fix `RWKV7Attention.__init__`

exhyy:
In the` __init__` function of `RWKV7Attention`, the `num_heads` parameter might be `None`, which causes an error in the calculation of `self.head_v_dim`. This issue can be resolved by changing it to use `self.num_heads` instead.

Test code is as follows:
```python
from fla.layers.rwkv7 import RWKV7Attention

attn = RWKV7Attention(
    mode='chunk',
    hidden_size=384,
    head_dim=64,
)
```

## Summary by CodeRabbit

- **Refactor**
  - Improved the internal calculation logic to ensure consistent dimension assignments.


yzhangcs:
@exhyy Thank you!

PR #256, opened 2025-03-29
[Gated DeltaNet] fix gdn kernel bugs on h100 when vdim=64

kugwzk:
## Summary by CodeRabbit

- **Chores**
  - Refined performance tuning settings for dynamic adjustment based on device capabilities, enhancing flexibility while maintaining existing options.


yzhangcs:
@kugwzk how about
`[2, 4] + [] if torch.cuda.get_device_capability()[0] >= 9 else [8]`

uniartisan:
We are moving towards Triton 3.3.0 adaptation, and blindly reducing wraps will significantly reduce the performance of other platforms.
e100ca10faff13e7a0214928637a3cfa
It seems that triton 3.3.0 will solve this problem? compiletimeasserterror.

I will pay a close attention on it, and maybe we can give some workaround solutions.

PR #268, opened 2025-04-01
Update fused_recurrent.py for inference with nomalization=true

yiyousong:
the current linear attention can save a $KV$ state cache. This works when normalization is not enabled. When normalization is enabled. the output should be $\frac{QKV}{QK1}$. we can see that $QK1$ or Q@sum(K) is missing earlier Keys

last pull request only modified one file, not sure why this happen, re-opened this, hope this version does contain two changes

### DIFF
--- a/fla/ops/linear_attn/fused_recurrent.py
+++ b/fla/ops/linear_attn/fused_recurrent.py
@@ -235,6 +235,7 @@ def fused_recurrent_linear_attn(
     v: torch.Tensor,
     scale: Optional[float] = None,
     initial_state: torch.Tensor = None,
+    cum_K: torch.Tensor = None,
     output_final_state: bool = False,
     normalize: bool = False,
     head_first: bool = True
@@ -245,7 +246,7 @@ def fused_recurrent_linear_attn(
         q, k, v = map(lambda x: x.transpose(1, 2), (q, k, v))
     o, final_state = FusedRecurrentLinearAttentionFunction.apply(q, k, v, scale, initial_state, output_final_state)
     if normalize:
-        o = normalize_output(q * scale, k, o)
+        o = normalize_output(q * scale, k, o,cum_K)
     if not head_first:
         o = o.transpose(1, 2)
     return o, final_state
--- a/fla/ops/linear_attn/utils.py
+++ b/fla/ops/linear_attn/utils.py
@@ -4,7 +4,9 @@ import torch
 
 
 @torch.jit.script
-def normalize_output(q, k, o):
+def normalize_output(q, k, o, cum_k=None):
     k = k.cumsum(-2)
+    if cum_k is not None:
+        k=k+cum_K
     z = (q * k).sum(-1, keepdim=True)
     return o / (z + 1e-10)


zhiyuan1i:
Thanks for contributing, can you @yiyousong  add tests to your contribution? This will improve the robustness of the code   

@yzhangcs  could you please give some comments?
tests: https://github.com/fla-org/flash-linear-attention/blob/main/tests/ops/test_linear_attn.py
layers: https://github.com/fla-org/flash-linear-attention/blob/main/fla/layers/linear_attn.py

yiyousong:
> Thanks for contributing, can you @yiyousong add tests to your contribution? This will improve the robustness of the code
> 
> @yzhangcs could you please give some comments? tests: https://github.com/fla-org/flash-linear-attention/blob/main/tests/ops/test_linear_attn.py layers: https://github.com/fla-org/flash-linear-attention/blob/main/fla/layers/linear_attn.py

Sorry, I don't think I understand how the tests works.

zhiyuan1i:
> > Thanks for contributing, can you @yiyousong add tests to your contribution? This will improve the robustness of the code
> > @yzhangcs could you please give some comments? tests: https://github.com/fla-org/flash-linear-attention/blob/main/tests/ops/test_linear_attn.py layers: https://github.com/fla-org/flash-linear-attention/blob/main/fla/layers/linear_attn.py
> 
> Sorry, I don't think I understand how the tests works.

You could have a try:)
```shell
pip install pytest
export COMPILER_MODE=1 # to speed up
pytest tests/ops/test_linear_attn.py
pytest tests/layers/test_linearatten_layer.py
```


You can see it will test function automatically. The thing you need to do is to test your `cum_k` and see if there is a need to change `fla/layers/linear_attn.py` because I see you only change the kernel.

yiyousong:
> @yiyousong Hello, could you please explain more on what does this arg mean and what's the purpose of imposing this arg

Linear attention without normalization equals to $\phi(Q)\phi(K)V$ or $\Sigma (q_i \Sigma (k_j v_j))$. 
$\Sigma_{j=1}^N (k_j v_j)=\Sigma_{j=c}^N (k_j v_j)+\Sigma_{j=1}^c (k_j v_j)=\Sigma_{j=c}^N (k_j v_j)+cache$
this is your code when initial_state!=None
The inputs $K$ should now contain tokens from position c to N (only tokens after cached positions) 


Linear attention with normalization equals to $\frac{\phi(Q)\phi(K)V}{\phi(Q)\phi(K)1}$ or $\Sigma_{i=1}^N (\frac{q_i \Sigma (k_j v_j)}{q_i \Sigma k_j})$.  
Focus on the $\Sigma k_j$. During generation, the inputs to the function is probably only just the last token (shorter than total when initial_state is not None). In this case the function will calculate the sum of K as $k_N$  which is significantly smaller than the sum of all Keys. To solve this, we need to cache the sum of the previous keys ($\Sigma_{i=1}^c k_i$). In this case the $\Sigma k_j$ can be recovered by add the cached sum ($\Sigma_{i=1}^c k_i$ ,shape [B,H,1,D]) to the cumulative sum of the current inputs ($\Sigma_{i=c}^N k_i$ ,shape [B,H,S,D]).
From the math, you can see that this issue only happens when both cache and normalization are used at the same time. 

However, implementation was harder than I thought, as the compiled function does not take if statement. So I cannot just simply add a few parameters. 

Maybe you need to change all the code involving normalization and cache

yzhangcs:
@yiyousong Thank you, good point! We do need to suuport this. But I dont think `cum_k` is a good name, some better APIs designs could be considered.
How about making normalization as a part of initial/final-state?

fla/ops/linear_attn/utils.py
```diff
@@ -4,7 +4,9 @@
 
 
 @torch.jit.script
-def normalize_output(q, k, o):
+def normalize_output(q, k, o, cum_k=None):
```
yzhangcs: @yiyousong Maybe we could pass `initial_state` as an arg with `cum_k` included for API consistency.

fla/ops/linear_attn/fused_recurrent.py
```diff
@@ -235,6 +235,7 @@ def fused_recurrent_linear_attn(
     v: torch.Tensor,
     scale: Optional[float] = None,
     initial_state: torch.Tensor = None,
+    cum_k: torch.Tensor = None,
```
yzhangcs: @yiyousong I think we could make initial_state a Tuple if normalize is True. What do you think?

fla/ops/linear_attn/fused_recurrent.py
```diff
@@ -235,6 +235,7 @@ def fused_recurrent_linear_attn(
...
     output_final_state: bool = False,
     normalize: bool = False,
```
yzhangcs, line 22: @yiyousong Could you please add some docstrings BTW.

fla/ops/linear_attn/utils.py
```diff
@@ -4,7 +4,9 @@
 
 
 @torch.jit.script
```
yzhangcs: Maybe we can use torch.compile instead.

yzhangcs:
@yiyousong Hello r u still working on this PR?

yiyousong:
> @yiyousong Hello r u still working on this PR?

I was evaluating using my own code. (only used fla.ops, not fla.layers). 
Based on my experience,  I believe although the "Z-state"(following naming from original linear attention) can be merged with the "memory-state", I would not recommend it, as "Z-state" is only needed when normalization is True.  Changes on the lower-api is unavoidable.

fla/ops/linear_attn/utils.py
yiyousong: I only use ops, so I like passing in separately. However, your code you choice. I believe it doesn't matter as long as you don't merge them into one tensor

fla/ops/linear_attn/utils.py
```diff
@@ -1,10 +1,13 @@
 # -*- coding: utf-8 -*-
+# Copyright (c) 2023-2025, Songlin Yang, Yu Zhang
 
 import torch
 
 
 @torch.jit.script
-def normalize_output(q, k, o):
+def normalize_output(q, k, o, z_state):
     k = k.cumsum(-2)
```
yzhangcs: Looks like we can directly return the last elem of k.cumsum as k.sum and return the updated z state?

fla/ops/linear_attn/fused_chunk.py
```diff
@@ -293,12 +295,17 @@ def fused_chunk_linear_attn(
...
             Initial state of shape `[B, H, K, V]`. Default: `None`.
+        z_state (Optional[torch.Tensor]):
+            Z state Of shape `[B, H, K, 1]. This is only needed when normalization is enabled. `. Default: `None`.
         output_final_state (Optional[bool]):
             Whether to output the final state of shape `[B, H, K, V]`. Default: `False`.
         normalize (bool):
             Whether to normalize the output. Default: `True`.
         head_first (Optional[bool]):
             Whether the inputs are in the head-first format. Default: `True`.
+        output_z_state (Optional[bool]):
```
yzhangcs: I think this option is unnecessary, as it must be triggered once `normalize=True` and `output_final_state=True`

yiyousong: Actually, this is included with default=False so that the current code will not break instantly.

yiyousong: But considering the original implementation is wrong, we should just break the code to notice users to update the code

yzhangcs: make sense to me

fla/ops/linear_attn/utils.py
yzhangcs: Could you use the equiv `fla.ops.utils.chunk_global_cumsum` here, which could be faster. Also, we will consider varlen mode in the future, which is not supported by torch.cumsum.

fla/ops/linear_attn/fused_recurrent.py
```diff
@@ -235,17 +235,54 @@ def fused_recurrent_linear_attn(
...
         scale = q.shape[-1] ** -0.5
     if not head_first:
         q, k, v = map(lambda x: x.transpose(1, 2), (q, k, v))
     o, final_state = FusedRecurrentLinearAttentionFunction.apply(q, k, v, scale, initial_state, output_final_state)
+    
     if normalize:
-        o = normalize_output(q * scale, k, o)
+        if z_state is None:
+            k_shape = list(k.shape)
+            k_shape[-2 ]= 1
```
yzhangcs: Removing the space in [-2] would be better.
How about directly init z_state by 
```py
z_state =  torch.zeros_like(k[..., 0, :]) if z_state is None else z_state 
```

yiyousong:
These changes are based on the code I changed to work for my model. I probably won't work on this further. 
Maybe after May 15th I may continue to update this, but currently I am too busy to debug code I won't use.

fla/ops/linear_attn/fused_recurrent.py
yzhangcs: also I think `[B, H, K, 1]` could be confusing, would `[B, H, K]` be better.
There's no cost for unsqueeze when updating z state

fla/ops/linear_attn/fused_chunk.py
```diff
@@ -293,12 +294,15 @@ def fused_chunk_linear_attn(
             If not provided, it will default to `1 / sqrt(K)`. Default: `None`.
         initial_state (Optional[torch.Tensor]):
             Initial state of shape `[B, H, K, V]`. Default: `None`.
+        z_state (Optional[torch.Tensor]):
+            Z state Of shape `[B, H, K, 1]. This is only needed when normalization is enabled. `. Default: `None`.
```
yzhangcs: Of -> of

yzhangcs: `. Default:` -> 'Default:'

fla/ops/linear_attn/fused_chunk.py
```diff
@@ -293,12 +294,15 @@ def fused_chunk_linear_attn(
...
         output_final_state (Optional[bool]):
```
yzhangcs, line 40: Let's change the doc string accordingly, as final state has been changed to Union[torch.Tensor, Tuple[torch.Tensor]]

yiyousong: I am not very familiar with writing docstrings, sorry. BTW, I think you can directly modify the code yourself, right?

fla/ops/linear_attn/utils.py
yiyousong: I'm not sure how. maybe you can directly modify the code?

PR #270, opened 2025-04-01
[RWKV7] use `tl.gather` to enhance performance

zhiyuan1i:


fla/ops/rwkv7/fused_addcmul.py
yzhangcs, line 1: @zhiyuan1i We should separate these changes from the current commit?

fla/utils.py
```diff
@@ -167,6 +167,13 @@ def _check_platform() -> Literal['nvidia', 'amd', 'intel', 'musa']:
 
 # Nvidia Ampere or newer, haven't check AMD and intel yet.
 is_tf32_supported = (is_nvidia and torch.cuda.get_device_capability(0)[0] >= 8)
+triton_have_gather = hasattr(triton, 'gather')
+
+
+if not triton_have_gather:
```
yzhangcs: @zhiyuan1i I dont's think it's a good name lol

fla/ops/rwkv7/fused_addcmul.py
zhiyuan1i: > @zhiyuan1i We should separate these changes from the current commit?

of course, i will try to rebase

fla/utils.py
zhiyuan1i: > @zhiyuan1i I dont's think it's a good name lol

you can see i'm poor on naming lol.  I will try to make it clear later.

yzhangcs: I prefer not making `fla.utils` too heavy, maybe we can directly call `hasattr(triton, 'gather')`

zhiyuan1i: > I prefer not making `fla.utils` too heavy, maybe we can directly call `hasattr(triton, 'gather')`

however, we need to patch triton to skip ast error

zhiyuan1i:
GLA has a numerical issue with triton 3.3.0 and needs to be investigated further.

PR #275, opened 2025-04-02
[Fox] Add model/layer/kernel impls w/ varlen support

yzhangcs:
## Summary by CodeRabbit

- **New Features**
	- Introduced the `ForgettingAttention` class to enhance attention mechanisms.
	- Released the `FoX` transformer-based model along with associated configuration options for improved language modeling.
	- Implemented optimized parallel computation features for attention mechanisms.
	- Updated the README to include details about the new FoX implementation and its associated resources.
- **Tests**
	- Added a comprehensive suite of tests for the `parallel_fox` function to validate its functionality and accuracy.


zhiyuan1i:
How do we test the correctness of the fox layer?
 Is it possible to move forward correctly under various premises? :)

yzhangcs:
> How do we test the correctness of the fox layer? Is it possible to layer forward correctly under various conditions? :)

We could run `python -m benchmarks.benchmark_training_throughput --name fox` to ensure the training is ok.
What do you think.

zhixuan-lin:
@yzhangcs I went over the code, and I have some suggestions:

1. Should we set `initializer_range` in `modeling_fox.py` and `configuration_fox.py` to `0.02` instead of `0.006`? Also, it looks like `initializer_range` in the standard transformer is also set to `0.006`, which you might want to change as well.
2. It is probably better to use Group RMSNorm for the QK-norm implementation. So instead of `RMSNorm(self.head_dim)` you do something like `GroupRMSNorm(num_groups=self.num_heads, hidden_size=self.hidden_size)`. This has been fixed in my official repository as well. See this [changelog](https://github.com/zhixuan-lin/forgetting-transformer?tab=readme-ov-file#changelog) and the [related code](https://github.com/zhixuan-lin/forgetting-transformer/blob/321e7fa4f41abb5466c5b4ff8ba60b3a18a48a4b/src/forgetting_transformer/model/forgetting_transformer/modeling_forgetting_transformer.py#L290-L296). In my code `RMSNorm(self.head_dim)` is just kept for backward compatibility but in your implementation you can just remove it to avoid confusion.
3. In the varlen forgetting attention implementation, it might be a good idea to mention in the docstring that the batch size should be 1 when `cu_seqlens` is not `None`
4. I see that `past_key_values` is currently not used in the forgetting attention layer. Just a head-up that if you want to implement this feature later, it would require storing past forget gate values in `past_key_values` as well, see [this file](https://github.com/zhixuan-lin/forgetting-transformer/blob/321e7fa4f41abb5466c5b4ff8ba60b3a18a48a4b/src/forgetting_transformer/model/forgetting_transformer/fgate_cache.py).
5. You could test the FoX layer and model against my official implementation (they are easily importable) to ensure they are correctly implemented. You can test them with QK-norm on or off.
6. This point is purely about code style and totally optional, but it might be a good idea to change `fox` or `FoX` in all filenames, class names, `model_type` in `PretrainedConfig` to `forgetting_transformer` or `ForgettingTransformer`. For filenames you can use `git mv` to retain git history. This has two purposes: (1) to avoid potential conflict and error with any other (future) models named `fox`, especially when registering the model class to the Hugging Face model registry (2) it is more explicit; people might not know what `fox` but they will know what `forgetting_transformer` is. But again, this is purely an optional suggestion. It is totally fine if you prefer the shorter `fox` and `FoX` for the filenames, class names, and `model_type` (they are more concise after all).

yzhangcs:
@zhixuan-lin Thank you for your great suggestions, all of them are very valuable!
I will fix them soon, and your PRs are welcome.

PR #329, opened 2025-04-08
[Triton] Fix `Call Undecorated gather Functio` on Triton 3.1.0

zhiyuan1i:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved an internal operation to consistently return the provided input under specific conditions, ensuring expected and predictable behavior.
	- Enhanced error calculation precision and introduced a mechanism for early exit based on absolute error tolerance.


zhiyuan1i:
I found that this pr exposes a flaw in CI recursively finding the corresponding modification.
No dependencies found for dplr, which is noteworthy and investigated

PR #360, opened 2025-04-16
Fix rwkv4/fused_recurrent.py backward

johanwind:
I found and fixed two bugs in rwkv4/fused_recurrent.py 
1.
The forward pass starts with `w = -torch.exp(w.float().contiguous())`. However, the backward pass doesn't take this into account. To get the correct gradient for w, we can simply multiply gw by w.
2.
The storage of gw and gu should use atomic adds. The current naive approach of loading, adding and storing, leads to race conditions which often causes the accumulated gw and gu to be incorrect. Since atomic adds aren't supported for bfloat16, this also requires making the gw and gu tensors float32.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved gradient accumulation for parameters to ensure more accurate and reliable updates.
  - Ensured gradient tensors use consistent precision for better stability during training.


yzhangcs:
@johanwind Hi, thank you for your contributions! Could you revise `atomic_add` to reduction outside, i.e., materializing a big tensor and saving the results to the tensor, and finally reduction outside. Generally we do not use any `atomic_add` in `fla`, and use reduction instead, which could lead to better performance.

yzhangcs (approved):

johanwind:
I replaced the atomic_add with a reduction in the style used in rwkv6/chunk.py.
In a simple test atomic_add was slightly faster than reduction, but I understand wanting to be consistent.
CodeRabbit pointed out that the return dtypes are different from input dtypes (this was also the case before my changes), so I also fixed that.

PR #361, opened 2025-04-17
Use `tl.exp2` for all gating operations

yzhangcs:
In https://github.com/triton-lang/triton/issues/2893

> exp2 is faster than exp because the special function unit inside the GPU implements exp2. When you use the exp intrinsic, it expands to a multiplication by 1/log(2) followed by the exp2 instruction. In this case, we can fold this 1/log(2) multiplier into the sm_scale multiplier so that we need only perform one multiplication per element rather than two.

## Summary by CodeRabbit

- **New Features**
  - Added support for base-2 exponential functions with safe computation handling.
  - Introduced a new mathematical constant to improve accuracy in logarithmic and exponential operations.
  - Updated attention kernel calculations to utilize base-2 logarithms and exponentials, enhancing performance and precision.
  - Added optional scaling to cumulative sum operations, allowing enhanced flexibility in computations.
- **Tests**
  - Expanded test coverage for attention scaling factors to ensure robustness across different scale values.

### DIFF
--- a/fla/ops/attn/parallel.py
+++ b/fla/ops/attn/parallel.py
@@ -11,7 +11,7 @@ from einops import reduce
 
 from fla.ops.utils import prepare_chunk_indices
 from fla.ops.utils.cumsum import chunk_global_cumsum
-from fla.ops.utils.op import exp2, log2, safe_exp2
+from fla.ops.utils.op import exp2, log2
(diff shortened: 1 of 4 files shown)


yzhangcs:
@yzhangcs be careful that `RCP_IN2` should be multiplied after `tl.dot`, otherwise would lead to big precision loss.

Issue #367, opened 2025-04-21
[Bug] Wrong initialization logic


Triang-jyed-driung:
### Checklist

- [x] I have checked [FAQs](https://github.com/fla-org/flash-linear-attention/blob/main/FAQs.md) and existing issues for similar problems
- [x] My GPU is H100 and I have installed `triton-nightly` built by fla team, and double checked [FAQs](https://github.com/fla-org/flash-linear-attention/blob/main/FAQs.md)
- [x] Please report this bug in English to ensure wider understanding and support

### Describe the Bug

As for `transformers==v4.51.3`, all initialization of weights is supposed to be done within `PreTrainedModel` class. 
See https://github.com/huggingface/transformers/blob/v4.51.3/src/transformers/modeling_utils.py#L2436
And also https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen2/modeling_qwen2.py#L330
for implementation details.

### Steps to Reproduce the Bug

Wrong logic for those files:
1. RWKV6: https://github.com/fla-org/flash-linear-attention/blob/main/fla/layers/rwkv6.py#L84
2. RWKV7: https://github.com/fla-org/flash-linear-attention/blob/main/fla/layers/rwkv7.py#L110
These functions are called in the wrong place at the wrong time.
Therefore, `if getattr(module, "_is_hf_initialized", False)` will never be `True`. 

### Expected Behavior

All initialization should be written in the `xxxPreTrainedModel` class. Like this:
```python
def _init_weights(
            self, 
            module
            # No other positional or keyword parameters here!
        ):
        std = self.config.initializer_range
        # You can access self.config for other initialization hyperparameters
        if isinstance(module, nn.Linear):
            # You can initialize its submodules
            module.weight.data.normal_(mean=0.0, std=std)
            if module.bias is not None:
                module.bias.data.zero_()
        elif isinstance(module, nn.Embedding):
            module.weight.data.normal_(mean=0.0, std=std)
            if module.padding_idx is not None:
                module.weight.data[module.padding_idx].zero_()
        elif isinstance(module, Qwen2RMSNorm):
            module.weight.data.fill_(1.0)
```

### Environment Information

1. Python: 3.13.2
2. Torch: 2.7.0 pre-release
3. Triton: 3.3.0 pre-release
4. Transformers: 4.51.3

zhiyuan1i:
[...]

PR #369, opened 2025-04-21
[RWKV]: Prevent initialization when loading pretrained weights

zhiyuan1i:
- Added guard `_is_hf_initialized` flag check in `_initialize_weights`


zhiyuan1i:
This PR fixed https://github.com/fla-org/flash-linear-attention/issues/367

Issue #392, opened 2025-05-10
[RFC] Add MLA layers into fla


yzhangcs:
### Proposal

We could adapt [hf impls](https://github.com/huggingface/transformers/blob/main/src/transformers/models/deepseek_v3/modeling_deepseek_v3.py#L328) into fla with support for varlen training and batched inference.

yzhangcs:
@toothacher17

toothacher17:
# Example
An example of code inside FLA repo: https://github.com/fla-org/flash-linear-attention/blob/main/fla/layers/attn.py

# Goal
Supporting varlen training and batched inference

PR #395, opened 2025-05-12
[WIP] Add MLA layers into fla

toothacher17:
Initial impl, draft for now, needs setting up tests

Relates to: https://github.com/fla-org/flash-linear-attention/issues/392

Local generation unit tests:

image

## Summary by CodeRabbit

- **New Features**
  - Introduced a new transformer architecture featuring multihead latent attention with rotary embeddings and LoRA adapters.
  - Added full support for causal language modeling with efficient caching and generation capabilities.
  - Integrated fused operations for improved performance including normalization, activation, and cross-entropy loss.
  - Enabled seamless model configuration and instantiation via Hugging Face Transformers integration.
  - Provided comprehensive testing for model generation accuracy.
  - Added public API exports for MLA components to simplify usage.
  - Included a detailed news announcement highlighting the MLA implementation.


fla/layers/mla.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+
+    def __init__(
+        self,
+        hidden_size: int = 2048,
+        num_heads: int = 32,
+        rope_theta: float = 10000.,
+        max_position_embeddings: Optional[int] = None,
+        q_lora_rank: Optional[int] = 1536,  # q lora rank is optional, None indicates no q lora
+        qk_rope_head_dim: int = 64,
```
yzhangcs, line 56: @toothacher17 How about mark it as Optional[int] = 64, in cases that we could provide support for NoPE in MLA

fla/layers/mla.py
```diff
@@ -0,0 +1,218 @@
...
+        # if attention_mask is not None, this is doing inference
+        if attention_mask is not None:
+            assert len(attention_mask.shape) == 2, (
+                "Expected attention_mask as a 0-1 matrix with shape [batch_size, seq_len] "
+                "for padding purposes (0 indicating padding). "
+                "Arbitrary attention masks of shape [batch_size, seq_len, seq_len] are not allowed."
+            )
+
+        # prepare q k v projections
+        batch_size, seq_length = hidden_states.shape[:-1]
```
yzhangcs: let's name it as `q_len` for inference considerations, the input length may be shorter than the kv cache length, as in layers/attn.py

fla/layers/mla.py
```diff
@@ -0,0 +1,218 @@
...
+            self.q_a_layernorm = RMSNorm(q_lora_rank)
+            self.q_b_proj = linear(q_lora_rank, self.num_heads * self.qk_head_dim)
+        else:
+            self.q_proj = linear(hidden_size, self.num_heads * self.qk_head_dim)
+
+        self.kv_a_proj_with_mqa = linear(
+            hidden_size,
+            self.kv_lora_rank + self.qk_rope_head_dim,
+        )
+        self.kv_a_layernorm = RMSNorm(self.kv_lora_rank)
```
yzhangcs: @toothacher17 Not sure if it's ok to provide an option for controlling this norm. What do you think?

fla/layers/mla.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+
+        self.scaling = self.qk_head_dim ** (-0.5)
+        if rope_scaling is not None:
+            mscale_all_dim = rope_scaling.get("mscale_all_dim", 0)
+            scaling_factor = rope_scaling["factor"]
+            if mscale_all_dim:
+                mscale = yarn_get_mscale(scaling_factor, mscale_all_dim)
+                self.scaling = self.scaling * mscale * mscale
+
```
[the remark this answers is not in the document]
toothacher17: self.scaling is actually used in flash_attn's interface. Updated accordingly

fla/layers/mla.py
```diff
@@ -0,0 +1,219 @@
...
+try:
+    from flash_attn import flash_attn_func, flash_attn_varlen_func
+except ImportError:
+    warnings.warn(
+        "Flash Attention is not installed. Please install it via `pip install flash-attn --no-build-isolation`",
+        category=ImportWarning
+    )
+    flash_attn_func = None
+
```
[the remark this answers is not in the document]
toothacher17: Resolved by adding flash_attn_varlen_func = None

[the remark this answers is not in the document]
toothacher17: Actually, after a second thought, there is no need for this try except. If the flash attn is not installed, just directly throw the ImportError and let the user install the flash-attn version.

fla/layers/mla.py
```diff
@@ -0,0 +1,219 @@
...
+            cu_seqlens_q=cu_seqlens,
+            cu_seqlens_k=cu_seqlens,
+            max_seqlen_q=max_seqlen,
+            max_seqlen_k=max_seqlen, 
+            causal=True,
+        )
+
+        if self.qk_head_dim != self.v_head_dim:
+            attn_output = attn_output[:, :, :, : self.v_head_dim]
+
```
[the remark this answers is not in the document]
toothacher17: code logic changed, this part no longer has issues

fla/layers/mla.py
toothacher17: I felt maybe we should not mark it as Optional? Because:
1. Since this class is designed to implement Deepseek's MLA, we should mostly follow their settings.
2. If we want NoPE, maybe setting this value to 0?

fla/layers/mla.py
toothacher17: got it! Changed as suggested

fla/layers/mla.py
toothacher17: Prefer not marking it as optional. As mentioned in DSV2's paper, it seems that this RMSNorm helps improve the training stability

toothacher17:
I resolved the comments and fixed the lint today, but the implementation logic is wrong. I'll fix the impl and add more tests later

PR #396, opened 2025-05-12
[DeltaProduct] fix query conv cache, remove extraneous query convs

timurcarstensen:
This PR fixes an issue where the query convolutional cache is never set which leads to incorrect text generation. It also removes unused query convs since we only need one (right now we initialise n_h query convs but only use the last one)

## Summary by CodeRabbit

- **Refactor**
  - Optimized query processing in the GatedDeltaProduct module to reduce redundant computations, resulting in improved efficiency with no changes to the user interface or outputs.

### DIFF
--- a/fla/layers/gated_deltaproduct.py
+++ b/fla/layers/gated_deltaproduct.py
@@ -119,15 +119,10 @@ class GatedDeltaProduct(nn.Module):
             ]
         )
         if use_short_conv:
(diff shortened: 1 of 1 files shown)


zhiyuan1i:
Thank you very much for your contribution, since I'm not the owner of the code, I would like to ask if this potential issue can be triggered by our unit tests? How can we avoid such errors in future code commits？

PR #434, opened 2025-06-08
Update mesa_net.py

jovoswald:
Change static lower bound to self.lambda_lower_bound

## Summary by CodeRabbit

- **Improvements**
  - Enhanced configurability by allowing the lower bound for certain internal calculations to be adjusted, instead of using a fixed value. No visible changes to the user interface or exported features.


yzhangcs:
Thank you

PR #505, opened 2025-07-02
Support chunk-parallel via combined `chunk` and `fused_chunk`

yzhangcs:


yzhangcs:
@yzhangcs This PR is not ready for merging as we do not observe some accelerations yet.

PR #531, opened 2025-07-17
[Tokenshift] Support SP and cache

zhiyuan1i:
## Summary by CodeRabbit

* **New Features**
  * Added optional caching support to token shift operations, enabling stateful processing across calls.
  * Updated documentation to describe new cache-related behavior and outputs.

* **Tests**
  * Introduced new tests to verify consistency of token shift outputs and gradients when using caching, including split and whole input scenarios.

### DIFF
--- a/fla/modules/token_shift.py
+++ b/fla/modules/token_shift.py
@@ -50,7 +50,7 @@ def token_shift_ref(
 
 @triton.heuristics({
     'IS_VARLEN': lambda args: args['cu_seqlens'] is not None,
-    'HAS_CACHE': lambda args: args['cache'] is not None,
+    'USE_INITIAL_STATE': lambda args: args['cache'] is not None,
 })
 @triton.autotune(
     configs=[
@@ -71,8 +71,8 @@ def token_shift_fwd_kernel_short(
     D: tl.constexpr,
     BD: tl.constexpr,
     IS_VARLEN: tl.constexpr,
-    HAS_CACHE: tl.constexpr,
-    OUTPUT_CACHE: tl.constexpr,
+    USE_INITIAL_STATE: tl.constexpr,
+    STORE_FINAL_STATE: tl.constexpr,
 ):
     i_b, i_t = tl.program_id(0), tl.program_id(1)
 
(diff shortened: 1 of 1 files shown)


zhiyuan1i:
I think we need to talk about the specific design issues. About Varlen. Varlen doesn't seem to need to use caching or SP, because we can simply slice between different sequences. Or we need to shard each batch in the same place. These treatments will be very different and will need to be discussed and designed. @yzhangcs

fla/modules/token_shift.py
```diff
@@ -49,7 +49,8 @@ def token_shift_ref(
 
 
 @triton.heuristics({
-    'IS_VARLEN': lambda args: args['cu_seqlens'] is not None
+    'IS_VARLEN': lambda args: args['cu_seqlens'] is not None,
+    'HAS_CACHE': lambda args: args['cache'] is not None,
```
yzhangcs: @zhiyuan1i Let's unify the names to initial_state & final_state as in linear attn kernels

yzhangcs:
@zhiyuan1i Does short and long has significant diffs in terms of speed?

zhiyuan1i:
> @zhiyuan1i Does short and long has significant diffs in terms of speed?

short is slightly faster, since it's a really simple kernel.

yzhangcs:
> short is slightly faster, since it's a really simple kernel.

in all cases?

zhiyuan1i:
> > short is slightly faster, since it's a really simple kernel.
> 
> in all cases?

haven't check yet :( only e2e testing, seems proportional to length
```
         T  naive_token_shift  fused_token_shift  naive_token_shift_bwd  fused_token_shift_bwd
0    128.0           0.025088           0.010992               0.084416               0.163040
1    256.0           0.041152           0.013952               0.105728               0.163616
2    512.0           0.083552           0.024480               0.207552               0.165664
3   1024.0           0.153312           0.041184               0.387136               0.172960
4   2048.0           0.293008           0.076288               0.741344               0.244480
5   4096.0           0.571776           0.146112               1.453344               0.472032
6   8192.0           1.127392           0.281088               2.877296               0.932496
7  16384.0           2.246016           0.553152               5.727904               1.849424
8  32768.0           4.482624           1.092736              11.432512               3.673760
```

PR #535, opened 2025-07-21
[ShortConv] Support cache in prefill

zhiyuan1i:
## Summary by CodeRabbit

* **New Features**
  * Added support for an optional cache tensor in causal 1D convolution, enabling use of initial state for incremental or streaming scenarios.
  * Updated the ShortConvolution module to accept and update a cache during the forward pass.

* **Tests**
  * Introduced new internal reference implementations and comprehensive tests for ShortConvolution with cache prefill and incremental decoding.
  * Added tests verifying mixed backend usage to ensure consistent outputs across different convolution backends.

### DIFF
--- a/fla/modules/convolution.py
+++ b/fla/modules/convolution.py
@@ -493,7 +493,7 @@ class CausalConv1dFunction(torch.autograd.Function):
     @staticmethod
     @input_guard
     def backward(ctx, dy: torch.Tensor, dht: Optional[torch.Tensor] = None):
-        if dht is None:
+        if dht is not None:
             raise NotImplementedError("The gradient of the final state is not supported yet.")
         x, weight, bias, residual, initial_state = ctx.saved_tensors
         dx, dw, db, dr, dh0 = causal_conv1d_bwd(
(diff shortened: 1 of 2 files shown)


zhiyuan1i:
backward will land later, similar to token-shift, support non-varlen first. then varlen.

yzhangcs:
@zhiyuan1i I've just updated the kernels.
sth remains to test & impl
1. we may need to check copying previous cache into the new ones once the seq is too short, say T < W.
2. for varlen or non-varlen conv, check if things are ok if taking rand init cache as input.
3. add support for bwd w. prev cache grad

zhiyuan1i:
> @zhiyuan1i I've just updated the kernels. sth remains to test & impl
> 
> 1. we may need to check copying previous cache into the new ones once the seq is too short, say T < W.
> 2. for varlen or non-varlen conv, check if things are ok if taking rand init cache as input.
> 3. add support for bwd w. prev cache grad

@yzhangcs We can implement items 1 and 2 in this PR, and item 3 will be addressed in another PR later.

PR #556, opened 2025-08-06
[Fix] Fix Triton Error for HeadDim < 16

zhiyuan1i:


zhiyuan1i:
Merge, since `tests/ops/test_simple_gla.py::test_simple_gla_to_mamba2[vary_ATrue-dtypetorch.float32]` fail on main branch

PR #569, opened 2025-08-20
add layer_classes=transformers.cache_utils.DynamicLayer

richardodliu:
When using the HuggingFace transformer model.generate method, it will raise an error if not init layer_classes.

## Summary by CodeRabbit

- Refactor
  - Updated cache initialization to use dynamic layer handling for better compatibility and stability during model runs.
- Performance
  - Improved reliability of state management under varying model depths, reducing edge-case failures.
- Chores
  - No public API changes; existing behavior remains consistent for end users.

### DIFF
--- a/fla/models/utils.py
+++ b/fla/models/utils.py
@@ -21,7 +21,7 @@ class Cache(transformers.cache_utils.Cache):
         self,
         seen_tokens: int = 0
     ) -> Cache:
-        super().__init__()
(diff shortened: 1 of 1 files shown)


richardodliu:
I find this test failed since I have the latest transformer version(4.55.0) in my conda env, while the workflow has 4.52.4.
This is the question.

PR #570, opened 2025-08-20
require transformers <= 4.53.3

richardodliu:
require transformers <= 4.53.3, otherwise transformers will raise an error when generate

## Summary by CodeRabbit

* Chores
  * Updated dependency constraints to bound the transformers package to versions 4.45.0–4.53.3 (inclusive).
  * This affects installation and environment resolution only; existing features and behavior remain unchanged.
  * No impact on configuration, APIs, or user workflows.
  * No action required unless you manage pinned dependencies; ensure your environment uses a transformers version within this range.

### DIFF
--- a/fla/models/utils.py
+++ b/fla/models/utils.py
@@ -21,7 +21,7 @@ class Cache(transformers.cache_utils.Cache):
         self,
         seen_tokens: int = 0
     ) -> Cache:
-        super().__init__()
+        super().__init__(layer_classes=transformers.cache_utils.DynamicLayer)
 
         self.states: List[Dict[str, Any]] = []
 
--- a/fla/models/utils.py
+++ b/fla/models/utils.py
@@ -21,7 +21,7 @@ class Cache(transformers.cache_utils.Cache):
         self,
         seen_tokens: int = 0
     ) -> Cache:
-        super().__init__(layer_classes=transformers.cache_utils.DynamicLayer)
+        super().__init__()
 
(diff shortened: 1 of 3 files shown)


zhiyuan1i:
Since https://github.com/vllm-project/vllm/blob/main/requirements/common.txt#L10C4-L10C23 only support:transformers >= 4.55.2
Should we merge this PR first and then proceed with adopting transformers, or should we just move forward directly?

CC @yzhangcs

richardodliu:
> Since https://github.com/vllm-project/vllm/blob/main/requirements/common.txt#L10C4-L10C23 only support:transformers >= 4.55.2 Should we merge this PR first and then proceed with adopting transformers, or should we just move forward directly?
> 
> CC @yzhangcs

Thanks for your comment. The reason require transformers <= 4.53.3 is because if  transformers > 4.53.3, I have found another error in my another pr https://github.com/fla-org/flash-linear-attention/pull/569.

zhiyuan1i:
> > Since https://github.com/vllm-project/vllm/blob/main/requirements/common.txt#L10C4-L10C23 only support:transformers >= 4.55.2 Should we merge this PR first and then proceed with adopting transformers, or should we just move forward directly?
> > CC @yzhangcs
> 
> Thanks for your comment. The reason require transformers <= 4.53.3 is because if transformers > 4.53.3, I have found another error in my another pr #569.

@richardodliu  could you please kind provide a script which could reproduce your error. Thanks

PR #577, opened 2025-08-26
[Log Linear Attention] add backward pass

2022tgoel:


fla/ops/log_linear_attn/chunk.py
```diff
@@ -907,6 +908,589 @@ def copy_last_chunk_kernel(
...
+@triton.heuristics({"IS_VARLEN": lambda args: args["cu_seqlens"] is not None})
+@triton.autotune(
+    configs=[
+        triton.Config({"BK": BK}, num_warps=num_warps, num_stages=num_stages)
+        for BK in [32, 64, 128]
+        for num_warps in [4]
+        for num_stages in [2, 3, 4]
+    ],
+    key=["H", "K", "V"],
+    restore_value=["dh", "dg_last"],
```
yzhangcs, line 919: @2022tgoel IIRC, this kwarg was introduced by triton>=3.1 or higher, could you consider using another brand-new dh variable or add some version checks?

fla/ops/log_linear_attn/chunk.py
```diff
@@ -1151,10 +1748,161 @@ def grid(meta):
...
+            L=L,
+            BT=BT,
         )
 
+        dg = chunk_local_cumsum(
+            dg,
+            chunk_size,
+            reverse=True,
+            offsets=None,
+            head_first=None,
```
yzhangcs: @2022tgoel I think we can directly delete the `head_first` arg for recently introduced kernels

yzhangcs:
Hi @2022tgoel, thanks for your excellent PR.

Here are a few optional suggestions for follow-up improvements:
1. Remove `safe_exp` usage: If possible, consider replacing all calls to `safe_exp`. This op is unsafe, and we recommend using explicit masking instead. We are planning to deprecate and remove this function in a future version. Of course current impls already LGTM.
2. Add model forward and generation checks to the test suite here: tests/models.
3. To help everyone see your contribution, could you please add a note about your changes to the NEWS and Models sections of the README?

Thanks again for the great work!

PR #585, opened 2025-09-07
[DeltaFormer] Add Model

Nathancgy:
- Introduces [DeltaFormer](https://arxiv.org/pdf/2505.19488) layer and model.
- Adds fused Triton ops for chunked deltaformer attention: `fla/ops/deltaformer/parallel.py`, `invcum.py`, and `naive.py`. The pre-attention calculates the u vector, which is a substitute for v that can be directly integrated into flash attention func.
- Kernels 1 and 2 are both softmax in this version, omitted alpha scaling factor for v in u's calculation.
- The current similarity calculation comes in the form of dot products between k and k (as in the original paper)
- Includes unit tests and integration updates for the new model.
- Supports varlen.
- Preserves backward compatibility. The optional `beta` defaults to all ones when not provided.

## Summary by CodeRabbit

- **New Features**
  - Adds the DeltaFormer model family (config, blocks, CausalLM) and a Triton-accelerated DeltaFormer attention operator with packed/variable-length support plus a naive reference and linear-algebra helpers.

- **Documentation**
  - Adds/updates README news entry and models table for DeltaFormer; merges blog link into the announcement.

- **Tests**
  - Adds unit tests for modeling, generation, and attention (including packed/variable-length cases).

- **Chores**
  - Bumps package version to 0.3.2 and updates public exports.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,5 +25,6 @@ jobs:
         TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
       run: |
-        python setup.py sdist bdist_wheel
-        twine upload dist/*
+        pip install -U build
+        python scripts/build_packages.py
+        twine upload dist-packages/*
--- a/.github/workflows/reusable-ci-tests.yml
+++ b/.github/workflows/reusable-ci-tests.yml
@@ -137,13 +137,13 @@ jobs:
               NIGHTLY_URL="https://download.pytorch.org/whl/nightly/${{ inputs.pytorch_cuda_version }}"
               echo "Using nightly index URL: $NIGHTLY_URL"
               $CONDA_BIN_PATH/pip install -U torch pytorch-triton --index-url $NIGHTLY_URL
-              $CONDA_BIN_PATH/pip install -U numpy packaging psutil ninja einops datasets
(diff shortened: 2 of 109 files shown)


yzhangcs:
@Nathancgy Hi, thanks for the PR.
Some issues before merging:
1. Try running `python -m benchmarks.benchmark_training_throughput --name deltaformer --batch_size 1 --seq_len 4096`, it appears that the default block size is too big.
2. All tests in varlen training and generation for deltaformer are skipped, are you planning to add supports for them in this PR?

Nathancgy:
@yzhangcs Hi, thanks for helping with the minor fixes!
1. I have just added autotune for all the kernels, the issue is now fixed. The test result (`python -m benchmarks.benchmark_training_throughput --name deltaformer --batch_size 1 --seq_len 4096`) on a H100 gives 9882.56 token/s for the throughput
2. I will try adding them tonight. But if they're not here by next morning, then I'll probably have to add them next weekend as I have an upcoming test week from school. Will try to get it done by eod.

fla/ops/deltaformer/invcum.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2023-2025, Songlin Yang, Yu Zhang
+
+import torch
+
+
+def forward(u, w):
```
yzhangcs, line 7: @Nathancgy let's give them more specific names

Nathancgy: will do, also likely to finish the in-kernel varlen support by eod!

PR #590, opened 2025-09-12
[GDN] fuse g calculations

isayoften:
Simple fusion that will look like:
image

instead of:
image

## Summary by CodeRabbit

* **Refactor**
  * Optimized the gating computation in the Gated DeltaNet layer to reduce compute overhead and improve runtime efficiency during training and inference.
  * Streamlined the forward path by delegating gating to a dedicated, compiled routine, enabling better performance without altering existing behavior or outputs.
  * Users can expect smoother throughput and potentially lower latency on supported hardware, with no changes required to existing workflows or configurations.

### DIFF
--- a/fla/layers/gated_deltanet.py
+++ b/fla/layers/gated_deltanet.py
@@ -31,6 +31,9 @@ def elu_p1(x):
 def sum_norm(x):
     return (x / x.sum(-1, keepdim=True)).to(x)
 
+@torch.compile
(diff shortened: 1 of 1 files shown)


zhiyuan1i:
Thanks for your contributions, I think people can directly compile the entire layer when needed, because we found in our tests that simple compiles sometimes bring NAN or trigger a recompilation when the shape changes, and maybe we'll customize a triton kernel when appropriate

PR #602, opened 2025-09-25
[DeltaFormer] Fixed testing ops error

Nathancgy:
- Fixed error in parallel.py
- Updated ops test from (B, H, T, D) to (B, T, H, D)
- Updated naive.py for deltaformer to adapt to the new dim
- Ops & Model tests pass

## Summary by CodeRabbit

- **New Features**
  - Adds a head-first two-stage DeltaFormer attention flow and explicit support for both BTHD and BHTD tensor layouts.

- **Bug Fixes**
  - Fixed query-handling in the parallel attention path, improved chunk assembly, and added boundary checks for safer attention execution.

- **Tests**
  - Tests moved to float16, updated to (B, T, H, D)/(B, T, H) layouts, added varlen coverage, and slightly adjusted tolerances.

- **Documentation**
  - Clarified layout conventions and the two-stage attention workflow; public API behavior preserved.

PR #603, opened 2025-09-30
[Deltaformer] kernel improvement; if-else optimization; change w to fp32; add 1e-9 to avoid nan

foreverpiano:
if-else optimization; change w to fp32; add 1e-9 to avoid nan

## Summary by CodeRabbit

- Bug Fixes
  - Improved numerical stability by adding a small epsilon to normalization.
  - Ensured consistent dtype handling with explicit float32 casting across forward paths.
  - Corrected masking and boundary handling for late-token processing to prevent incorrect updates.
- Refactor
  - Restructured forward-pass processing into two phases for clearer control flow and consistent masking.
  - Simplified data flow by lifting computations out of inner loops for readability and maintainability.


foreverpiano:
@yzhangcs Any plan for adding tma torch_desc version? I have got one for this. @Nathancgy

Nathancgy:
@foreverpiano thanks for the pr (the two separate loops look good)! For adding tma torch_desc version, it is up to zhangyu to decide. But in the meanwhile, if you need to test the model while making further improvements, I recommend using the version from [this pr](https://github.com/fla-org/flash-linear-attention/pull/602), where I fixed a few accuracy issues. This version passes the ops test and should be correct.

zhiyuan1i:
Closing this. The kernel path has zero CI coverage (all test_deltaformer cases skip on the runner), there are no before/after numbers, and the +1e-9 NaN guard treats a symptom rather than the cause. The loop split also drops the boundary mask when (T-C)%BLOCK_T≠0 — the same chunk-straddle class of bug that #1105 later had to fix properly. The branch is conflicting with main, and the useful parts have been superseded by #1062 and #1105.

PR #611, opened 2025-10-14
feat: add dtest

garrett361:
Just adding the minimal example from the [dtest package here](https://github.com/garrett361/dtest). Will build out into full CP tests. 

## Summary by CodeRabbit

* **Documentation**
  * Clarified how to run distributed tests by noting they rely on an external test harness and pointing readers to its installation instructions.

* **Tests**
  * Added a simulated coarse-grained parallel sharding test that validates forward-pass correctness by comparing aggregated unsharded outputs to per-rank results.
  * Introduced a real distributed test that exercises an all-reduce operation across ranks to ensure correct multi-process behavior.

### DIFF
--- a/tests/ops/cp/README.md
+++ b/tests/ops/cp/README.md
@@ -1,0 +1,2 @@
+Distributed tests use the `DTest` class [from here](https://github.com/garrett361/dtest). Follow the
+README there to install.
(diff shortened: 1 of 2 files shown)


garrett361:
Made against the wrong PR target!

Issue #625, opened 2025-10-30
[Bug] Kimi LInear crashing at fla.kda.gate + Triton


Qubitium:
### Checklist

- [x] I have checked [FAQs](https://github.com/fla-org/flash-linear-attention/blob/main/FAQs.md) and existing issues for similar problems
- [x] Please report this bug in English to ensure wider understanding and support

### Describe the Bug

Not sure if this is Kimi modeling bug or something to do with fla.  It crashes during inference. 

<img width="1200" height="872" alt="Image" src="https://github.com/user-attachments/assets/b00bdaad-173c-456a-8c0a-8dc7df9ab0b0" />

<img width="1698" height="2050" alt="Image" src="https://github.com/user-attachments/assets/07c7762f-1c68-409b-9f40-55c11d4cbddc" />

### Steps to Reproduce the Bug

Run Kimi Linear model on HF with Nvidia A100

### Expected Behavior

Not crash.

### Environment Information

1. Torch: 2.9.0+cu130
2. Triton: 3.5.0

Instsalled flash linear via:

```
# uninstall both packages first to ensure a successful upgrade
pip uninstall fla-core flash-linear-attention -y && pip install -U git+https://github.com/fla-org/flash-linear-attention
```

zhiyuan1i:
Have you tried with `export CUDA_LAUNCH_BLOCKING=1`?   I will check this out. Since our CI only have H100.

Qubitium:
@zhiyuan1i  Here is the stacktrace with `CUDA_LAUNCH_BLOCKING=1`

```py
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../.cache/huggingface/modules/transformers_modules/Kimi_hyphen_Linear_hyphen_48B_hyphen_A3B_hyphen_Instruct/modeling_kimi.py:778: in forward
    hidden_states = self.self_attn(
../vm313t/lib/python3.13t/site-packages/torch/nn/modules/module.py:1775: in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../vm313t/lib/python3.13t/site-packages/torch/nn/modules/module.py:1786: in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #627, opened 2025-10-31
Fix kda.gate not enforcing contiguous mem layout

Qubitium:
Fix: https://github.com/fla-org/flash-linear-attention/issues/625

The `fla.kda.gate` triton kernel assumes the modeling framework will load the weights in contiguous memory. 

SGLang/vLLM may do this (I believe they almost always do this for performance reasons) but HF or other code that hook into the transformers modeling code which may move `weight` memory around due to offloading requirements may not. 

The pr simply ensures the conv1d weights/tensors are contiguous before launching triton kernel. There should be no performance impact as it is no-op if already contiguous. 

@yzhangcs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of causal 1D convolution by adding input validation to the forward path, reducing runtime errors and improving stability during convolution operations.
  * Enhances safety for edge-case inputs and lowers likelihood of invalid-input failures in streaming or sequence-processing workflows.


zhiyuan1i:
Thank you for helping us investigate, we already have a decorator and I'll help you fix it

zhiyuan1i:
It's very quirky, we have decorator @input_guard at the entrance, this one is to keep all the tensors continuous


https://github.com/fla-org/flash-linear-attention/blob/dbed6faf4e3e8eab709f4f4a47dcce0ed3833246/fla/modules/convolution.py#L648


https://github.com/fla-org/flash-linear-attention/blob/dbed6faf4e3e8eab709f4f4a47dcce0ed3833246/fla/utils.py#L135-L166

Qubitium:
> It's very quirky, we have decorator @input_guard at the entrance, this one is to keep all the tensors continuous
> 
> https://github.com/fla-org/flash-linear-attention/blob/dbed6faf4e3e8eab709f4f4a47dcce0ed3833246/fla/modules/convolution.py#L648
> 
> https://github.com/fla-org/flash-linear-attention/blob/dbed6faf4e3e8eab709f4f4a47dcce0ed3833246/fla/utils.py#L135-L166

Nice! Let me use that instead.

Qubitium:
@zhiyuan1i  Now using `@input_guard` helper. But I am a little confused after re-looking at the code. There appears to be @input_guard alredy existing at the entrance for calls to `def causal_conv1d_fwd()` why do I need to apply this fix to resolve crash?

sustcsonglin:
lgtm!

PR #641, opened 2025-11-14
[Perf] add chunk_indices parameter to avoid redundant computation

zhiyuan1i:
Add optional chunk_indices parameter across core operators to allow pre-computed chunk indices passed from upstream, avoiding redundant prepare_chunk_indices() calls.

Changes:
- Add chunk_indices parameter to convolution, token_shift, KDA/GLA ops
- Skip chunk indices computation when pre-computed ones are provided
- Replace cu_seqlens[1:] - cu_seqlens[:-1] with torch.diff() for clarity
- Update gradient computation to pass through chunk_indices

Affected modules:
- fla/modules/convolution.py, token_shift.py
- fla/ops/gla/chunk.py
- fla/ops/kda/*.py
- fla/ops/common/chunk_*.py
- fla/ops/utils/cumsum.py, index.py, solve_tril.py

This optimization is particularly beneficial for varlen training scenarios where chunk indices may be pre-computed and reused across multiple ops.

## Summary by CodeRabbit

* **New Features**
  * Support for supplying precomputed chunk indices across variable-length sequence ops, enabling consistent chunking across forward/backward passes and reducing redundant recomputation.

* **Improvements**
  * Functions now lazily derive chunk indices from existing sequence metadata when not provided, preserving backward compatibility.
  * Internal utilities updated to use standard PyTorch operations for consistency.

### DIFF
--- a/fla/ops/kda/chunk_inter.py
+++ b/fla/ops/kda/chunk_inter.py
@@ -152,10 +152,10 @@ def chunk_kda_bwd_dqkwg(
     chunk_indices: torch.LongTensor | None = None,
 ):
     B, T, H, K, V = *k.shape, v.shape[-1]
-    BT = min(chunk_size, max(16, triton.next_power_of_2(T)))
+    BT = chunk_size
 
     if chunk_indices is None and cu_seqlens is not None:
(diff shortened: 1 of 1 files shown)


sustcsonglin (approved):
lgtm!

PR #643, opened 2025-11-14
[KDA]: Improve intra-kernel efficiency

zhiyuan1i:
## Summary by CodeRabbit

* **Refactor**
  * Consolidated per-iteration logic into batched, vectorized operations for faster, more efficient computation.
  * Simplified tuning configurations to reduce complexity while preserving compatibility.
* **New Features**
  * Improved runtime kernel selection with a low‑resource variant to better support devices with limited shared memory.
* **Chores**
  * No changes to public APIs.

### DIFF
--- a/fla/ops/kda/chunk_intra.py
+++ b/fla/ops/kda/chunk_intra.py
@@ -114,7 +114,7 @@ def chunk_kda_fwd_kernel_intra_sub_inter(
     **autotune_cache_kwargs,
 )
 @triton.jit(do_not_specialize=['T'])
-def chunk_kda_fwd_kernel_intra_sub_inter_below_sm90(
+def chunk_kda_fwd_kernel_intra_sub_inter_low_resource(
     q,
     k,
     g,
@@ -539,11 +539,11 @@ def chunk_kda_fwd_intra(
     Akk = torch.zeros(B, T, H, BT, device=k.device, dtype=output_dtype)
     grid = (NT, NC * NC, B * H)
 
-    if check_shared_mem('hopper'):
+    if check_shared_mem('ampere'):
         chunk_kda_inter_kernel = chunk_kda_fwd_kernel_intra_sub_inter
         BK0 = BK
     else:
-        chunk_kda_inter_kernel = chunk_kda_fwd_kernel_intra_sub_inter_below_sm90
+        chunk_kda_inter_kernel = chunk_kda_fwd_kernel_intra_sub_inter_low_resource
         BK0 = 64
 
     chunk_kda_inter_kernel[grid](


fla/ops/kda/chunk_intra.py
```diff
@@ -69,30 +68,30 @@ def chunk_kda_fwd_kernel_intra_sub_inter(
 
     b_Aqk = tl.zeros([BC, BC], dtype=tl.float32)
     b_Akk = tl.zeros([BC, BC], dtype=tl.float32)
-    for i_k in range(tl.cdiv(K, BK)):
```
sustcsonglin, line 72: if K = 256, would it have out of shared memory issue?

zhiyuan1i: Test have been pasted on Hopper, since we lack ampere. Maybe we should keep BK=64 for A100 and Ada GPUs

fla/ops/kda/chunk_intra.py
```diff
@@ -160,21 +156,37 @@ def chunk_kda_fwd_kernel_intra_sub_intra(
...
-        p_gk += H*K
+    b_kt_block = tl.load(p_kt, boundary_check=(0, 1)).to(tl.float32)
+    b_gk_block = tl.load(p_gk, boundary_check=(0, 1)).to(tl.float32)
+
+    o_j = tl.arange(0, BC)
+    max_j = min(BC, T - i_t * BT - i_i * BC)
+    m_j = o_j < max_j
+
+    # [BC, BC, BK]
+    b_ktg = b_kt_block[None, :, :] * exp(b_g[:, None, :] - b_gk_block[None, :, :])
```
sustcsonglin: you should first mask out the upper triangular region of the `b_g[:, None, :] - b_gk_block[None, :, :]` scores. otherwise, for some i < j,
`exp(b_g[:, None, :] - b_gk_block[None, :, :])[i, j]` can become `nan`

fla/ops/kda/chunk_intra.py
```diff
@@ -160,21 +156,37 @@ def chunk_kda_fwd_kernel_intra_sub_intra(
...
+    m_Aqk = o_i[:, None, None] >= o_j[None, :, None]
+    m_Aqk = m_Aqk & m_j[None, :, None]
+    b_Aqk = tl.sum(b_q[:, None, :] * b_ktg * m_Aqk, -1) * scale
+    b_Aqk = tl.where(m_j[None, :], b_Aqk, 0.)
+
+    m_Akk = o_i[:, None, None] > o_j[None, :, None]
+    m_Akk = m_Akk & m_j[None, :, None]
+    b_Akk = tl.sum(b_k[:, None, :] * b_ktg * m_Akk, -1)
+    b_Akk = tl.where(m_j[None, :], b_Akk, 0.)
```
sustcsonglin, line 274: i’m a bit concerned about shared-memory usage on sm<90 cards, since this materializes a `[16, 16, 128]`-sized tensor, which isn’t small and could easily exceed the available shared memory for lower capacity card?

fla/ops/kda/chunk_intra.py
```diff
@@ -160,21 +156,37 @@ def chunk_kda_fwd_kernel_intra_sub_intra(
...
-        b_kt = tl.load(p_kt, mask=m_k, other=0).to(tl.float32)
-        b_gk = tl.load(p_gk, mask=m_k, other=0).to(tl.float32)
-        b_ktg = b_kt[None, :] * exp(b_g - b_gk[None, :])
-        b_Aqk = tl.sum(b_q * b_ktg, 1)
-        b_Aqk = tl.where(o_i >= j, b_Aqk * scale, 0.)
-        b_Akk = tl.sum(b_k * b_ktg, 1)
-        b_Akk = tl.where(o_i > j, b_Akk, 0.)
-        tl.store(Aqk + o_A + j, b_Aqk, mask=m_A)
-        tl.store(Akk + o_A + j, b_Akk, mask=m_A)
-        p_kt += H*K
```
sustcsonglin, line 176: i’m aware of the slowness in this part. could you provide profiling results for this change? also, can you update the GLA intra-chunk kernel accordingly?

sustcsonglin (changes requested):
fix nan. fix block size for low capacity GPU cards.

fla/ops/kda/chunk_intra.py
zhiyuan1i: @sustcsonglin  Thanks for your advice!! Fixed by tl.where

fla/ops/kda/chunk_intra.py
zhiyuan1i: Since maintaining BK=64 introduces unnecessary branch judgments and affects speed, I think we can still set BK=64 for cards below hopper and separate the conditions for a compile-time constant @sustcsonglin

zhiyuan1i:
> fix nan. fix block size for low capacity GPU cards.

Fixed.

PR #646, opened 2025-11-17
[Tril] Enable precision autotune[skip test]

zhiyuan1i:
- Fix critical bug where DOT_PRECISION_AUTOTUNE_LIST could be None when is_tma_supported=True due to .append() returning None
- Add DOT_PRECISION to autotune configs for dynamic precision selection
- Simplify precision validation by removing is_amd dependency
- Support tf32x3 precision mode in solve_tril operations

## Summary by CodeRabbit

* **Refactor**
  * Optimized precision handling logic with improved TMA support detection
  * Streamlined autotune configurations for enhanced performance


zhiyuan1i:
This PR doesn't change the logic of the default precision, as the environment variables need to be set manually

PR #663, opened 2025-11-25
fix: prevent kernel recompile

onenonlim:
## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized kernel compilation caching to reduce redundant recompilations and improve startup performance.
  * Enhanced autotuning configurations for normalization operations to better balance performance across hardware setups.
  * Improved kernel specialization handling to minimize compilation overhead while maintaining computational efficiency.

### DIFF
--- a/fla/modules/fused_norm_gate.py
+++ b/fla/modules/fused_norm_gate.py
@@ -25,10 +25,10 @@ from fla.utils import autotune_cache_kwargs, get_multiprocessor_count, input_gua
         for BT in [16, 32, 64]
         for num_warps in [4, 8, 16]
     ],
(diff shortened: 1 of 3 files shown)


gogongxt:
lgtm.  May someone can tell me why close this pr?

NJUpsj:
LGTM. I would like to know why this PR was closed.

In my use case, KDA is used and the sequence lengths within the same batch are highly uneven, with a very large span. For example:

`cu_seqlens = [0, 242877, 424059, 538877, 553376, 580196, 613727, ...]`


This pattern triggers a large number of autotuning runs, which makes training extremely slow (more than 3× slowdown in our experiments).

In our internal solution, we addressed this issue by modifying the `NB` parameters in the same three files to avoid excessive autotuning, and this approach has proven to be very effective.

In addition, when `use_short_conv=True` in KDA, there is also `NB` autotuning logic in `fla/modules/convolution.py`, which further triggers repeated autotuning and contributes significantly to the slowdown.

PR #669, opened 2025-11-28
[L2Norm] Avoid recompilation for variable-length inputs

retonym:
The `T` argument was previously marked as `tl.constexpr`, forcing recompilation whenever the sequence length changed. This PR fixes this issue to improve performance in variable-length training (e.g., SFT).
It also removes the `NB` argument, which was unused in the kernel.

## Summary by CodeRabbit

* **Refactor**
  * Adjusted layer-normalization kernel parameter handling so a previously compile-time parameter is now treated at runtime.
  * Prevented compile-time specialization on that parameter to simplify kernel invocation paths.
  * No changes to normalization results, behavior, or performance; purely internal maintenance improvements.

### DIFF
--- a/fla/ops/kda/chunk_intra.py
+++ b/fla/ops/kda/chunk_intra.py
@@ -17,6 +17,8 @@ else:
 # ============================================================================
 # Fused inter + solve_tril kernel: compute off-diagonal Akk and solve in one pass
 # ============================================================================
+
+
 @triton.heuristics({
     'IS_VARLEN': lambda args: args['cu_seqlens'] is not None,
 })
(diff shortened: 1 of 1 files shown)


zhiyuan1i:
LGTM. But I prefer to use Varlen and Autotune for better performance, and we found a 50% performance gap for different tune configurations

gogongxt (commented):
The BD `argument` also requires the input element size, so it may still trigger recompilation.

zhiyuan1i:
Maybe we can remove the recompile but keep autotune

retonym:
> The BD `argument` also requires the input element size, so it may still trigger recompilation.

Hi @gogongxt 
As I understand it, there are very few scenarios where a model's head dimension varies within the same training session.  We may not worry about that.

retonym:
> Maybe we can remove the recompile but keep autotune

Hi @zhiyuan1i , restore `NB` for autotuning & recompilation, only remove `T` tl.constexpr.

gogongxt:
> > The BD `argument` also requires the input element size, so it may still trigger recompilation.
> 
> Hi @gogongxt As I understand it, there are very few scenarios where a model's head dimension varies within the same training session. We may not worry about that.

In infra scenario, the input element size will vary.
```python
    MAX_FUSED_SIZE = 65536 // x.element_size()
    BD = min(MAX_FUSED_SIZE, triton.next_power_of_2(D))
```
Therefor the `BD` will change dynamically.

yzhangcs:
Thank you

PR #687, opened 2025-12-18
Temporary workaround to disable TritonGPUHoistTMEMAlloc in b_dk += tl.dot(tl.trans(b_dA), b_kb)

rucnyz:
Temporarily fix https://github.com/fla-org/flash-linear-attention/issues/638

In the original code `b_dk += tl.dot(tl.trans(b_dA), b_kb)`, the compiler fuses the addition into the dot op, generating IR like %136 (b_dk new) = tt.dot %134 (tl.trans(b_dA)), %127 (b_kb), %135 (b_dk_old), using %135 as the accumulator.
It seems the `TritonGPUHoistTMEMAlloc` pass identifies that %135 (input) and %136 (output) share the same TMEM buffer and attempts an optimization. However, on B200, it incorrectly calculates the dominance, moving the definition of %135 after its use in %136, causing the `operand #2 does not dominate this use` error.
The inline_asm forces the compiler to separate the dot and add instructions. And seems like only inline_asm will work and all other methods like + 0.0, * 1.0 and even bitcasts will still lead to the compiler optimize them away and re-enable the fusion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect dot-product fusion on NVIDIA Blackwell GPUs by adding a targeted runtime workaround to ensure correct math results.

* **New Features**
  * Added Blackwell device detection and a safe dot-product path that selects the appropriate implementation at runtime.

### DIFF
--- a/fla/ops/kda/chunk.py
+++ b/fla/ops/kda/chunk.py
@@ -4,9 +4,8 @@ import torch
 
 from fla.modules.l2norm import l2norm_bwd, l2norm_fwd
 from fla.ops.common.chunk_delta_h import chunk_gated_delta_rule_bwd_dhu, chunk_gated_delta_rule_fwd_h
-from fla.ops.common.chunk_o import chunk_bwd_dv_local
-from fla.ops.gla.chunk import chunk_gla_bwd_dA, chunk_gla_fwd_o_gk
-from fla.ops.kda.chunk_inter import chunk_kda_bwd_dqkwg
+from fla.ops.gla.chunk import chunk_gla_fwd_o_gk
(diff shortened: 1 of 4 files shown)


zhiyuan1i:
Thanks for the contribution, but it looks like this PR will break the FLA repository's general support for GPUs like AMD. Can you feed back to triton-lang/triton at the same time?

yzhangcs:
@rucnyz could you refactor this into a small helper function? That way, we can conditionally use inline assembly based on the device type while keeping the main logic clean.

zhiyuan1i:
LGTM.

wilsonyqm:
seems like the fix is being deleted in #692 https://github.com/fla-org/flash-linear-attention/pull/692/files

might need to resend the fix.

PR #701, opened 2025-12-29
[KDA] Add lowerbound gate function

zhiyuan1i:
Reason:

- When using TensorCore with tile size M=16 for cumulative sum (cumsum) operations, the FP32 exp() function risks overflow when inputs exceed [-85, 85]. A cumsum across 16 tiles can easily violate this range.
- This PR adds a lower_bound parameter to constrain each tile's minimum value, ensuring the accumulated result stays within safe bounds. By limiting each tile to ≥ -5.0(for example), the worst-case accumulation (-5.0 × 16 = -80) remains safely within the [-85, 85] non-overflow zone for exp().
- Default behavior unchanged (lower_bound=None): Uses original `naive_kda_gate` (not recommended)
- With safety bound (lower_bound=-5.0): Uses `naive_kda_lowerbound_gate` with validated range [-5.0, 0.0)
- Recommendation: Set lower_bound=-5.0 for all TensorCore-based to prevent numerical instability

## Summary by CodeRabbit

* **New Features**
  * Added configurable lower-bound gating option for KDA operations and exposed it in fused gate interfaces
  * Added optimized chunked cumsum path with optional lower-bound behavior for varied sequence shapes

* **Bug Fixes**
  * Adjusted backward-path gradient handling for chunked gate computations to correct bias gradient flow

* **Tests**
  * Added tests covering the lower-bound gate variant across configurations

### DIFF
--- a/fla/ops/kda/gate.py
+++ b/fla/ops/kda/gate.py
@@ -265,9 +265,9 @@ def kda_gate_bwd(
         lower_bound=lower_bound,
     )
 
-    dg = dg.view_as(g)
-    dA = dA.sum(0).view_as(A_log)
-    dbias = dg.view(-1, H * K).sum(0) if dt_bias is not None else None
+    dg = dg.view_as(g).type_as(g)
+    dA = dA.sum(0).view_as(A_log).type_as(A_log)
+    dbias = dg.view(-1, H * K).sum(0).to(dt_bias) if dt_bias is not None else None
 
     return dg, dA, dbias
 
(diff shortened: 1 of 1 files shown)


fla/ops/kda/gate.py
```diff
@@ -287,4 +334,134 @@ def fused_kda_gate(
...
+    output_dtype: torch.dtype | None = torch.float,
+    chunk_indices: torch.LongTensor | None = None,
+    lower_bound: float | None = None,
+    **kwargs,
+) -> torch.Tensor:
+    if cu_seqlens is not None:
+        assert g.shape[0] == 1, "Only batch size 1 is supported when cu_seqlens are provided"
+    assert len(g.shape) == 4
+
+    if head_first:
```
yzhangcs, line 438: we should remove all deprecated head_first args for the new fn

PR #703, opened 2025-12-30
[KDA] Speed up chunk_kda by introducing lowerbound gate

zhiyuan1i:
After profiling, we found that the original chunk_kda kernel is heavily bound by CUDA-core computation. To maintain numerical stability along the diagonal of the (causal-mask) attention matrix we had to split each 64-element chunk a second time—running a 16-element tile on Tensor Cores and looping over the remainder with CUDA cores. This introduced enormous overhead: in Triton those loops are hard to optimize, almost impossible to overlap, and, given the fixed FLOP/byte ratio of the hardware, inevitably slower.

By switching the activation function we enabled pure Tensor-Core execution with M = 16, which delivered a large end-to-end speed-up. To regain extra precision—and to keep the option of M = 32 Tensor Cores—we added a median shift to the log-domain cumulative-sum, reducing the accuracy loss at M = 16 while still allowing M = 32. Further Triton tweaks (e.g., transposing with M = 64 Tensor Cores) have so far yielded only marginal gains. At this point chunk_kda is a purely memory-bound kernel.

## Summary by CodeRabbit

* **New Features**
  * Added runtime controls for gating and recompute (safe_gate, disable_recompute, lower_bound, cu_seqlens) and exposed intermediate outputs to improve memory/recompute handling.

* **Bug Fixes**
  * Improved numeric stability and masking for variable-length sequences; standardized explicit float32 casts and TF32 precision handling; added explicit kernel barrier for correctness.

* **Tests**
  * Expanded tests to cover new gating/recompute options and kernel vs. naive gating paths.

* **Chores**
  * Benchmarks updated to exercise the new gating option.

### DIFF
--- a/fla/ops/common/chunk_delta_h.py
+++ b/fla/ops/common/chunk_delta_h.py
@@ -144,7 +144,7 @@ def chunk_gated_delta_rule_fwd_kernel_h_blockdim64(
         last_idx = min((i_t + 1) * BT, T) - 1
         if USE_G:
             m_t = (i_t * BT + tl.arange(0, BT)) < T
-            b_g_last = tl.load(g + bos * H + last_idx * H + i_h)
+            b_g_last = tl.load(g + bos * H + last_idx * H + i_h).to(tl.float32)
             p_g = tl.make_block_ptr(g + bos * H + i_h, (T,), (H,), (i_t * BT,), (BT,), (0,))
             b_g = tl.load(p_g, boundary_check=(0,)).to(tl.float32)
             if USE_EXP2:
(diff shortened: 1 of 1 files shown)


Nathancgy (approved):

wfloveiu:
Good optimization. Here is a question, in  `chunk_intra.py`: `if not USE_SAFE_GATE:`，if set safe_gate=True，here dosn't do forward substitution on diagonals，and directly compute off-diagonals subblocks.  So where do you do the inverse of diagonals?

PR #715, opened 2026-01-18
[Fix][GDN] convert b_q dtype in bwd_dhu kernel when USE_G is enabled

slowlyC:
## Summary

When `USE_G` is enabled in `chunk_gated_delta_rule_bwd_kernel_dhu_blockdim64`, multiplying `b_q` by `b_g_exp[None, :]` causes the data type to change from the original dtype (e.g., bfloat16) to float32 (b_g's dtype).

This PR adds `.to(b_q.dtype)` to preserve the original data type after the multiplication, ensuring correct dtype in the subsequent `tl.dot` operations.

## Changes

- Add `.to(b_q.dtype)` after `b_q * b_g_exp[None, :]` for all K dimension blocks (b_dh1-b_dh4)
- Remove redundant `.to(b_q.dtype)` in `tl.dot` calls since b_q now has the correct dtype

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data type consistency in mixed-precision operations to reduce potential computation errors.

### DIFF
--- a/fla/ops/common/chunk_delta_h.py
+++ b/fla/ops/common/chunk_delta_h.py
@@ -397,13 +397,13 @@ def chunk_gated_delta_rule_bwd_kernel_dhu_blockdim64(
         b_q = tl.load(p_q, boundary_check=(0, 1))
         if USE_G:
             b_dh1 *= bg_last_exp
-            b_q = b_q * b_g_exp[None, :]
+            b_q = (b_q * b_g_exp[None, :]).to(b_q.dtype)
         if USE_GK:
             if USE_EXP2:
                 b_dh1 *= exp2(b_gk_last1[:, None])
             else:
                 b_dh1 *= exp(b_gk_last1[:, None])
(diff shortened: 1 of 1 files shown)


zhiyuan1i:
Thanks for the contribution! @slowlyC 
However, I'd like to keep tf32 precision for the dh backward matmul. 
The tf32 usage here was intentional: b_q is typically L2-normalized with values in the 0-1 range, and when multiplied by small gate values, bf16 precision becomes insufficient.
I've compared the results and observed that dh0 error increases by ~17% (diff) and ~31% (ratio) with this change. This is particularly concerning because errors in dh accumulate across chunks during the backward pass - each chunk's gradient computation depends on the dh state from subsequent chunks.
This becomes even more critical in context parallelism (CP) for ultra-long sequences, where earlier ranks depend on dh0 from later ranks to compute their gradients. The precision loss would propagate and accumulate across all CP ranks, potentially affecting training stability at scale.
KDA-related ops also use tf32 matmul for the same reason. Unless there's a significant performance bottleneck here, I'd prefer to keep the current precision.



**The test is only tested on 2048 tokens.** CC @yzhangcs

PR #739, opened 2026-02-08
[MAMBA2] fix initialization for mamba2

mayank31398:
The init in FLA repo for mamba2 is incorrect.
This PR fixes the issue.
After the fix, Mamba2 outperforms GDN at 7B MoE scale (1B active params).
The difference between wrong and fixed init is significant.

Training curves: https://wandb.ai/mayank31398/mamba-test?nw=nwusermayank31398&panelDisplayName=train/lm_loss&panelSectionName=train

## Summary by CodeRabbit

* **Refactor**
  * Model initialization now uses stochastic per-head discretization parameters instead of a fixed deterministic sequence, increasing initialization variability across heads.
  * This is an internal change only: public interfaces, control flow, and external behavior are unchanged. May slightly affect training/run-to-run variability but requires no action from users.

### DIFF
--- a/fla/layers/mamba2.py
+++ b/fla/layers/mamba2.py
@@ -173,7 +173,11 @@ class Mamba2(nn.Module):
         # instantiate once and copy inv_dt in init_weights of PretrainedModel
         # hard coded for now
         dt_init_floor = 1e-4
-        dt = torch.exp(torch.rand(self.num_heads) * (math.log(self.time_step_max) - math.log(self.time_step_min)) + math.log(self.time_step_min))
+        dt = torch.exp(
+            torch.rand(self.num_heads) * (
+                math.log(self.time_step_max) - math.log(self.time_step_min)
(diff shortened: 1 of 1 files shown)


mayank31398:
One big concern: if a user is training any model with FSDP-2, since all tensors are DTensors, it will skip initialization according to the FLA logic.
I dont know how to fix this.

mayank31398:
AFAIK this PR still doesnt fix if the meta init is done with DTensors. I am not sure how to fix that

sustcsonglin:
> One big concern: if a user is training any model with FSDP-2, since all tensors are DTensors, it will skip initialization according to the FLA logic. I dont know how to fix this.

@yzhangcs i remembered you fixed this before. any thoughts?

sustcsonglin:
@mayank31398 thanks for ur pr! could you fix lint errors?

PR #750, opened 2026-02-18
[Misc] remove redundant dot precision param in KDA recompute_w_u

KevinZeng08:
Thanks for the wonderful work about linear attention algorithm and implementation, which has been widely used in frontier models.

In KDA's `recompute_w_u_fwd_kda_kernel`, it is a little confusing about the DOT_PRECISION input parameter because the input Akk and V are both always fp16 or bf16.
https://github.com/fla-org/flash-linear-attention/blob/7ef6685c08e91400d222f10f9e16cf0187a52933/fla/ops/kda/wy_fast.py#L72
In Triton, if inputs are not in dtype f32, this option is ignored. So we can safely remove it.
https://github.com/triton-lang/triton/blob/5ce5c68c2a3122a61c282058d652769254637b27/python/triton/language/core.py#L2120 

## Summary by CodeRabbit

* **Refactor**
  * Simplified numerical-precision handling inside performance-critical kernels by removing explicit precision toggles and streamlining autotuning configuration; maintains existing behavior while reducing internal complexity and potential configuration overhead.

### DIFF
--- a/fla/ops/kda/wy_fast.py
+++ b/fla/ops/kda/wy_fast.py
@@ -6,7 +6,7 @@ import triton.language as tl
 
 from fla.ops.utils import prepare_chunk_indices
 from fla.ops.utils.op import exp2
-from fla.utils import IS_TF32_SUPPORTED, autotune_cache_kwargs, check_shared_mem
+from fla.utils import autotune_cache_kwargs, check_shared_mem
 
 
 @triton.heuristics({


zhiyuan1i:
Yes, this is a legacy issue. We initially used fp32 to store Akk, but then we migrated to lower precision storage.

zhiyuan1i (approved):

PR #773, opened 2026-03-08
[GDN] Enhance Triton 3.2 compatibility

winglet0996:
The fix avoids performing broadcast multiplication directly on the encoded tensor of the transposed `dot_op`. Instead, we first scale the original `b_k` by `beta` and then transpose it before passing it to `tl.dot(...)`, thereby circumventing the encoding incompatibility issue in Triton 3.2.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal computation handling for weighted backpropagation paths to optimize performance and numerical stability.

### DIFF
--- a/fla/modules/conv/cp/ops.py
+++ b/fla/modules/conv/cp/ops.py
@@ -72,6 +72,7 @@ class CausalConv1dFunctionCP(torch.autograd.Function):
         W: int,
         group: dist.ProcessGroup | None,
         is_first_rank: bool,
+        pre_num_conv_tokens: int = 0,
     ) -> None:
         """Correct dx gradients for CP backward pass by communicating with next rank.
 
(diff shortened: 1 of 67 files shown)


zhiyuan1i:
Thank you for your PR, we had noticed this before. However, whether to officially support triton3.2 is still worth discussing, because in addition to compilation errors, triton3.2 also has some hidden precision errors that were fixed in later versions. Officially supporting triton3.2 would require us to pay more costs, running CI and dedicating manpower to fix issues. Given the current development of triton, I recommend the latest release version, just like the torch version that the FLA repository depends on, we only support 3.3 and subsequent versions. Looking forward to more comments from maintainers

winglet0996:
Thank you for your reply and the detailed explanation!
The main motivation behind this PR is to accommodate legacy environments. On some older computing clusters (such as CentOS 7), it is not feasible to install higher versions of Triton due to  reasons like glibc version constraints / system compiler limitations. Therefore, we proposed this PR to enable support for Triton 3.2 in these specific cases.
However, I completely understand your considerations. Prioritizing stability and avoiding hidden precision errors is indeed crucial, and the maintenance cost is a valid concern. I respect the team's decision to focus on newer versions for official support.

zhiyuan1i (approved):

PR #806, opened 2026-03-31
[L2Norm] Fix bf16 numerical stability by calculating norm in f32 then normalising in input dtype

tmct:
We were witnessing numerical instability in downstream operations like the matrix inversion in gated delta rule - the output tensor didn't have norm close enough to 1 bf16.... This change removed that instability.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced numerical stability in L2 normalization computations to deliver more accurate and consistent results across different input scenarios.

### DIFF
--- a/fla/modules/l2norm.py
+++ b/fla/modules/l2norm.py
@@ -32,9 +32,11 @@ def l2norm_fwd_kernel1(
     cols = tl.arange(0, BD)
     mask = cols < D
 
-    b_x = tl.load(x + cols, mask=mask, other=0.0).to(tl.float32)
-    b_rstd = 1 / tl.sqrt(tl.sum(b_x * b_x) + eps)
-    b_y = b_x * b_rstd
+    # compute norm in f32 for stability
+    b_x = tl.load(x + cols, mask=mask, other=0.0)
+    b_x_f32 = b_x.to(tl.float32)
+    b_rstd = tl.rsqrt(tl.sum(b_x_f32 * b_x_f32) + eps)
+    b_y = b_x * b_rstd.to(b_x.dtype)
     tl.store(y + cols, b_y, mask=mask)
     tl.store(rstd + i_t, b_rstd)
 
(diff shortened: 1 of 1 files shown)


sustcsonglin:
i thought we already convert input to fp32?     

`b_x = tl.load(x + cols, mask=mask, other=0.0).to(tl.float32)`

tmct:
We did, but I need to hold onto the half precision value before we do that now.

sustcsonglin:
could you explain why only using fp32 to calculate the `b_rstd` and using bf16 to do rescale is more numerical stable? it's not obvious to me

sustcsonglin:
could you attach a corner-case test such that the original implementation has numerical issue while your PR can fix?

sustcsonglin (approved):
look good to me but i don't fully understand the mechanism now

yzhangcs:
Hi @tmct, thanks for the PR! I'm still a bit confused about the mechanism here.

As I understand it, the change keeps `b_x` in bf16 and only uses f32 for computing the norm (`b_rstd`). But then `b_rstd` is cast back to bf16 before multiplying:

```python
b_y = b_x * b_rstd.to(b_x.dtype)  # bf16 * bf16
```

My question is: **why not keep `b_rstd` in f32 for the multiplication?** Something like:

```python
b_y = b_x.to(tl.float32) * b_rstd  # f32 * f32, then cast back when storing
```

Intuitively, keeping `b_rstd` in f32 should preserve more precision. Could you explain why casting `b_rstd` to bf16 before multiplication actually improves numerical stability in your case?

Thanks!

tmct:
Thanks for your comments and review. I'm going to try to dig into exactly why this is helping in my use case. My understanding was that conversions from something exactly norm-1 in f32 to bf16 resulted in something not close enough to norm-1 in bf16, but that's more of a hunch and you raise a good point so I should try and understand more deeply why empirically this fixes it for me!

PR #839, opened 2026-04-16
[Fix] flatten batched qkv in varlen cu_seqlens path

lxr-tech:
1. **Problem**: squeeze(0) does not flatten [B, L, H, D] when B>1, causing shape/semantics mismatch in flash_attn_varlen_func with cu_seqlens (see https://github.com/fla-org/flash-linear-attention/blob/main/fla/layers/attn.py#L158). 
2. **Fix**: flatten q/k/v to packed token layout before varlen call, then reshape back.
3. **Test**: add B>1 + cu_seqlens regression test in tests/layers/test_attn_varlen_pack_layout.py.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected variable-length attention for batched inputs so outputs are properly reshaped and stable when using packed/batched layouts.

* **Tests**
  * Added tests validating batched variable-length attention across multiple sizes, ensuring outputs have correct shapes and contain finite values.


yzhangcs:
@lxr-tech Hello, could you fix the lint check first.

lxr-tech:
Hi @yzhangcs, I have debugged and fixed this a few times, and now all checks are passing.

Sorry for the trouble, and thanks for your patience.

PR #840, opened 2026-04-16
[MoBA] Integrate MOBA and FlashMOBA

ReyJerry:
This pull request introduces two state-of-the-art sparse attention modules—`MobaAttention` and `FlashMoBA`—to the FLA codebase, along with the necessary integration and registration steps.

* **Background**: MoBA (Mixture-of-Block Attention) originates from MoonshotAI [MoonshotAI/MoBA](https://github.com/MoonshotAI/MoBA), while `FlashMoBA` is its highly optimized CUDA implementation developed by the MIT HAN Lab [mit-han-lab/flash-moba](https://github.com/mit-han-lab/flash-moba).
* **Motivation**: Considering both are typical and highly advanced sparse attention mechanisms, I have refactored and wrapped them into an FLA-compatible format. The goal is to enrich the scope of models supported by FLA and provide the open-source community with a unified, convenient interface to easily utilize MoBA and Flash-MoBA.
* **Technical Integration**:
    * Added the `MobaAttention` module implementation in `fla/layers/moba.py`, which provides a native mixture of block attention mechanism, supporting chunking and output gating.
    * Added the `FlashMoBA` attention module implementation in `fla/layers/moba.py`, supporting efficient mixture of block attention.
    * Introduced the `moba_attn_varlen` function (and its naive counterpart) in `fla/ops/moba/` as the core operator for the module.
    * Registered `MobaAttention` in the main package imports and `__all__` lists in `fla/__init__.py` and `fla/layers/__init__.py`.

**Test plan:**

* **Framework Compatibility**: Verified that both `MobaAttention` and `FlashMoBA` are fully compatible with the FLAME training framework.
* **Evaluation Compatibility**: Successfully tested compatibility with HuggingFace's `lm-evaluation-harness` for downstream task evaluations.

**Breaking changes:**

* None. This PR strictly adds new modules and operators without modifying existing public APIs.[](url)

## Summary by CodeRabbit

* **New Features**
  * Two new attention layers (FlashMoBA, MobaAttention) added to the public API.
  * Mixture-of-block attention with configurable chunking and top‑k selection for selective attention.
  * Efficient variable‑length attention with caching support, rotary embeddings, and optimized native ops for faster execution and lower memory.


pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
     "torch>=2.7.0",
     "transformers",
     "einops",
+    "flash-attn>=2.7.0",
```
yzhangcs: I don't think we should take flash-attn as a hard dependency. It would be better to defer the check and only raise an error when the user actually enables the fast path during MoBA layer initialization.

fla/layers/__init__.py
```diff
@@ -12,6 +12,7 @@
 from .comba import Comba
 from .delta_net import DeltaNet
 from .deltaformer import DeltaFormerAttention
+from .flash_moba import FlashMoBA
```
yzhangcs: simply moba would be okay imo

pyproject.toml
ReyJerry: Thanks for the feedback! I completely agree. I have removed the global try...except block and the raise ImportError inside the __init__ method for the MoBA layers. The flash-attn check is now deferred, so the layer initialization no longer takes it as a hard dependency. The updates have been pushed. Please let me know if there's anything else needed!

fla/layers/__init__.py
ReyJerry: Thanks for the suggestion! That makes perfect sense, and I totally agree. I have removed the separate FlashMoBA layer and unified the interface. Now there is only a single MobaAttention class exposed in fla.layers. Under the hood, it seamlessly routes to the highly optimized flash_moba CUDA operator if initialized with use_flash_moba=True, and defaults to the native Triton implementation otherwise. The __init__.py imports have also been updated to reflect this clean API. The changes have been pushed. Please let me know if this looks good to you!

yzhangcs (approved):

zhiyuan1i (approved):

PR #869, opened 2026-04-29
[KDA][AMD]for kda kernel,fix core dump on AMD GPU and tune the config for AMD branch

binding7012:
Original code run on AMD gpu would core dump. this pr fix it, and also tune the config for AMD gpu


binding7012:
Hi @yzhangcs  pls help to review this PR, thanks~

zhiyuan1i:
Thanks for the PR! ROCm support is definitely on our radar. We're currently working on a more unified backend abstraction design to avoid scattering hardware-specific branches across kernels. We'd prefer to land KDA's AMD support through that framework rather than merging this directly. Let's keep this open as a reference for now?

PR #895, opened 2026-05-13
[Enhance] Extend tensor_cache to a bounded queue

HAOCHENYE:
# Motivation

When using FLA for training, I noticed that `tensor_cache` frequently caused cache misses. This led to significant lag in the CPU timeline during training, thereby affecting the GPU’s kernel execution.

It was later discovered that this issue stemmed from a conflict between Domino EP’s parallel processing strategy and the `tensor_cache` logic of FLA.

If each forward pass of the model processes only one batch of data, then the meta-information for all data in that batch remains the same. As a result, multiple decoder layers can share the same cache.

```python

for layer in layers:
    layer(
        hidden_states,   # The shape of hidden_states is exactly the same.
        meta_info,          # Meta-information of the samples, such as cu_seq_lens, max_length, etc.
    )

# The cache can be hit.
```

But solutions like DominoEP process multiple batches in a single forward operation.

```python
for layer in layers:
    for hidden_states in hidden_states_list:
        hidden_states = layer(
            hidden_states,   # The shape of each hidden_states is different.
            meta_info,    # The meta information also varies.
         )
....
```
The for loop in memory causes cache misses every time it’s executed.

# Modifications

A new concept, `FLA_TENSOR_CACHE_SIZE`, has been added, enabling caching of multiple inputs.

### DIFF
--- a/fla/utils.py
+++ b/fla/utils.py
@@ -140,14 +140,14 @@ def tensor_cache(
         Callable[..., torch.Tensor]:
             A wrapped version of ``fn`` backed by an identity-based bounded cache.
     """
-    last_args: deque = deque(maxlen=FLA_TENSOR_CACHE_SIZE)
+    cached: deque = deque(maxlen=FLA_TENSOR_CACHE_SIZE)
 
     @functools.wraps(fn)
     def wrapper(*args: Any, **kwargs: Any) -> Any:
         if FLA_DISABLE_TENSOR_CACHE:
             return fn(*args, **kwargs)
 
-        for cached_args, cached_kwargs, cached_result in last_args:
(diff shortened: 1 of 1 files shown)


HAOCHENYE:
@gemini

yzhangcs (approved):

yzhangcs (commented):
Please rename the variable `last_args` to `cached` — it now stores `(args, kwargs, result)` triples, so `last_args` is misleading.

```suggestion
    cached: deque = deque(maxlen=FLA_TENSOR_CACHE_SIZE)
```

And update the loop accordingly:

```python
for cached_args, cached_kwargs, cached_result in cached:
```

and:

```python
cached.append((args, kwargs, result))
```

PR #909, opened 2026-05-20
Make varlen index helpers torch.compile-friendly

kasper0406:
### Problem

`prepare_chunk_indices`, `prepare_position_ids`, `prepare_sequence_ids` and `prepare_token_indices` built their per-segment `arange`s via `.tolist()` + a Python list comprehension over `cu_seqlens`. That **graph-breaks under Dynamo** (recompiling every batch as segment count / lengths shift) and dispatches **`O(segments)` ATen ops** in eager mode.

### Change

A shared `_segmented_arange` helper — one `repeat_interleave` + `cumsum` + `arange` + gather:

```
seg_id    = repeat_interleave(arange(n_segs), counts)
seg_start = pad(cumsum(counts))[:-1]
intra_idx = arange(total) - seg_start[seg_id]
```

`prepare_chunk_indices` feeds it `ceil(len / chunk_size)`; `prepare_position_ids` feeds it raw lengths; `prepare_sequence_ids` / `prepare_token_indices` become compile-clean transitively. Output is **bit-identical** for every valid `cu_seqlens` (segment length ≥ 1).

### Compile behaviour

Verified under `torch.compile(fullgraph=True, dynamic=True)` by counting Dynamo's graphs:
- **No graph break** — traces into one graph per helper.
- **Multi-segment** inputs reuse a **single graph: zero recompiles**.
- A **single-segment** input is one *additional* graph (size-1 `counts` hits PyTorch's 0/1 dynamic-shape specialization). Bounded at **2 graphs total**, not a per-batch recompile.

### Device sync in the CUDA varlen case

`prepare_chunk_indices` has a data-dependent output shape, so `repeat_interleave` with no `output_size=` reads `counts.sum()` host-side to allocate — one device→host sync on CUDA.

### Tests

- **`tests/ops/utils/test_index_compile.py`** (new): equivalence + compile contract for all four helpers; asserts the distinct-graph count directly (1 / 1 / 2), so a recompile regression fails the test.
- **`tests/ops/test_index.py`**: existing correctness sweep still passes.


yzhangcs:
Hi, I think this is a useful cleanup for Dynamo capture, but I’m not sure the PR fully addresses the CUDA varlen case.
The old `.tolist()` path was definitely a graph break, so replacing it with tensor ops makes the helpers more traceable. 

However, for `prepare_chunk_indices`, the output shape is still data-dependent. 
It needs to return a compact tensor with shape `[sum(ceil(lens / chunk_size)), 2]`, and if `cu_seqlens` is on CUDA, that output size is data-dependent, so the host still has to know `sum(chunk_counts)` to allocate it. 
`repeat_interleave(counts)` without a host-known `output_size` can still sync for exactly this reason.

kasper0406:
@yzhangcs correct, the output size is data-dependent, so `repeat_interleave` without `output_size` keeps one runtime D2H sync. This PR doesn't remove that, it only removes the graph break and per-batch recompile (the size is now an unbacked SymInt instead of .tolist()).

There's a `cu_seqlens_cpu` parameter that could avoid the sync by doing the sizing host-side, but I haven't worked through whether that's sync-free end to end across the call sites. I can look into it as a follow-up.

yzhangcs:
LGTM — but let's land it for the right reason.

The eager win is real and that's the case for merging. The rewrite replaces the `O(segments)` `.tolist()` + per-segment `arange` loop with a fixed handful of ops, so cost is flat in segment count instead of linear. Crossover is ~8–16 segments, which packed 8k/16k pretraining clears comfortably, and the output stays bit-identical. Raw per-call numbers (`chunk_size=64`, CUDA-event timing, `tensor_cache` bypassed):

```
prepare_chunk_indices [ragged]        prepare_position_ids [ragged]
 n_seg    old(us)   new(us) speedup     n_seg    old(us)   new(us) speedup
     8     256.84   304.80    0.84x         8     248.67   247.40    1.01x
    16     324.88   307.76    1.06x        16     457.81   245.41    1.87x
    32     461.30   305.68    1.51x        32     871.32   248.96    3.50x
    64     732.44   310.15    2.36x        64    1676.44   247.49    6.77x
   128    1276.00   305.57    4.18x       128    3303.42   247.50   13.35x
  1024    8891.93   315.34   28.20x      1024   25901.40   251.16  103.13x
```

Cost tracks segment count, not the length distribution (uniform inputs give the same crossover). If anything the isolated loop flatters the old path — in a real step those per-segment launches also clog the launch queue and bubble the main stream.

Where I'd push back is the "compile-friendly / no graph break" framing. `repeat_interleave(x, counts)` is itself a data-dependent output-shape op, so under `fullgraph=True` it graph-breaks unless `capture_dynamic_output_shape_ops=True` — which the PR doesn't set. The new `test_index_compile.py` is red as committed (12/16 fail on `aten.repeat_interleave.Tensor's output shape depends on input data`); the size only becomes an unbacked SymInt with that flag on. So this trades the `.tolist()` break for a `repeat_interleave` one rather than removing it.

Two asks before merge:
- Reframe the description around the eager win (flat in segment count, fewer launches), and either set the capture flag in the tests with a note on its cost, or drop the compile asserts.
- Follow-up to close out the D2H sync: pass `output_size=` to `repeat_interleave`, computed host-side from `cu_seqlens_cpu`.

<details><summary>Benchmark script</summary>

```python
import argparse

import torch
import triton

from fla.ops.utils.index import _segmented_arange, prepare_lens


def old_chunk_indices(cu_seqlens, chunk_size):
    indices = torch.cat([
        torch.arange(n, device=cu_seqlens.device)
        for n in triton.cdiv(prepare_lens(cu_seqlens), chunk_size).tolist()
    ])
    return torch.stack([indices.eq(0).cumsum(0) - 1, indices], 1).to(cu_seqlens)


def new_chunk_indices(cu_seqlens, chunk_size):
    chunk_counts = (prepare_lens(cu_seqlens) + (chunk_size - 1)).div(chunk_size, rounding_mode='floor')
    seg_id, intra_chunk_idx = _segmented_arange(chunk_counts)
    return torch.stack([seg_id, intra_chunk_idx], 1).to(cu_seqlens)


def old_position_ids(cu_seqlens, chunk_size=1):
    return torch.cat([
        torch.arange(n, dtype=cu_seqlens.dtype, device=cu_seqlens.device)
        for n in prepare_lens(cu_seqlens).unbind()
    ])


def new_position_ids(cu_seqlens, chunk_size=1):
    _, position_ids = _segmented_arange(prepare_lens(cu_seqlens))
    return position_ids.to(cu_seqlens)


def make_cu_seqlens(n_seg, mode, device, seed=0):
    g = torch.Generator().manual_seed(seed)
    if mode == 'uniform':
        lens = torch.full((n_seg,), 512, dtype=torch.int64)
    elif mode == 'ragged':
        lens = torch.randint(64, 4096, (n_seg,), generator=g, dtype=torch.int64)
    else:
        raise ValueError(mode)
    cu = torch.cat([torch.zeros(1, dtype=torch.int64), lens.cumsum(0)])
    return cu.to(device=device, dtype=torch.int32)


def time_us(fn, cu, chunk_size, iters, warmup=20):
    for _ in range(warmup):
        fn(cu, chunk_size)
    torch.cuda.synchronize()
    start = torch.cuda.Event(enable_timing=True)
    end = torch.cuda.Event(enable_timing=True)
    start.record()
    for _ in range(iters):
        fn(cu, chunk_size)
    end.record()
    torch.cuda.synchronize()
    return start.elapsed_time(end) / iters * 1e3


def run_eager(args, device):
    pairs = [
        ('prepare_chunk_indices', old_chunk_indices, new_chunk_indices),
        ('prepare_position_ids', old_position_ids, new_position_ids),
    ]
    for name, old_fn, new_fn in pairs:
        for mode in ('uniform', 'ragged'):
            print(f'\n=== {name}  [{mode} lengths]  chunk_size={args.chunk_size} ===')
            print(f'{"n_seg":>7} {"old (us)":>12} {"new (us)":>12} {"speedup":>9}  {"match":>6}')
            for n_seg in [1, 2, 4, 8, 16, 32, 64, 128, 256, 1024]:
                cu = make_cu_seqlens(n_seg, mode, device)
                ok = torch.equal(old_fn(cu, args.chunk_size), new_fn(cu, args.chunk_size))
                o = time_us(old_fn, cu, args.chunk_size, args.iters)
                n = time_us(new_fn, cu, args.chunk_size, args.iters)
                print(f'{n_seg:>7} {o:>12.2f} {n:>12.2f} {o/n:>8.2f}x  {str(ok):>6}')


def run_compile(args, device):
    import torch._dynamo
    from torch._dynamo.utils import counters
    shapes = [make_cu_seqlens(n, 'ragged', device, seed=i) for i, n in enumerate([4, 8, 16, 32, 64])]
    for label, capture in [('default config', False), ('capture_dynamic_output_shape_ops=True', True)]:
        print(f'\n=== compile: new_chunk_indices  ({label}) ===')
        torch._dynamo.config.capture_dynamic_output_shape_ops = capture
        torch._dynamo.reset()
        counters.clear()
        compiled = torch.compile(lambda cu: new_chunk_indices(cu, args.chunk_size), fullgraph=True, dynamic=True)
        try:
            for cu in shapes:
                assert torch.equal(compiled(cu), new_chunk_indices(cu, args.chunk_size))
            print(f"  fullgraph trace OK, unique_graphs={counters['stats'].get('unique_graphs', 0)} over {len(shapes)} shapes")
        except Exception as e:
            print(f'  GRAPH BREAK / FAILURE: {type(e).__name__}: {str(e).splitlines()[0]}')
    torch._dynamo.config.capture_dynamic_output_shape_ops = False


def main():
    p = argparse.ArgumentParser()
    p.add_argument('--chunk-size', type=int, default=16)
    p.add_argument('--iters', type=int, default=100)
    p.add_argument('--skip-compile', action='store_true')
    args = p.parse_args()
    if not torch.cuda.is_available():
        raise SystemExit('This benchmark requires a CUDA device.')
    device = torch.device('cuda')
    print(f'torch={torch.__version__}  iters={args.iters}')
    run_eager(args, device)
    if not args.skip_compile:
        run_compile(args, device)


if __name__ == '__main__':
    main()
```
</details>

PR #966, opened 2026-06-21
[Ops] Tile prepare_block_csr counting-sort over query blocks

yzhangcs:
## What

Speed up `prepare_block_csr` — the inversion of NSA's per-query block selection into the per-block CSR that the `bwd_dkv` kernel consumes — by tiling its counting-sort kernel over query blocks instead of launching one program per query.

## Why

The previous kernel ran one program per query (grid `(T, B*H)`), so the `[*, S]` block-index loads and the count/scatter atomics were uncoalesced and launch-overhead bound. Tiling `BT` queries per program turns them into vectorized `[BT, S]` loads and atomics. For varlen, each query's global block base is now found with a vectorized in-kernel binary search over `cu_seqlens` into `chunk_offsets`, which also lets us drop the host-side `token_indices` array.

The output contract (`csr_indices` / `csr_offsets`) is unchanged, so `tests/ops/utils/test_csr.py` passes unmodified and the NSA backward consumer is unaffected.

## Speedup

Latency, dense, `S=16` (µs, lower is better):

| T     | before | after | speedup |
|------:|-------:|------:|--------:|
| 2048  |    152 |   113 |   1.3x  |
| 8192  |    609 |   127 |   4.8x  |
| 32768 |   1895 |   362 |   5.2x  |
| 65536 |   3357 |   526 |   6.4x  |

Gains hold across block sizes (4.6–11.7× at a matched selected-token budget).

### DIFF
--- a/fla/ops/utils/csr.py
+++ b/fla/ops/utils/csr.py
@@ -9,53 +9,67 @@ import torch
 import triton
 import triton.language as tl
 
-from fla.ops.utils.index import prepare_chunk_offsets, prepare_token_indices
+from fla.ops.utils.index import prepare_chunk_offsets
 
 
 @triton.heuristics({
     'IS_VARLEN': lambda args: args['cu_seqlens'] is not None,
     'USE_BLOCK_COUNTS': lambda args: isinstance(args['block_counts'], torch.Tensor),
 })
-@triton.jit(do_not_specialize=['T'])
+@triton.jit(do_not_specialize=['T', 'N'])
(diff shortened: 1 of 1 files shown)


yzhangcs:
Verified on GPU: `tests/ops/utils/test_csr.py` 18 passed (frozen output contract unchanged) and the NSA consumer `tests/ops/test_nsa.py` 47 passed (fwd + bwd, dense + varlen).

PR #973, opened 2026-06-23
[Ops] Expand triton-ascend backends(fla/modules, fla/ops/utils) and upgrade CANN to 9.0.0

zheliuyu:
## Summary

This PR expands Ascend NPU support on the triton-ascend backend and bumps the pinned NPU software stack to CANN 9.0.0 / PyTorch 2.7.1 / triton-ascend 3.2.1.

**New or extended triton-ascend backends**
- **Modules — activations**: dispatch `gelu_fwd` / `gelu_bwd` / `sqrelu_fwd` / `sqrelu_bwd` to fixed-launch Triton kernels on NPU (avoid `@torch.compile` on Ascend).
- **Modules — token_shift**: fall back to the long kernel when `N * T > 65535` to respect Ascend Triton 2-D grid limits.
- **Ops — Based**: add `triton_ascend` backends for `parallel_based` and `fused_chunk_based`, including forward and backward, with NPU-specific grid tiling and feature-dim verifiers.
- **Ops — utils/cumsum**: add `triton_ascend` backends for `chunk_global_cumsum` and `chunk_local_cumsum` with fixed launch configs.

**Compatibility fixes for triton-ascend 3.2.1**
- Relax `FusedCrossEntropyLoss` device check to allow NPU tensors.
- Increase `causal_conv1d` backward tile `BD` from 4 to 8 for large feature dims (`D >= 1024`) on Ascend.

**Infra / packaging**
- Upgrade `ascend-a2-ci` to CANN 9.0.0, torch/torch_npu 2.7.1, torchvision 0.22.1, triton-ascend 3.2.1 (osinfra PyPI index).
- Sync `[npu]` extra in `pyproject.toml` and NPU install docs in `INSTALL.md`.
- Skip `torch.compile` recompile-contract tests in `tests/ops/utils/test_index.py` on NPU (unsupported today).

## Test plan

Verified locally on Ascend Atlas A2 (CANN 9.0.0, torch 2.7.1, torch_npu 2.7.1, triton-ascend 3.2.1):

```bash
source /usr/local/Ascend/ascend-toolkit/set_env.sh

pytest -s -v --exitfirst tests/modules tests/ops/utils tests/ops/test_based.py

### CODE REVIEW

File: fla/ops/based/backends/triton_ascend/parallel.py [hunk 1]
```diff
@@ -0,0 +1,471 @@
...
+        dv_base + t_k[:, None] * V + o_v[None, :],
+        tl.cast(b_dv, dv.dtype.element_ty, fp_downcast_rounding='rtne'),
+        mask=m_tk[:, None] & m_v[None, :],
+    )
+
+
+def _launch_parallel_kernel(kernel, g0, g1, g2, args, kwargs):
+    for nblocks, bstart in _npu_flat_grids(g0, g1, g2):
+        kernel[(nblocks,)](*args, BLOCK_START=bstart, **kwargs)
+        torch.npu.synchronize()
```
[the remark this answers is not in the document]
zheliuyu: Addressed both comments: removed torch.npu.synchronize() from the flat-grid launch loops in parallel.py and fused_chunk.py. The sync was a conservative workaround and is unnecessary because BLOCK_START gives each chunk disjoint pid ranges with no read-after-write dependency between launches.

Verified on Ascend A2 with pytest -s -v tests/ops/test_based.py — 5/5 passed.

zheliuyu:
@zhiyuan1i This PR is ready for review.

A few notes:
1. The main goal of this PR is to enable NPU CI.
2. The current test coverage has been expanded to `tests/modules`, `tests/ops/utils`, and `tests/ops/test_based.py`. However, these kernels have only passed the tests — performance hasn't been optimized yet, so the CI runs verrrrrrrry slowly right now. We'll optimize it soon.
3. The failing nvidia-ci job shows `ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device`, which seems unrelated to the code changes in this PR. See: https://github.com/fla-org/flash-linear-attention/actions/runs/28072747063/job/83110637020?pr=973#step:3:281

fla/ops/based/backends/triton_ascend/parallel.py
```diff
@@ -0,0 +1,470 @@
...
+        ).to(tl.float32)
+
+        b_ds = tl.dot(b_do, b_v, allow_tf32=False)
+        b_ds = b_ds + b_dz[:, None] * w_v0
+        b_ds = tl.where(m_s, b_ds, 0) * scale
+        b_s = tl.dot(b_q, tl.trans(b_k), allow_tf32=False)
+        b_s = tl.where(m_s, b_s, 0)
+        b_dq += tl.dot(b_ds + b_ds * b_s, b_k, allow_tf32=False)
+
+    b_dq *= scale
```
zhiyuan1i: **Bug: intra-chunk `dq` is scaled by `scale**2` instead of `scale`.**

`b_dq *= scale` here runs after *both* the inter-chunk loop (L200) and the intra-chunk loop (L220), but the intra-chunk `b_ds` already carries an inline `* scale` (L239). Net effect:

```
dq = scale * (inter) + scale**2 * (intra)
```

The CUDA reference (`fla/ops/based/parallel.py:157`) places `b_dq *= scale` *between* the two loops, so the total is `scale * (inter + intra)` — exactly one `scale` on the whole gradient. That matches the math: `dq_i = scale * sum_j Ds_ij * (1 + b_s_ij) * k_j`.

With the default `scale = K**-0.5`, the intra (causal diagonal) contribution is under-estimated by a factor of `1/scale`: ~4x at K=16, ~11x at K=128. When `T <= BTL (64)` the sequence is a single chunk, so the entire `dq` is off by that factor. `dk`/`dv` are unaffected (separate dkv kernel), so a gradcheck would show `dq` mismatching while `dk`/`dv` match — a useful way to confirm.

Suggested fix — move `b_dq *= scale` to between the two loops (same as the CUDA file and your own `fused_chunk.py:283`):

```python
    for start in range(0, chunk_start, BTS):
        ...
        b_dq += tl.dot(b_ds * (1 + b_s), b_k, allow_tf32=False)

    b_dq *= scale          # <-- move here, between the loops

    for start in range(chunk_start, (i_c + 1) * BTL, BTS):
        ...
        b_ds = tl.where(m_s, b_ds, 0) * scale
```

fla/modules/backends/triton_ascend/rotary.py
```diff
@@ -144,7 +144,7 @@ def rotary_embedding_fwdbwd_npu(
     else:
         assert seqlen_offsets + T <= TR
 
-    y = torch.empty_like(x) if not inplace else x
+    y = torch.zeros_like(x) if not inplace else x
```
zhiyuan1i, line 147: Minor: switching to `zeros_like` makes `test_rotary_left_padding_no_uninit_leak` pass on NPU, but it can mask a kernel-coverage gap with a wrong (but finite) 0 instead of the correct rotated value. Could you confirm that any position the kernel skips (e.g. per-batch `seqlen_offsets` pushing `o_cs >= TR`) is *supposed* to be 0 per the reference, rather than written by the kernel or copied from `x`? If so this is fine; otherwise it may be worth fixing the root cause (why the kernel doesn't cover those lanes on NPU) directly.

zhiyuan1i (changes requested):
Thanks for the PR — the overall design is clean and the dispatch wiring is consistent with the existing `BackendRegistry` pattern. I read the new NPU backends line-by-line against their CUDA counterparts.

**Verified correct:** `cumsum` (faithful port; autotune replaced by fixed launch configs), `fused_chunk_based` (forward + backward match CUDA, `scale` placement correct), activations `gelu`/`sqrelu` (formulas and the `@dispatch`-over-`@torch.compile` pattern are right), `token_shift` grid fallback, `fused_cross_entropy` device check, and the CANN 9.0.0 CI/packaging changes.

**Blocking issue:** the `parallel_based` NPU backward over-scales the intra-chunk `dq` by an extra `scale` — see the inline comment on `parallel.py:244`. This yields wrong gradients and should be fixed before merge.

**Test gap (couldn't comment inline since the file is untouched):** `tests/ops/test_based.py` only asserts inside `if dtype == torch.float32:` (L53, L65), but every parametrized case is `torch.float16`. So on the NPU CI the file is effectively a smoke test — it never checks forward/backward correctness, which is exactly why the `dq` bug above slips through. Please add fp16 assertions with a loose tolerance (fp16-vs-fp32 reference, e.g. rtol/atol ~1e-2; a 4x error is far beyond that) or an explicit fp32 case.

Requesting changes only because of the `parallel.py` `dq` bug; everything else LGTM.

fla/modules/backends/triton_ascend/rotary.py
zheliuyu: Thanks for flagging this. The `zeros_like` change mirrors GPU #967: under left-padding, `o_cs = o_t + seqlen_offset` can be negative, so the kernel intentionally skips those rows via `m_t` and does not store to them. The regression test `test_rotary_left_padding_no_uninit_leak` guards against uninitialized output reuse (which `empty_like` exposed).

`test_rotary_left_padding` separately verifies that all in-range tokens are rotated correctly. For the skipped padding lanes, the reference behavior is “defined, not rotated”—zero is the intended fill, not a mask for missing kernel coverage.

fla/ops/based/backends/triton_ascend/parallel.py
zheliuyu: Based NPU backends (`parallel_based` / `fused_chunk_based`) are deferred to a follow-up PR.

`tests/ops/test_based.py` only checks correctness for fp32, while CI runs fp16—so it was more of a smoke test than a real correctness check. Tightening the fp16 assertions surfaced backward issues (e.g. `parallel_based` `dq`) that still need more iteration to fix. We'd rather not hold up this PR while those kernels are still being sorted out.

This PR stays focused on enabling NPU CI for `fla/modules` and `fla/ops/utils`. Based support can land once the kernels and tests are in better shape.

PR #977, opened 2026-06-23
[Perf] Generalize fused q/k/v short convolution across layers

zhiyuan1i:
## Summary
- Generalizes the dense no-cache q/k/v short-convolution fusion from #972 into a shared `fused_qkv_short_conv` helper (plus the `can_fuse_qkv_short_conv` guard) in `fla.modules.convolution`.
- Routes every q/k/v short-conv layer through the helper so they share one dense-path fusion instead of each carrying a copy.
- The helper reads the split sizes from the conv weights, so it transparently handles standard, grouped (GQA), and multi-householder channel counts, and accepts an optional attention mask that mirrors `ShortConvolution.forward`.
- Each layer keeps its own cache / varlen guard and any layer-specific post-activation, so the non-dense fallback and existing semantics are unchanged.

## Layers
Fused path added: GatedDeltaNet, GDN2, KDA, Comba, GatedDeltaProduct, DeltaNet, GLA, MultiscaleRetention, GSA, ABC, SimpleGLA, LightNet.

Intentionally left unchanged: MoM (varlen / single-kv path) and MesaNet (q/k only, no `v_conv`).

## Test
- New `tests/modules/test_fused_qkv_short_conv.py`: compares the fused helper against three separate `ShortConvolution` calls (forward and backward) across split / bias / activation / mask variants, and checks the `can_fuse_qkv_short_conv` predicate.
- Updated `tests/layers/test_gated_deltanet.py` to patch the new guard predicate.

Follow-up to #972.

### DIFF
--- a/fla/layers/abc.py
+++ b/fla/layers/abc.py
@@ -17,6 +17,7 @@ from einops import rearrange
 from fla.layers.utils import get_layer_cache, update_layer_cache
 from fla.modules import FusedRMSNormGated, RMSNorm, RotaryEmbedding, ShortConvolution
 from fla.modules.activations import swiglu, swish
+from fla.modules.convolution import can_fuse_qkv_short_conv, fused_qkv_short_conv
 from fla.ops.abc.chunk import chunk_abc
 
 if TYPE_CHECKING:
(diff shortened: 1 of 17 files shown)


Costa-SM:
Heyo @zhiyuan1i ! 
Is there something I can help you with on this PR for us to get it to a merge-ready state ? I wanted to do some more experimentation with the GDN layer, but I don't want to step on your toes.

Issue #981, opened 2026-06-26
[Question]FLA Conv1D training stuck on Ascend NPU due to coreDim limit overflow


thinkcccc:
**Training Phenomenon**
When training Qwen3.5-27B using the Conv1D implementation from FLA across multiple ms-swift branches (main-0530, main-0623, main-0625), training consistently freezes right after step 7 with no error messages output in the logs.
Training environment: 4-machine cluster, DeepSpeed-ZeRO3, **FLA 0.5.1-conv1d**, CANN 9.0.0, triton-ascend 3.2.1.
**Root Cause Localization**
We added custom debug hooks to narrow down the crash source:
On node2/rank19/local_rank3, the backward kernel causal_conv1d_bwd_kernel triggers an illegal NPU kernel launch parameter error under the condition seq_len=4265, chunk_indices=(67,2).
The root conflict is coreDim=85440, which exceeds the hardware limit of 65535.
This failure can be stably reproduced on a single standalone NPU card.
**Mitigation Attempted**
We switched to the development branch from **PR #973** ([https://github.com/fla-org/flash-linear-attention/pull/973](https://link.wtturl.cn/?target=https%3A%2F%2Fgithub.com%2Ffla-org%2Fflash-linear-attention%2Fpull%2F973&scene=im&aid=497858&lang=zh)). This PR explicitly implements a logic: fall back to the long kernel whenever N * T > 65535 to comply with Ascend Triton's 2D grid dimension limits.
However, the original stuck-training issue still occurs after switching to this branch.

zheliuyu:
Thanks for reporting this issue.

I've confirmed that the tensor shapes produced during training in `ms-swift` with `flash-linear-attention` enabled are not yet supported by the `conv_1d` NPU kernel. We'll add support for these shapes in the `conv_1d` kernel as soon as possible — please stay tuned.

zheliuyu:
@thinkcccc Hi, I've opened a PR for this: https://github.com/fla-org/flash-linear-attention/pull/992

It chunks `triton-ascend` `causal_conv1d` launches when the grid product exceeds 65535, fixes dense backward indexing under chunked launches, and adds varlen/dense regression tests. 

Locally on Atlas 800T A3 (CANN 9.0), `pytest tests/modules/test_conv.py`  passes (75 passed).

I don't have the exact Qwen3.5-27B setup from this report. Could someone help test on NPU with a similar packed-varlen + large-D workload to confirm training no longer hangs?

[...]

PR #997, opened 2026-06-30
[Conv] Harden triton-ascend causal_conv1d silu/add against Ascend launch limits

zheliuyu:
## Summary

Follow-up to #992: conv NT/N-axis chunking is in place, but `causal_conv1d` still applied swish/silu and residual add via fixed `block=1024` 1-D launches. Under large `B*T*D` (reported in #981 as `_silu_kernel` / `coreDim > 65535`), training can still fail even after the conv kernels are fixed.

This PR is intentionally **defensive**: it hardens the remaining elementwise postprocess path so future large-shape workloads are less likely to hit Ascend launch/compile limits, rather than only patching the one shape that already failed in the wild.

## Defensive changes

- **Chunked elementwise launches**: `_launch_silu`, `_launch_silu_bwd`, and `_launch_add` now iterate with `ELEM_OFFSET` and cap each 1-D grid at 65535 (`_elementwise_launch_iters`), mirroring the chunking discipline already used for conv kernels.
- **Safer inputs**: flatten-based silu/add paths call `.contiguous()` before launch to avoid silent misreads on non-contiguous views.
- **Large-D forward tiling**: tighten fwd tile config for `D >= 8192` (`BT <= 8`) to stay within Ascend UB limits before postprocess runs.

Scope is limited to `fla/modules/backends/triton_ascend/causal_conv1d.py`; no public API changes and no torch fallback.

## Tests

- Extend `test_conv_dense_initial_state_backward_large_nt` with `B1_T8192_D4096` / `B1_T8192_D8192`.
- Extend `test_conv_varlen_initial_state_backward_random` with `packed_128x32_D4096` / `packed_128x32_D8192`.

### DIFF
--- a/fla/modules/backends/triton_ascend/causal_conv1d.py
+++ b/fla/modules/backends/triton_ascend/causal_conv1d.py
@@ -18,6 +18,13 @@ from fla.utils import input_guard
 STATIC_WARPS = 2
 # Ascend Triton rejects grids whose product exceeds 65535 (see fla/modules/token_shift.py).
 _NPU_MAX_TRITON_GRID = 65535
+_ELEM_BLOCK = 2048
+
+
+def _elementwise_launch_iters(numel: int):
+    n_blocks = triton.cdiv(numel, _ELEM_BLOCK)
+    for block_off in range(0, n_blocks, _NPU_MAX_TRITON_GRID):
+        yield min(_NPU_MAX_TRITON_GRID, n_blocks - block_off), block_off * _ELEM_BLOCK
 
 
 def _npu_chunk_size(T: int, BT: int) -> int:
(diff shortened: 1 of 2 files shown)


zheliuyu:
@zhiyuan1i  This pr is ready for review.

Follow-up to #992: `causal_conv1d` still applies `swish/silu` and residual add with a fixed block=1024, which can exceed limits under large shapes. 

This PR hardens the postprocessing path to prevent hitting Ascend limits in the future, rather than only patching the shape that already failed.

zhiyuan1i (approved):
LGTM. Verified the `_elementwise_launch_iters` + `ELEM_OFFSET` math covers every element exactly once with each 1-D grid <= 65535, and the new `D >= 8192` tile branch composes correctly with the existing NT chunking. Style/lint is clean (ruff passes with the repo config) and NPU A2 CI passes including the new D=8192 cases.

Non-blocking follow-ups for a separate PR: the `_use_seq_bwd` path (`causal_conv1d_bwd_seq_kernel`) still uses a fixed block=1024 unchunked grid, and a one-line note on the 1024 -> 2048 block bump would help future readers.

PR #998, opened 2026-07-01
[GDN] Add FlashQLA backend dispatch

Erix025:
## Summary

Add [FlashQLA](https://github.com/QwenLM/FlashQLA) as a backend for `chunk_gated_delta_rule`. FlashQLA is a TileLang-based fused forward/backward kernel for the Gated Delta Rule, developed by Qwen Team (Alibaba Cloud). It replaces the multi-kernel Triton path on supported hardware (SM90/SM100), providing both forward and backward pass support including varlen, GVA, and `state_v_first`. It also introduce intra-card CP to achieve better performance.

**Changes:**

- Register `FlashQLABackend` in the GDR backend dispatch system (`priority=3`)
- Verifier rejects unsupported configurations: non-Hopper/SM100 GPU, K != 128, V != 128, `use_gate_in_kernel`, `use_beta_sigmoid_in_kernel`, `allow_neg_eigval`, inter-card CP
- Export `IS_NVIDIA_SM100` from `fla.utils` for the compute capability check
- Add correctness tests (dense, varlen, `state_v_first`, GVA) and verifier rejection tests

**Files changed:**

| File | Change |
|------|--------|
| `fla/ops/gated_delta_rule/backends/__init__.py` | New — register `FlashQLABackend` in `gdr_registry` |
| `fla/ops/gated_delta_rule/backends/flashqla.py` | New — `FlashQLABackend` with verifier + dispatch |
| `fla/ops/gated_delta_rule/chunk.py` | Add `@dispatch('gated_delta_rule')` decorator |
| `fla/utils/_device.py` | Add `IS_NVIDIA_SM100` |
| `fla/utils/__init__.py` | Export `IS_NVIDIA_SM100` |
| `tests/ops/test_gdn.py` | Add FlashQLA correctness + verifier rejection tests |

## Test plan

- [x] `test_flashqla_chunk` — dense fwd+bwd, 3 configs (B/T/H vary, D=128), ratio ≤ 0.008
- [x] `test_flashqla_chunk_varlen` — variable-length sequences, 3 cu_seqlens configs
- [x] `test_flashqla_chunk_state_v_first` — `state_v_first=True` vs `False`, bit-exact (ratio=0)
- [x] `test_flashqla_chunk_gva` — grouped value attention (H<HV), 2 configs
- [x] `test_flashqla_verifier_rejects` — all 6 rejection branches + acceptance path
- [x] Ruff lint passes on all changed files
- [x] All 11 tests pass on NVIDIA H200, verified stable across multiple runs

```
pytest tests/ops/test_gdn.py -k "flashqla" -v
```

## Note

- FlashQLA requires the `flash_qla` package ([QwenLM/FlashQLA](https://github.com/QwenLM/FlashQLA)). Tests skip gracefully when the package is not installed.
- Disable with `FLA_FLASH_QLA=0`.


fla/ops/gated_delta_rule/backends/flashqla.py
```diff
@@ -0,0 +1,107 @@
...
+        cu_seqlens_cpu: torch.LongTensor | None = None,
+        cp_context: FLACPContext | None = None,
+        **kwargs,
+    ) -> tuple[bool, str | None]:
+        if not (IS_NVIDIA_HOPPER or IS_NVIDIA_SM100):
+            return False, "FlashQLA requires NVIDIA SM90 or SM100"
+        if q.shape[-1] != 128:
+            return False, f"FlashQLA requires K=128, got {q.shape[-1]}"
+        if v.shape[-1] != 128:
+            return False, f"FlashQLA requires V=128, got {v.shape[-1]}"
```
zhiyuan1i, line 68: The verifier doesn't check dtype. The default Triton path supports fp32, but flash_qla asserts `q.dtype != torch.float32`, so with flash_qla installed an fp32 call raises AssertionError instead of falling back to Triton. Should the verifier reject fp32 (and any other unsupported dtype)?

fla/ops/gated_delta_rule/backends/flashqla.py
```diff
@@ -0,0 +1,107 @@
...
+        use_qk_l2norm_in_kernel: bool = False,
+        state_v_first: bool = False,
+        cu_seqlens: torch.LongTensor | None = None,
+        cu_seqlens_cpu: torch.LongTensor | None = None,
+        cp_context: FLACPContext | None = None,
+        **kwargs,
+    ):
+        import flash_qla
+
+        return flash_qla.chunk_gated_delta_rule(
```
zhiyuan1i, line 102: The default Triton path returns `final_state` in fp32 (`chunk_delta_h.py` allocates it as `torch.float32`), but flash_qla returns it in the input dtype (bf16). This silently changes behavior once the backend is enabled. Is this intended, or should `final_state` be cast to fp32 to match?

zhiyuan1i (commented):
Thanks for the PR! The backend dispatch looks clean. A few things to confirm (inline):

fla/ops/gated_delta_rule/backends/flashqla.py
```diff
@@ -0,0 +1,107 @@
...
+            k=k,
+            v=v,
+            g=g,
+            beta=beta,
+            scale=scale,
+            initial_state=initial_state,
+            output_final_state=output_final_state,
+            use_qk_l2norm_in_kernel=use_qk_l2norm_in_kernel,
+            state_v_first=state_v_first,
+            cu_seqlens=cu_seqlens,
```
zhiyuan1i, line 113: The PR description mentions intra-card CP for better performance, but `auto_cp=True` is never passed (FlashQLA defaults it to False), so CP isn't actually active through FLA. Should it be enabled, or should the description be updated?

tests/ops/test_gdn.py
```diff
@@ -1284,3 +1285,319 @@ def test_prepare_wy_repr_bwd_no_g(B: int, T: int, H: int, HV: int, D: int):
...
+        for test in [
+            (1, 1024, 4, 128),
+            (2, 2048, 8, 128),
+            (1, 4096, 16, 128),
+        ]
+    ],
+)
+def test_flashqla_chunk(B, T, H, D, monkeypatch):
+    torch.manual_seed(42)
+    dtype = torch.bfloat16
```
zhiyuan1i: All FlashQLA tests use bf16; fp16 isn't covered. If fp16 is supported, could you add a test? If not, the verifier should reject it.

tests/ops/test_gdn.py
```diff
@@ -1284,3 +1285,319 @@ def test_prepare_wy_repr_bwd_no_g(B: int, T: int, H: int, HV: int, D: int):
...
+        scale=scale,
+        initial_state=h0_kv.clone(),
+        output_final_state=True,
+        use_qk_l2norm_in_kernel=True,
+        state_v_first=False,
+    )
+    ((ref_kv * do).sum() + (ref_ht_kv * dht_kv).sum()).backward(retain_graph=True)
+    ref_dq, ref_dk, ref_dv, ref_dg, ref_dbeta, ref_dh0 = q.grad, k.grad, v.grad, g.grad, beta.grad, h0_kv.grad
+
+    assert_close("o", ref_kv, tri_vk, 1e-4)
```
zhiyuan1i: The PR description says `state_v_first` is "bit-exact (ratio=0)", but this uses a `1e-4` tolerance. If it's truly bit-exact, could this use `rtol=0, atol=0`?

fla/ops/gated_delta_rule/backends/flashqla.py
Erix025: Thanks, fixed. The verifier now rejects unsupported input dtypes and only dispatches FlashQLA for fp16/bf16 q/k/v with matching dtype; fp32 now falls back to the Triton path.

fla/ops/gated_delta_rule/backends/flashqla.py
Erix025: I checked this again and the returned `final_state` is already fp32 on my side. Could you point me to the exact case or line where you observed it returning bf16? I may be missing a path, but the current output I see matches the Triton fp32 behavior.

fla/ops/gated_delta_rule/backends/flashqla.py
Erix025: Fixed. Now we pass `auto_cp=True` from the FLA wrapper, and we also updated FlashQLA so intra-card auto CP is enabled by default there as well. The PR description now matches the actual dispatch behavior.

tests/ops/test_gdn.py
Erix025: Fixed. I added fp16 coverage to the FlashQLA dtype test and kept bf16 coverage as well, so both supported input dtypes now run through fwd/bwd correctness checks.

tests/ops/test_gdn.py
Erix025: Fixed. I tightened the `state_v_first` assertions to exact equality (`rtol=0, atol=0`), and the test still passes, so this now matches the bit-exact claim in the PR description.

Erix025:
> Thanks for the PR! The backend dispatch looks clean. A few things to confirm (inline):

Thanks for the careful review. We addressed the verifier dtype checks, fp16 test coverage, `state_v_first` exactness check, and the `auto_cp` path. For `final_state`, I rechecked and it appears to already be fp32; I left a question on that thread to confirm which case showed bf16.

yzhangcs (approved):

PR #1019, opened 2026-07-13
[Misc] Add faceted benchmark visualization and multi-device benchmark support

sunyi0505:
## Summary

Adds **faceted visualization** to the FLA benchmark workflow and aligns module/ops benchmark output and device handling, making results easier to read and compare across commits and hardware backends (including Ascend NPU).

### 1. New faceted visualization module (`benchmarks/visualize.py`)

- Supports two input types:
  - **Triton `perf_report` CSV** (module benchmark output)
  - **`benchmark_results.json`** (cross-commit comparison from `scripts/run_benchmark_compare.py`)
- For multi-dimensional shape sweeps (B/T/H/D/L), automatically picks the x-axis and splits remaining dimensions into **facet subplots**
- Separates Forward / Backward providers and generates `*_faceted.png` plus an HTML gallery (`results.html`)
- Comparison mode produces a **percent-change bar chart** and a **base vs head latency chart**
- CLI entry point: `python -m benchmarks.visualize perf-report|comparison`

### 2. New shared module benchmark runner (`benchmarks/modules/runner.py`)

- Introduces `run_module_benchmark()`: runs Triton `perf_report` and optionally auto-generates faceted plots
- Auto-infers `save_dir` (e.g. `benchmark_layernorm.py` → `./layernorm_benchmark`)
- Auto-detects `plot_name` and calls `visualize_perf_report_csv` after the benchmark finishes

### 3. Updated 6 module benchmark scripts

These scripts now use `run_module_benchmark()` instead of calling `benchmark.run()` directly, and auto-generate visualizations after each run:

- `benchmark_activations.py`
- `benchmark_conv.py`
- `benchmark_cross_entropy.py`
- `benchmark_l2norm.py`
- `benchmark_layernorm.py`
- `benchmark_tokenshift.py`

### 4. Multi-device support for ops benchmarks (`benchmarks/ops/run.py`)

- Replaces hardcoded `'cuda'` with `fla.utils.device_name`
- Adds `_device_synchronize()` for consistent device sync during warmup and timing
- Extends `_get_machine_info()` for NPU: records `device_platform`, NPU device name, and memory
- Adds `_format_machine_line()` for consistent machine info output (CUDA / NPU / PyTorch version)

### 5. Ops verify updates (`benchmarks/ops/verify.py`)

- Profiler selects `ProfilerActivity` based on device type (CUDA uses `self_cuda_time_total`, otherwise CPU)
- Input generation and synchronization aligned with `run.py` for non-CUDA devices

### 6. Cross-commit comparison script enhancement (`scripts/run_benchmark_compare.py`)

- Adds `--plot` flag to auto-generate visualization plots after comparison completes

### 7. Dependency update (`pyproject.toml`)

- Adds `pandas` to the `benchmark` optional extra (required for CSV parsing in visualization)

---

## Usage Examples

```bash
# Module benchmark (auto-generates plots)
python benchmarks/modules/benchmark_activations.py
```
image

```shell
# Manually visualize an existing CSV
python -m benchmarks.visualize perf-report benchmarks/modules/activation_benchmark/activation_performance.csv

# Cross-commit comparison + visualization
python scripts/run_benchmark_compare.py --base HEAD --output results.json --plot plots/
```

### DIFF
--- a/benchmarks/ops/run.py
+++ b/benchmarks/ops/run.py
@@ -141,6 +141,8 @@ import tempfile
 
 import torch
 
+from fla.utils import device_name
+
 # Import registry — works both as a package (python -m benchmarks.ops.run)
 # and standalone (python /tmp/fla_bench_xxx/run.py) for cross-commit use.
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
@@ -153,6 +155,27 @@ from registry import (  # noqa: E402
 )
 
 logger = logging.getLogger(__name__)
+
+
+def _device_synchronize(device: str | None = None) -> None:
+    """Synchronize the active accelerator before/after timed runs."""
+    dev = device or device_name
+    dev_mod = getattr(torch, dev, None)
+    if dev_mod is not None and hasattr(dev_mod, 'synchronize'):
+        dev_mod.synchronize()
+
+
+def _format_machine_line(info: dict) -> str:
+    gpu = info.get('gpu_name', 'N/A')
+    pytorch = info.get('pytorch_version', 'N/A')
+    platform = info.get('device_platform', 'N/A')
+    if platform == 'cuda':
+        backend = f"CUDA {info.get('cuda_version', 'N/A')}"
+    elif platform == 'npu':
+        backend = f"NPU: CANN {info.get('cann_version', 'N/A')}"
+    else:
+        backend = platform.upper() if platform != 'N/A' else 'N/A'
+    return f"Machine: {gpu} | {backend} | PyTorch {pytorch}"
 
 
 def _import_op(config: OpConfig):
@@ -185,11 +208,14 @@ def _get_git_label() -> str:
 
 
 def _get_machine_info() -> dict:
+    from fla.utils import device_platform
+
     info = {
         'hostname': socket.gethostname(),
         'platform': platform.platform(),
         'pytorch_version': torch.__version__,
         'cuda_version': torch.version.cuda or 'N/A',
+        'device_platform': device_platform,
         'git_label': _get_git_label(),
     }
     try:
@@ -198,7 +224,13 @@ def _get_machine_info() -> dict:
     except Exception:
         info['triton_version'] = 'N/A'
 
-    if torch.cuda.is_available():
+    if device_platform == 'npu' and hasattr(torch, 'npu') and torch.npu.is_available():
+        props = torch.npu.get_device_properties(0)
+        info['gpu_name'] = torch.npu.get_device_name(0)
+        info['gpu_count'] = torch.npu.device_count()
+        info['gpu_memory_gb'] = round(props.total_memory / (1024**3), 1)
+        info['cann_version'] = getattr(torch.version, 'cann', None) or 'N/A'
+    elif torch.cuda.is_available():
         info['gpu_name'] = torch.cuda.get_device_name(0)
         info['gpu_count'] = torch.cuda.device_count()
         info['gpu_memory_gb'] = round(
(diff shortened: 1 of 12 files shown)


sunyi0505:

zhiyuan1i (changes requested):
Thanks for the PR — the faceted visualization and the device generalization look good overall, and the comparison-JSON contract lines up with what `run_benchmark_compare.py` writes. One correctness bug to fix before merging:

**`visualize.py`: IndexError in two-row (fwd+bwd) mode when there are more than 5 facet groups**

When a CSV has both forward and backward providers (`n_mode_rows > 1`), `_grid_shape` caps the grid at `ncols = min(n_facets, 5)` with `nrows = n_mode_rows` (= 2). The main loop then indexes subplots as `axes[mode_row, facet_idx]`, so with 6+ facet combinations (e.g. a B×D sweep with 2×3 shape values), `facet_idx = 5` goes out of bounds and plotting crashes with `IndexError: index 5 is out of bounds for axis 1 with size 5`.

The single-row branch handles this correctly by wrapping to the next row (`row = facet_idx // ncols; col = facet_idx % ncols`); the two-row branch is missing the equivalent wrapping. The "hide unused facet columns" loop (`range(n_facets, ncols)`) also only covers the `n_facets <= ncols` case, which suggests the overflow case wasn't considered.

Suggested fixes (either one):

- Wrap in two-row mode as well: grid of `(n_mode_rows * ceil(n_facets / ncols), ncols)` with `axes[mode_row * rows_per_mode + facet_idx // ncols, facet_idx % ncols]`, or
- Simply don't cap the columns in two-row mode (`ncols = n_facets`) — wide, but correct.

**Minor, non-blocking:**

1. The 6 module scripts now do `from runner import run_module_benchmark`, which only resolves when run as `python benchmarks/modules/benchmark_xxx.py`; `python -m benchmarks.modules.benchmark_xxx` raises `ModuleNotFoundError: runner`. A try/except falling back to `from benchmarks.modules.runner import ...` would keep both invocation styles working.
2. With `visualize=True` (the default), a missing matplotlib/pandas install makes the script exit with a traceback *after* the benchmark itself has completed and printed. Consider catching `ImportError` around the visualization call in `run_module_benchmark` and degrading to a warning.
3. `--plot` in `run_benchmark_compare.py` silently does nothing unless `--output` is also given; a warning would improve discoverability.

sunyi0505:
Thanks for your suggestions. I have fixed all the issues. @zhiyuan1i

benchmarks/ops/run.py
```diff
@@ -155,6 +157,27 @@
...
+
+
+def _format_machine_line(info: dict) -> str:
+    gpu = info.get('gpu_name', 'N/A')
+    pytorch = info.get('pytorch_version', 'N/A')
+    platform = info.get('device_platform', 'N/A')
+    if platform == 'cuda':
+        backend = f"CUDA {info.get('cuda_version', 'N/A')}"
+    elif platform == 'npu':
+        backend = 'NPU'
```
zheliuyu: Suggest staying consistent with "CUDA" and printing the `CANN + {version}` field.

sunyi0505: added

zhiyuan1i (approved):
All four points from the previous review are addressed in ee3df18:

- The two-row grid now wraps correctly via `rows_per_mode` — I brute-forced the indexing for `n_mode_rows ∈ {1,2}` × `n_facets ∈ [1,12]` and every subplot lands in bounds with no overlaps, including the hide-unused-slots loop.
- The try/except import fallback keeps both `python benchmarks/modules/benchmark_xxx.py` and `python -m benchmarks.modules.benchmark_xxx` working.
- Visualization now degrades to a warning when matplotlib/pandas are missing.
- `--plot` without `--output` prints a warning.

Also confirmed the CANN version line from the other review thread. LGTM, approving.

sunyi0505:
Hi @zhiyuan1i , this PR has been approved and all CI checks are green. Could you help merge when you have a moment? Thanks!

PR #1020, opened 2026-07-13
[Ops] Add triton-ascend recompute_w_u_fwd backend dispatch for GDN

OsirisDuan:


zheliuyu:
Once PR #1019 is merged, it would be helpful to include the benchmark output from `python benchmarks/ops/run.py --op chunk_gdn` to facilitate a direct performance comparison between the pre‑ and post‑optimization versions.

PR #1023, opened 2026-07-14
[Ops] Rewrite NPU chunk_scaled_dot_kkt_fwd

OsirisDuan:
## Summary

Rewrite the Ascend NPU (`triton_ascend` backend) implementation of `chunk_scaled_dot_kkt_fwd` to use a 1-D AICore task-loop grid in place of the upstream 2-D `(NT, B*HV)` grid + host-side `NT_OFFSET`/`BH_OFFSET` sub-launches, and to replace the previous `tl.make_block_ptr` + `boundary_check` loads/stores with explicit `ptr + offset + mask` indexing that mirrors the GPU mainline kernel. As part of the rewrite:

- `T` is cast to `tl.int64` once at the top of the kernel, so `i_b * T` / `bos * H * K` style pointer arithmetic can no longer overflow `int32` for long sequences (this subsumes the varlen `bos/eos` overflow fix from #1118 — we use int64 for both branches and for `T` itself, since `make_block_ptr` is no longer used and the int32-shape constraint no longer applies).
- `BK` is chosen by `_get_fwd_bk` (UB-safe row-tile sizing via `compute_row_tile_block_size`) instead of `triton.autotune`, with a fallback of 16 and a cap of 128.
- The `chunk_scaled_dot_kkt_fwd_verifier` on the `TritonAscendCommonBackend` now actually checks `IS_NPU`, `k.device.type == "npu"`, and the supported dtype set (`fp32`/`fp16`/`bf16`) instead of returning `True, None` unconditionally.

The GPU path in `fla/ops/common/chunk_scaled_dot_kkt.py` is untouched; only the `triton_ascend` backend and its `__init__.py` verifier are modified.

## Test plan

The changed kernel is invoked through `fla.ops.common.chunk_scaled_dot_kkt_fwd` (via `@dispatch('common')`) by the following ops on NPU:

- `fla.ops.gated_delta_rule` (via `fla/ops/gated_delta_rule/backends/triton_ascend/chunk_fwd.py`)
- `fla.ops.delta_rule` (via `fla/ops/delta_rule/wy_fast.py`)
- `fla.ops.gated_delta_product` (via `fla/ops/gated_delta_product/chunk.py`)
- `fla.ops.path_attn` (via `fla/ops/path_attn/parallel.py`)

Existing main-repo test files that exercise this code path on NPU (to be run on Ascend hardware before merge):

- `tests/ops/test_gdn_kernels.py` — direct `chunk_scaled_dot_kkt_fwd` coverage via `test_chunk_kkt_solve`; indirect coverage through `test_gdn_gate_fwd`, `test_gdn_gate_chunk_cumsum`, `test_chunk_fwd_o`, `test_chunk_gated_delta_rule_fwd_h[_varlen_many_documents]`, `test_prepare_wy_repr_bwd[_varlen]`, `test_gdn_full_bwd`, `test_chunk_gated_delta_rule_bwd_dhu[_varlen_many_documents]` (dense + varlen).
- `tests/ops/test_gdn.py` — `test_chunk[_with_chunk_size|_varlen|_varlen_prefill|_gate_in_kernel[_gqa|_varlen]|_state_v_first|_beta_sigmoid_in_kernel]` exercise the full GDN layer that calls the rewritten kernel.
- `tests/ops/test_delta.py` — `test_chunk`, `test_chunk_with_chunk_size`, `test_chunk_varlen`.
- `tests/ops/test_gdp.py` — `test_chunk`, `test_chunk_varlen`.
- `tests/ops/test_solve_tril.py` — directly calls `chunk_scaled_dot_kkt_fwd` with `cu_seqlens` (varlen path).

Run with:

```
pytest tests/ops/test_gdn_kernels.py tests/ops/test_gdn.py tests/ops/test_delta.py tests/ops/test_gdp.py tests/ops/test_solve_tril.py
```

Hardware: Ascend NPU. Both dense and `cu_seqlens` (varlen, many-doc) cases are covered by the above.

python benchmarks/ops/run.py --op chunk_gdn

```
===========================================================================================================================
  Machine: Ascend910B3 | NPU: CANN 9.0.0-beta.2 | PyTorch 2.7.1+cpu
===========================================================================================================================
  fwd        B      T    H    D  op                                   main[0f0f0c97](ms) 260714/c...[fb04a82f](ms)  speedup
          -----------------------------------------------------------------------------------------------------------------
             1   8192   96  128  chunk_gdn                                       323.358                    93.655    3.45x
          -----------------------------------------------------------------------------------------------------------------
             2  16384   16  128  chunk_gdn                                       221.405                    73.195    3.02x
          -----------------------------------------------------------------------------------------------------------------
             4   2048   16  128  chunk_gdn                                        53.584                    16.509    3.25x
          -----------------------------------------------------------------------------------------------------------------
             4   4096   64  128  chunk_gdn                                       425.553                   122.923    3.46x
          -----------------------------------------------------------------------------------------------------------------
             8   1024    8   64  chunk_gdn                                        24.072                     5.704    4.22x
          -----------------------------------------------------------------------------------------------------------------
             8   2048   32  256  chunk_gdn                                       269.464                   115.775    2.33x
===========================================================================================================================
  fwdbwd     B      T    H    D  op                                   main[0f0f0c97](ms) 260714/c...[fb04a82f](ms)  speedup
          -----------------------------------------------------------------------------------------------------------------
             1   8192   96  128  chunk_gdn                                       528.580                   298.662    1.77x
          -----------------------------------------------------------------------------------------------------------------
             2  16384   16  128  chunk_gdn                                       369.012                   220.120    1.68x
          -----------------------------------------------------------------------------------------------------------------
             4   2048   16  128  chunk_gdn                                        88.748                    51.918    1.71x
          -----------------------------------------------------------------------------------------------------------------
             4   4096   64  128  chunk_gdn                                       696.982                   392.892    1.77x
          -----------------------------------------------------------------------------------------------------------------
             8   1024    8   64  chunk_gdn                                        33.390                    15.125    2.21x
          -----------------------------------------------------------------------------------------------------------------
             8   2048   32  256  chunk_gdn                                       585.052                   430.715    1.36x
===========================================================================================================================
```

## Breaking changes

None. The change is internal to the `triton_ascend` backend; the public `chunk_scaled_dot_kkt_fwd(k, g, beta, cu_seqlens, chunk_size, output_dtype, chunk_indices)` signature is unchanged. The verifier is now stricter (rejects non-NPU devices and unsupported dtypes instead of always returning `True, None`), which aligns actual dispatch behavior with what the kernel implementation supports.

## Checklist

### DIFF
--- a/fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
+++ b/fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
@@ -73,9 +73,9 @@ def chunk_scaled_dot_kkt_fwd_kernel_npu(
     IS_VARLEN: tl.constexpr,
     USE_G: tl.constexpr,
 ):
-    bt_stride = B.to(tl.int64) * T.to(tl.int64)
+    T = T.to(tl.int64)
+    bt_stride = B.to(tl.int64) * T
     core_id = tl.program_id(0)
-    T64 = T.to(tl.int64)
 
     for task_id in tl.range(core_id, task_num, num_core):
         i_t_i = task_id // bh_step
@@ -84,39 +84,39 @@ def chunk_scaled_dot_kkt_fwd_kernel_npu(
         if IS_VARLEN:
             i_n, i_t = (
                 tl.load(chunk_indices + i_t_i * 2).to(tl.int32),
-                tl.load(chunk_indices + i_t_i * 2 + 1).to(tl.int32),
+                tl.load(chunk_indices + i_t_i * 2 + 1).to(tl.int64),
             )
-            bos = tl.load(cu_seqlens + i_n)
-            eos = tl.load(cu_seqlens + i_n + 1)
-            T = eos.to(tl.int32) - bos.to(tl.int32)
+            bos = tl.load(cu_seqlens + i_n).to(tl.int64)
+            eos = tl.load(cu_seqlens + i_n + 1).to(tl.int64)
+            T = eos - bos
         else:
-            bos = i_b.to(tl.int64) * T64
-            eos = bos + T64
-            i_t = i_t_i
-        o_t = tl.arange(0, BT)
-        o_t_fp32 = o_t.to(tl.float32)
+            bos = i_b.to(tl.int64) * T
+            eos = bos + T
+            i_t = i_t_i.to(tl.int64)
+        o_t = i_t * BT + tl.arange(0, BT)
+        m_t = o_t < T
 
-        p_beta = tl.make_block_ptr(beta + i_h * bt_stride + bos, (T,), (1,), (i_t * BT,), (BT,), (0,))
-        b_beta = tl.load(p_beta, boundary_check=(0,))
+        p_beta = beta + i_h * bt_stride + bos + o_t
+        b_beta = tl.load(p_beta, mask=m_t, other=0.0)
 
         b_A = tl.zeros([BT, BT], dtype=tl.float32)
         for i_k in range(tl.cdiv(K, BK)):
-            p_k = tl.make_block_ptr(
-                k + (bos * H + i_h // (HV // H)) * K, (T, K), (H * K, 1), (i_t * BT, i_k * BK), (BT, BK), (1, 0)
-            )
-            b_k = tl.load(p_k, boundary_check=(0, 1))
+            o_k = i_k * BK + tl.arange(0, BK)
+            p_k = k + (bos * H + i_h // (HV // H)) * K + o_t[:, None] * (H * K) + o_k[None, :]
+            b_k = tl.load(p_k, mask=m_t[:, None] & (o_k < K)[None, :], other=0.0)
             b_A += tl.dot(b_k, tl.trans(b_k))
 
         if USE_G:
-            p_g = tl.make_block_ptr(g + i_h * bt_stride + bos, (T,), (1,), (i_t * BT,), (BT,), (0,))
-            b_g = tl.load(p_g, boundary_check=(0,))
+            p_g = g + i_h * bt_stride + bos + o_t
+            b_g = tl.load(p_g, mask=m_t, other=0.0)
             b_g_diff = b_g[:, None] - b_g[None, :]
             b_A *= exp2(b_g_diff)
 
         b_A *= b_beta[:, None]
-        b_A = tl.where(o_t_fp32[:, None] > o_t_fp32[None, :], b_A, 0)
-        p_A = tl.make_block_ptr(A + (bos * HV + i_h) * BT, (T, BT), (BT * HV, 1), (i_t * BT, 0), (BT, BT), (1, 0))
-        tl.store(p_A, b_A.to(p_A.dtype.element_ty), boundary_check=(0, 1))
+        m_A = (o_t[:, None] > o_t[None, :]) & (m_t[:, None] & m_t)
+        b_A = tl.where(m_A, b_A, 0)
+        p_A = A + (bos * HV + i_h) * BT + o_t[:, None] * (BT * HV) + tl.arange(0, BT)[None, :]
+        tl.store(p_A, b_A.to(p_A.dtype.element_ty), mask=m_t[:, None])
 
 
 @input_guard
--- a/fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
+++ b/fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
@@ -16,7 +16,29 @@ import triton.runtime.driver as driver
 
 from fla.ops.utils import prepare_chunk_indices
 from fla.ops.utils.op import exp2
-from fla.utils import autotune_cache_kwargs, input_guard
+from fla.utils import input_guard
+from fla.utils.ascend_ub_manager import compute_row_tile_block_size
+
+# Peak live fp32 tiles: b_A[BT,BT], b_k[BT,BK], plus tl.dot intermediate buffer
+_CHUNK_SCALED_DOT_KKT_MEM_MULT = 5.0
+_SAFETY_MARGIN = 0.85
+_FALLBACK_BK = 16
+_MAX_BK_FWD = 128
+
+
+def _get_fwd_bk(BT: int, K: int) -> int:
+    """UB-safe BK tile size for chunk_scaled_dot_kkt_fwd on NPU."""
+    return compute_row_tile_block_size(
+        BT,
+        K,
+        _CHUNK_SCALED_DOT_KKT_MEM_MULT,
+        tiling_row=False,
+        safety_margin=_SAFETY_MARGIN,
+        dtype_size=4,
+        fallback=_FALLBACK_BK,
+        min_block=16,
+        max_block=min(_MAX_BK_FWD, triton.next_power_of_2(K)),
+    )
 
 
 def get_npu_properties():
@@ -29,16 +51,6 @@ def get_npu_properties():
         "IS_VARLEN": lambda args: args["cu_seqlens"] is not None,
         "USE_G": lambda args: args["g"] is not None,
     }
-)
-@triton.autotune(
-    configs=[
-        triton.Config({'BK': BK}, num_warps=num_warps, num_stages=num_stages)
-        for BK in [32, 64, 128]
-        for num_warps in [2, 4, 8]
-        for num_stages in [2, 3, 4]
-    ],
-    key=['H', 'HV', 'K', 'BT', 'IS_VARLEN'],
-    **autotune_cache_kwargs,
 )
 @triton.jit(do_not_specialize=["T", "B", "bh_step", "task_num", "num_core"])
 def chunk_scaled_dot_kkt_fwd_kernel_npu(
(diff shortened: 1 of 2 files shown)


OsirisDuan:
@zheliuyu Please review this PR.

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
```diff
@@ -12,57 +12,35 @@
...
+@triton.autotune(
+    configs=[
+        triton.Config({'BK': BK}, num_warps=num_warps, num_stages=num_stages)
+        for BK in [32, 64, 128]
+        for num_warps in [2, 4, 8]
+        for num_stages in [2, 3, 4]
+    ],
+    key=['H', 'HV', 'K', 'BT', 'IS_VARLEN'],
+    **autotune_cache_kwargs,
+)
```
zheliuyu: blocking: full [BT,BT] + autotune BK≤128 drops the UB-aware sizing used elsewhere (#1001). Fine if faster, but please clamp/filter configs or confirm the UB envelope on 910B/C so we don’t hit intermittent UB/compile failures outside the bench shapes.

zheliuyu: nit: on triton-ascend, num_warps/num_stages are ignored (see Ascend docs /
"[WARNING] Please DO NOT tune arg ..."). Prefer autotuning BK only (and maybe
multibuffer) to avoid a 3×3 wasted search.

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
```diff
@@ -71,73 +49,59 @@ def chunk_scaled_dot_kkt_fwd_kernel_npu(
...
-            p_A = tl.make_block_ptr(A_base, (T, BT), (HV * BT, 1), (i_tc_s, c * BC), (BC, BC), (1, 0))
-            tl.store(p_A, b_A.to(p_A.dtype.element_ty), boundary_check=(0, 1))
+            p_g = tl.make_block_ptr(g + i_h * bt_stride + bos, (T,), (1,), (i_t * BT,), (BT,), (0,))
+            b_g = tl.load(p_g, boundary_check=(0,))
+            b_g_diff = b_g[:, None] - b_g[None, :]
+            b_A *= exp2(b_g_diff)
+
+        b_A *= b_beta[:, None]
+        b_A = tl.where(o_t_fp32[:, None] > o_t_fp32[None, :], b_A, 0)
+        p_A = tl.make_block_ptr(A + (bos * HV + i_h) * BT, (T, BT), (BT * HV, 1), (i_t * BT, 0), (BT, BT), (1, 0))
```
zhiyuan1i: Possible int32 overflow in the pointer offsets. `task_id`/`bos`/`bt_stride` are all int32 here, so `(bos * HV + i_h) * BT` — and likewise `(bos * H + i_h // (HV // H)) * K` above and `i_h * bt_stride` — can overflow on large inputs. A varlen batch with total T = 1M tokens, HV = 32, BT = 64 already reaches `bos*HV*BT ≈ 2e9` (near INT32_MAX); HV = 96 overflows outright. The reference kernel guards against this with `tl.program_id(1).to(tl.int64)`. Please promote the offset computation to int64 (e.g. cast `bos` or `task_id` to `tl.int64` before the multiplies).

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
```diff
@@ -150,37 +114,57 @@ def chunk_scaled_dot_kkt_fwd_npu(
...
-            'IS_VARLEN': cu_seqlens is not None,
-            'NT_OFFSET': 0,
-            'BH_OFFSET': 0,
-        },
+
+    num_core = get_npu_properties()["num_aicore"]
+    bh_step = B * HV
+    task_num = NT * bh_step
+    g_arg = torch.permute(g, (2, 0, 1)).contiguous() if g is not None else g
+    beta_arg = torch.permute(beta, (2, 0, 1)).contiguous() if beta is not None else beta
```
zhiyuan1i: nit: the `if beta is not None else beta` fallback is dead code — line 141 already dereferences `beta.shape[2]`, and the kernel loads `beta` unconditionally, so `beta` can never be `None` at this point. Suggest writing this as an unconditional permute (the `g` guard above is real and should stay).

zhiyuan1i (changes requested):
Thanks for the rewrite — the semantics now match the reference kernel and the 1.27-1.28x speedup is nice. Two things before this lands:

1. (requesting change) The pointer offset arithmetic stays in int32 and can overflow on large varlen inputs — see inline comment.
2. (nit) A dead `beta is not None` branch in the host wrapper.

Also +1 to @zheliuyu's blocking comment on validating the UB envelope for the fixed BK autotune space.

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
OsirisDuan: Fixed.

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
OsirisDuan: Fixed.

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
OsirisDuan: fixed

OsirisDuan:
@zhiyuan1i Please review this PR.

zheliuyu:
@OsirisDuan  It is recommended to rebase onto the current `main` branch to better evaluate the performance gains.   :)

OsirisDuan:
> @OsirisDuan It is recommended to rebase onto the current `main` branch to better evaluate the performance gains. :)

Done.

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
```diff
@@ -71,73 +61,62 @@ def chunk_scaled_dot_kkt_fwd_kernel_npu(
...
+        i_bh = task_id % bh_step
+        i_b, i_h = i_bh // HV, i_bh % HV
+        if IS_VARLEN:
+            i_n, i_t = (
+                tl.load(chunk_indices + i_t_i * 2).to(tl.int32),
+                tl.load(chunk_indices + i_t_i * 2 + 1).to(tl.int32),
+            )
+            bos = tl.load(cu_seqlens + i_n)
+            eos = tl.load(cu_seqlens + i_n + 1)
+            T = eos.to(tl.int32) - bos.to(tl.int32)
```
zhiyuan1i: **(Blocking, regression of the previous review point)** `cu_seqlens` is int32 in practice, so `bos` stays int32 here, and the pointer math below — `(bos * H + i_h // (HV // H)) * K` (L105), `(bos * HV + i_h) * BT` (L118), and `beta + i_h * bt_stride + bos` (L99) — overflows for large packed varlen batches (total_T=1M, HV=96, BT=64 → 6.1e9). Commit 6d8408a had this fixed with `.to(tl.int64)`; the final commit dropped it — I assume the int64 `T` was what actually broke triton-ascend (block-ptr shapes?). Could you keep `T` as int32 but promote only the pointer base, e.g. keep this line as-is and add `bos = bos.to(tl.int64)` right after? If int64 offsets themselves hit a triton-ascend limitation, please document it and let's discuss a host-side envelope check instead.

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
```diff
@@ -71,73 +61,62 @@ def chunk_scaled_dot_kkt_fwd_kernel_npu(
...
+        else:
+            bos = i_b.to(tl.int64) * T64
+            eos = bos + T64
```
zhiyuan1i: Nit: dead variable — `eos` is never used in the non-varlen branch.

zhiyuan1i (changes requested):
Thanks for the rebase — the merge conflict with #1070 is gone, A2 CI is green, and dropping the autotune in favor of `_get_fwd_bk()` (UB-aware sizing) addresses @zheliuyu's blocking concern nicely. The 1D core-grid rewrite also matches our Ascend conventions.

One remaining blocker from my previous review (inline): the int32 overflow is back in the varlen path — the last commit ("dtype bug fix") dropped the `.to(tl.int64)` that an earlier commit had on `bos`. As of HEAD the pointer math at L105/L118 is int32 again and overflows for large packed varlen batches.

Non-blocking nits: the kernel now writes the full [BT,BT] tile (upper triangle zeroed in-kernel), so the `torch.zeros` init inherited from #1070 can go back to `torch.empty` (matches the GPU reference, saves a full-buffer memset) — your call, worth a comment either way. Also: dead `eos = bos + T64` in the non-varlen branch (inline); was dropping `allow_tf32=False` intentional (only affects fp32 inputs)? And one benchmark row in the PR body still shows "[...truncated]".

The H100 (PyTorch 2.12) job is just queued on the self-hosted runner — not a failure.

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
OsirisDuan: Fixed.

fla/ops/common/backends/triton_ascend/chunk_scaled_dot_kkt.py
OsirisDuan: Fixed.

OsirisDuan:
@zhiyuan1i The current PR #1023 has been updated. Could you please review this PR? Thanks!

zhiyuan1i (approved):

PR #1024, opened 2026-07-15
[Perf] Optimize the performance of causal conv1D on Ascend NPU

ChengQianqian:
Optimize the performance of causal conv1D
优化点：

Perf：
image

### DIFF
--- a/fla/modules/backends/triton_ascend/causal_conv1d.py
+++ b/fla/modules/backends/triton_ascend/causal_conv1d.py
@@ -129,6 +129,7 @@ def _npu_bwd_tile_config(
 @triton.heuristics({
     'HAS_WEIGHT': lambda args: args['weight'] is not None,
     'HAS_BIAS': lambda args: args['bias'] is not None,
+    'HAS_RESIDUAL': lambda args: args['residual'] is not None,
     'USE_INITIAL_STATE': lambda args: args['initial_state'] is not None,
     'IS_VARLEN': lambda args: args['cu_seqlens'] is not None,
 })
@@ -138,6 +139,7 @@ def causal_conv1d_fwd_kernel(
     y,
     weight,
     bias,
+    residual,
     cu_seqlens,
     initial_state,
     chunk_indices,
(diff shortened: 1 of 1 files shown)


fla/modules/backends/triton_ascend/causal_conv1d.py
```diff
@@ -771,52 +1021,47 @@ def _launch_fwd_core(
...
 
     stride_x_n, stride_x_t, stride_x_d = x.stride()
-    y = torch.zeros_like(x, memory_format=torch.contiguous_format)
-    stride_y_n, stride_y_t, stride_y_d = y.stride()
+    y = torch.empty_like(x, memory_format=torch.contiguous_format)
 
-    max_nt = _npu_max_axis_chunks(triton.cdiv(D, BD), B)
-    kernel_kwargs = dict(
+    grid = (triton.cdiv(D, BD), NT_GRID, B)
+    causal_conv1d_fwd_kernel[grid](
```
zheliuyu: Grid size regression: `_launch_fwd_core` now launches the forward kernel in one shot with `grid = (cdiv(D, BD), NT_GRID, B)` (BD=128, NT_GRID=cdiv(NT, 4)), but the old code chunked the NT axis via `_npu_max_axis_chunks` to stay under Ascend's 65535 grid limit.

This can overflow at boundary shapes — e.g. `D=8192, T=8192, B=32, BT=64` gives `64 × 32 × 32 = 65536`. Please either restore the launch loop (or equivalent chunking), or show that grid stays ≤65535 for all supported shapes and add a test for this boundary case.

fla/modules/backends/triton_ascend/causal_conv1d.py
```diff
@@ -425,21 +468,179 @@ def causal_conv1d_bwd_seq_kernel(
...
+        stride_dy_n=stride_dy_n,
+        stride_dy_t=stride_dy_t,
+        stride_dy_d=stride_dy_d,
+        stride_dx_n=stride_dx_n,
+        stride_dx_t=stride_dx_t,
+        stride_dx_d=stride_dx_d,
+        num_warps=num_warps,
+    )
+    if dht is not None:
+        kernel_kwargs['num_stages'] = 1
```
sunyi0505: The parameters num_stages and num_warps have no effect under the triton-ascend backend. Please remove these two parameters from your kernel code.

zhiyuan1i (approved):

PR #1026, opened 2026-07-15
[Fix] fall back to Triton when TileLang has no usable nvcc compiler

### LINKED ISSUE #949: [Bug] nvcc related crash on first forward pass - resolved with FLA_DISABLE_BACKEND_DISPATCH=1
### Checklist

### Describe the Bug

When I run my training script after installing flash-linear-attention[cuda] , I get a complicated crash involving nvcc and incompatible headers. There is no nvcc on my system. I captured the full output of the script, including the crash, in the attached file.

[error.txt](https://github.com/user-attachments/files/29069869/error.txt)

Thanks to Claude Code, I found a workaround. I set FLA_DISABLE_BACKEND_DISPATCH=1 and the system works. 

If this cannot be easily fixed, could this workaround be added to the readme or FAQ?

Use standard Trainer to train a Qwen 3.5 model after installing flash-linear-attention[cuda] and causal-conv1d . The crash apparently happens at the first forward pass.

### Expected Behavior

training goes on

### Environment Information

1. Torch: 2.11.0
2. Triton: 3.6.0
3. CUDA 13.0 ; container image based on the NVidia CUDA 13.0 *runtime* image with Ubuntu 24.04

mvanhorn:
## Summary
`TileLangBackend.is_available()` returned true whenever the `tilelang` package was importable, even on machines with no CUDA compiler. TileLang then failed at kernel-build time instead of falling back to Triton. This gates availability on a usable `nvcc`: `shutil.which("nvcc")`, then `CUDA_HOME/bin/nvcc`, then the `nvidia-cuda-nvcc` wheel. When none is found it logs a one-line hint and reports unavailable, so the existing Triton path is used.

## Why this matters
Installing `tilelang` without a matching CUDA toolkit is common (it pulls in as a transitive dep), and today that silently routes work to a backend that cannot compile. The check is cheap and `@cache`d, so it runs once per process.

## Testing
Added `tests/ops/test_backends.py` covering: nvcc on PATH, nvcc under `CUDA_HOME`, the `nvidia-cuda-nvcc` wheel present, and none available (falls back, logs the hint). `FLA_TILELANG=0` still disables TileLang explicitly.

Closes #949

AI was used for assistance.

### DIFF
--- a/fla/ops/common/backends/tilelang/__init__.py
+++ b/fla/ops/common/backends/tilelang/__init__.py
@@ -13,25 +13,53 @@ hardware-specific regressions (see #640). Can also be forced via FLA_TILELANG=1.
 
 from __future__ import annotations
 
+import logging
 import os
+import shutil
+from functools import cache
+from pathlib import Path
 
 import torch
+from torch.utils import cpp_extension
 
 from fla.ops.backends import BaseBackend
 from fla.utils import IS_NVIDIA_HOPPER, TRITON_ABOVE_3_4_0, find_spec_cached
 
+logger = logging.getLogger(__name__)
+
 _TILELANG_AVAILABLE = find_spec_cached("tilelang") is not None
 
 
-class TileLangBackend(BaseBackend):
+@cache
+def _has_usable_nvcc() -> bool:
+    if shutil.which("nvcc") is not None:
+        return True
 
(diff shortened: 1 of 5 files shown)


mvanhorn:
Appreciate you taking the Triton fallback, @zhiyuan1i - no-nvcc setups can run again without hand-editing configs.

PR #1027, opened 2026-07-16
[Fix] Correct causal convolution benchmark dispatch and Ascend CI Python setup

sunyi0505:
# Summary

- Fix causal convolution benchmark provider matching so forward-backward cases execute backward correctly.
- Use the Ascend CI container’s bundled Python 3.11 instead of running actions/setup-python.

# Changes

- Reorder provider checks to prevent *_fwdbwd from being incorrectly matched as *_fwd.
- Handle the FLA convolution tuple output before calling backward.
- Remove the redundant Python setup step from Ascend A2 CI.


sunyi0505:

PR #1033, opened 2026-07-16
[Fix] Avoid concurrent LSE stores in value-split attention

### LINKED ISSUE #1030: [Bug] Value-split attention kernels issue concurrent LSE stores
## Describe the bug

Four value-split attention kernels allow every value program to store the same LSE element concurrently:

- parallel attention
- parallel NSA
- parallel WALL attention
- WALL attention decode

The LSE calculation depends on score-side state and is independent of `i_v`, so all value programs target the same address with the same computed result. The stores are non-atomic and unsynchronized.

## Steps to reproduce

Use a value dimension that produces `NV = ceil(V / BV) > 1`, such as `V=320` on the affected forward paths. For each query/head tile, programs `i_v=0`, `i_v=1`, and any later value programs all execute `tl.store(p_lse, ...)` to the same pointer.

The issue follows from the launch grid and store address: `i_v` appears in the output-value pointer but not in `p_lse`.

## Expected behavior

Exactly one value program should own each LSE store. Other value programs should write only their disjoint output-value tiles. Since LSE is independent of the value tile, selecting `i_v=0` as the owner preserves the computed result.

morluto:
## Summary

Parallel attention, NSA, WALL attention, and WALL decode launch one program per value tile. Every value program previously wrote the same LSE element concurrently, even though LSE is independent of the value tile.

Restrict each LSE store to `i_v == 0`, leaving one writer while preserving the computed value.

Fixes #1030.

## Ownership before and after

The value programs write disjoint output tiles but share one LSE address for each query/head tile:

```text
program        output ownership           old LSE writer        fixed LSE writer
i_v = 0        o[..., 0:BV]               yes                   yes
i_v = 1        o[..., BV:2*BV]            yes                   no
i_v = 2        o[..., 2*BV:3*BV]          yes                   no
```

Under the old mapping, all programs issued a non-atomic store to the same `p_lse`. There is no ordering or synchronization between those programs.

## Why one writer preserves the result

The stored LSE is derived from score-side state (`q`, `k`, gates or sink bias, `b_m`, and `b_acc`). The value tile affects the `v` load and `o` accumulator, but it does not affect `b_m` or `b_acc`. Every `i_v` therefore computes the same LSE.

For every positive `NV`, exactly one launched value program satisfies `i_v == 0`. The guard removes the competing stores without changing the value written.

## Regression coverage

- attention: the forced split-forward case consumes the resulting LSE during backward
- NSA: `V=320` coverage compares the split-value output and LSE against a single-value-tile launch
- WALL: a forced split-forward case covers backward, while `V=320` covers forward and decode
- decode remains forward-only because it has no backward path

## Test plan

- `python3 -m compileall -q fla/ops/attn/parallel.py fla/ops/nsa/parallel.py fla/ops/wall_attn/decode.py fla/ops/wall_attn/parallel.py tests/ops/test_attn.py tests/ops/test_nsa.py tests/ops/test_wall_attn.py`
- `ruff check fla/ops/attn/parallel.py fla/ops/nsa/parallel.py fla/ops/wall_attn/decode.py fla/ops/wall_attn/parallel.py tests/ops/test_attn.py tests/ops/test_nsa.py tests/ops/test_wall_attn.py`
- affected-file pre-commit hooks

## Compatibility

No API or numerical behavior changes are intended.

### DIFF
--- a/fla/ops/attn/parallel.py
+++ b/fla/ops/attn/parallel.py
@@ -177,7 +177,8 @@ def parallel_attn_fwd_kernel(
     b_o = b_o / b_acc[:, None]
     b_m += log2(b_acc)
     tl.store(p_o, b_o.to(p_o.dtype.element_ty), boundary_check=(0, 1))
-    tl.store(p_lse, b_m.to(p_lse.dtype.element_ty), boundary_check=(0,))
+    if i_v == 0:
+        tl.store(p_lse, b_m.to(p_lse.dtype.element_ty), boundary_check=(0,))
 
 
 @triton.jit
--- a/fla/ops/nsa/parallel.py
+++ b/fla/ops/nsa/parallel.py
@@ -276,7 +276,8 @@ def parallel_nsa_fwd_kernel(
     b_o = b_o / b_acc[:, None]
     b_m += log(b_acc)
     tl.store(p_o, b_o.to(p_o.dtype.element_ty), boundary_check=(0, 1))
-    tl.store(p_lse, b_m.to(p_lse.dtype.element_ty))
+    if i_v == 0:
+        tl.store(p_lse, b_m.to(p_lse.dtype.element_ty))
 
 
 @triton.heuristics({
--- a/fla/ops/nsa/parallel.py
+++ b/fla/ops/nsa/parallel.py
@@ -656,7 +656,10 @@ def parallel_nsa_bwd(
     G = HQ // H
     BS = block_size
     BK = max(triton.next_power_of_2(K), 16)
-    BV = min(128, max(triton.next_power_of_2(v.shape[-1]), 16))
+    # dq/dk accumulate softmax deltas reduced over the full value dim (delta from
+    # parallel_attn_bwd_preprocess spans all of V), so BV must span all of V (NV == 1)
+    # to avoid cross-program over-subtraction — same contract as fla/ops/attn/parallel bwd.
+    BV = max(triton.next_power_of_2(v.shape[-1]), 16)
     NV = triton.cdiv(V, BV)
 
     delta = parallel_attn_bwd_preprocess(o, do)
--- a/fla/ops/wall_attn/decode.py
+++ b/fla/ops/wall_attn/decode.py
@@ -153,7 +153,8 @@ def parallel_wall_attn_decode_kernel(
     b_o = b_o / b_acc[:, None]
     b_m += log2(b_acc)
     tl.store(p_o, b_o.to(p_o.dtype.element_ty), boundary_check=(0, 1))
-    tl.store(p_lse, b_m.to(p_lse.dtype.element_ty), boundary_check=(0,))
+    if i_v == 0:
+        tl.store(p_lse, b_m.to(p_lse.dtype.element_ty), boundary_check=(0,))
 
 
 def parallel_wall_attn_decode(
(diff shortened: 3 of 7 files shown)


zhiyuan1i:
Pushed a commit (ac7a4206) fixing the root cause your new value-split test exposed on H100 CI.

The problem runs deeper than the concurrent LSE stores: in `parallel_nsa_bwd`, both the dq and dkv kernels combined their value tile's partial dp with the full-V delta, and the per-tile outputs were summed on the host — so delta was over-subtracted NV-1 times whenever NV > 1 (V > 128), silently corrupting dq/dk (dq maxdiff 22.6 vs reference at V=320, matching this PR's CI failure exactly). dv is unaffected (per-tile disjoint stores).

Fix: BV now spans all of V in the backward (NV == 1), so one program reduces over the full value dim — same contract as `fla/ops/attn/parallel.py` bwd. The forward split is untouched.

Verified on H200: V=320 repro dq/dk error ratios go from 1.06/0.91 to <8e-4; your regression test plus a varlen variant pass; full `tests/ops/test_nsa.py` 47 passed. Your LSE-store guards are orthogonal to this and can stay as-is.

morluto:
brilliant thanks!

zhiyuan1i:
Follow-up on the test-ops stall: it is not a failing test but a **compile-time stall**. `test_wall_attn.py` sets `TRITON_F32_DEFAULT=ieee` process-wide; the V=320 fwd case needs BV=256 tiles, and under IEEE fp32 dots each of the 24 autotuned configs costs anywhere from 6s to 5+ minutes of ptxas time (measured on H200: BT=128/BV=256 configs = 82-300+s each, i.e. >29 min per swept key; every T/window variant re-sweeps). Files first, config-level bill of health second, so the job reads as a hang.

Pushed two commits:

- 65c14fc9 `[Fix] Prune oversized wall-attn fwd autotune tiles` — drops BT×BV>16384 configs (the 128 KiB fp32-accumulator ones that must spill and can never win) via early_config_prune. V≤128 sweeps are unchanged. With this, even the giant V=320 case completes (full test_wall_attn.py: 30 passed in 1h39m vs. previously aborted at 4h+).
- 1667bed9 `[Test] Exercise wall fwd value-split with a small BV-forced case` — replaces the V=320 case with V=128 + forced BV=64 (same NV=2 split-path and LSE single-writer coverage at a fraction of the compile cost): full test_wall_attn.py now finishes in **30 passed / 37m52s** on H200.

Longer-term suggestions (not in this PR): switch the wall tests to tf32x3 to kill the ieee compile tax at the source, and drop `T` from the autotune key so shapes stop re-sweeping per sequence length.

morluto:
Follow-up on the longer-term items mentioned above:

- #1041 reduces repeated compilation by reusing Wall autotune results across similar sequence and cache lengths. Exact lengths still control the kernel’s loops and masks; only the configuration-selection key is bucketed and cached.
- #1040 separates tests by purpose. Numerical-reference tests remain on IEEE precision, while decode, streaming, and cache-structure tests get a guarded alternative precision mode. IEEE remains the default.

Both build on #1033 and will be rebased after it lands.

While investigating wide value dimensions, I also opened #1042 for a related backward-pass limitation. Wall backward currently avoids incorrect split gradients by forcing one tile to span the entire value dimension. #1042 instead lets backward use bounded value tiles: each tile computes its own contribution to the softmax correction, and those partial gradients are reduced once afterward. This preserves the full-width gradient without subtracting the same correction once per tile. That fix does not depend on #1033 and can be reviewed separately.

Issue #942, opened 2026-06-09
[RFC] Roadmap: NPU support for flash-linear-attention


zheliuyu:
Thanks to the flash-linear-attention community for merging our native NPU support PR.

This roadmap outlines upcoming plans for native NPU support. Feedback and discussion are welcome.

## Native NPU Support in flash-linear-attention

Native support is built on top of flash-linear-attention's dispatch system:

- https://github.com/fla-org/flash-linear-attention/pull/741

We have validated this approach for NPU adaptation work:

- https://github.com/fla-org/flash-linear-attention/pull/927

The following diagram illustrates the flow for adapting kernels under `fla/modules/`:

```
User / model code (unchanged)
        │
        ▼
fla/modules/<module>.py              # Public API + @dispatch('modules') hooks only
        │
        ▼
fla/ops/backends                     # Runtime backend selection (BackendRegistry)
        │
        ▼
fla/modules/backends/triton_ascend/  # NPU Triton kernels + launchers
        ├── __init__.py              # TritonAscendBackend registration
        ├── rotary.py
        ├── fused_cross_entropy.py
        ├── fused_linear_cross_entropy.py
        ├── activations.py
        ├── fused_kl_div.py
        ├── layernorm.py
        ├── grpo.py
        └── causal_conv1d.py
```

## Dependency Installation

Note that flash-linear-attention already supports other devices such as XPU and AMD:

- https://github.com/fla-org/flash-linear-attention/pull/938
- https://github.com/fla-org/flash-linear-attention/pull/937

## NPU CI Setup

Future native adaptations will introduce a substantial amount of `triton-ascend` code, making NPU CI especially important. We plan to use the CI tooling provided by `ascend-gha-runners`:

- https://github.com/fla-org/flash-linear-attention/pull/944
- https://github.com/fla-org/flash-linear-attention/pull/973
- https://github.com/fla-org/flash-linear-attention/pull/987

Reference：

- https://github.com/ascend-gha-runners/docs/blob/main/docs/user-manual-gha-en.md
- https://github.com/ascend-gha-runners/docs/blob/main/docs/user-manual-gha-zh.md

This is our top priority at the current stage, and work is still in progress.

## Adaptation Order

At a high level, adaptation work is organized as follows:

- [x] `fla/modules/`
    - https://github.com/fla-org/flash-linear-attention/pull/927
    - https://github.com/fla-org/flash-linear-attention/pull/992
[...]

PR #1036, opened 2026-07-17
[Perf] Optimize triton-ascend L2Norm with row tiling on Ascend NPU

sunyi0505:
## Summary
- Rewrite Ascend L2Norm fwd/bwd from one-row-per-program kernels to `[BT, BD]` row-tiled kernels, with `BT` chosen via `compute_row_tile_block_size` under UB constraints (capped at 128).
- Recalibrate peak-memory multipliers (`fwd: 6→2`, `bwd: 8→4`) so small-D shapes can use larger `BT` without UB overflow under Ascend multi-buffering.
- Launch with `T_OFFSET` + `iter_axis_launch_chunks` instead of host-side tensor slicing; mark `T` as `do_not_specialize`; drop unused `eps` from the bwd wrapper/backend dispatch.

Tested with Atlas 800T A3(X86)
image
baseline: https://github.com/fla-org/flash-linear-attention/issues/942#issuecomment-4978264807

gdn:
image

### DIFF
--- a/fla/modules/backends/triton_ascend/__init__.py
+++ b/fla/modules/backends/triton_ascend/__init__.py
@@ -226,7 +226,7 @@ class TritonAscendBackend(BaseBackend):
         eps=1e-6,
     ):
         from fla.modules.backends.triton_ascend.l2norm import l2norm_bwd_npu
-        return l2norm_bwd_npu(y, rstd, dy, eps)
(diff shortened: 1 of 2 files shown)


sunyi0505:

zhiyuan1i (approved):

PR #1040, opened 2026-07-18
[Test] Add guarded Wall structural FP32 precision mode

### LINKED ISSUE #1038: [Feature Request] Scope Wall IEEE dots by test intent
## Feature Request

Scope Wall FP32 dot precision according to each test's purpose instead of setting IEEE precision process-wide for the entire Wall operator test module.

Numerical-oracle tests should continue to use IEEE. Structural and optimized-path consistency tests should have a separately validated precision mode without changing their shapes or tolerances.

## Motivation

Wall's eager-reference, backward-gradient, finite-difference, and numerical-stress tests are sensitive to precision and should retain IEEE dots. Large decode, streaming, and cache-layout tests primarily exercise path structure or consistency, but currently inherit the same process-wide setting.

Changing the environment alone is not sufficient in a long-lived test process: Wall autotune selections and compiled kernels created under one precision mode must not be reused after switching modes. Precision state must also be restored before other test modules run.

## Your Contribution

I can contribute explicit numerical-oracle and structural fixtures, precision-transition cache isolation, environment restoration, and validation using the existing test shapes and tolerances.

morluto:
> [!NOTE]
> Follow-up to #1033. This branch is based on that PR; until it lands, the intended follow-up is the final commit in this branch.

Closes #1038.

## Summary

Replace the process-wide IEEE setting in `test_wall_attn.py` with explicit precision fixtures based on each test's purpose.

Tests that act as numerical oracles continue to request IEEE FP32 dots:

- eager-reference forward and backward comparisons;
- forced value-split reference coverage;
- gradient-existence coverage;
- finite-difference gate-gradient checks;
- scalar-gate reference checks;
- long-sequence numerical-stress coverage.

Structural and optimized-path tests use a separately configurable precision:

- decode-versus-prefill consistency;
- streaming decode;
- cache-layout coverage.

The structural setting remains IEEE by default. It can be evaluated with `tf32x3` through `FLA_TEST_WALL_STRUCTURAL_F32_PRECISION=tf32x3`; candidate tests skip when `IS_TF32_SUPPORTED` is false. No tolerance, shape, or production-kernel behavior changes.

## Correctness boundary

This change does not assume that `tf32x3` is numerically interchangeable with IEEE for every Wall operation. Tests that determine numerical correctness continue to request IEEE explicitly. The configurable precision applies only to tests classified as structural or optimized-path consistency checks, and those tests retain their existing shapes and tolerances.

Whether `tf32x3` satisfies those tolerances is an execution result, not an algebraic guarantee.

## Precision-isolation invariant

Represent the relevant process state as $(E,C)$, where $E$ is `TRITON_F32_DEFAULT` and $C$ contains Wall's process-local autotune and compiled-kernel caches.

Changing precision to $P$ performs

$$
(E,C)\longrightarrow(E,\varnothing)\longrightarrow(P,\varnothing).
$$

The next Wall launch can therefore only populate the caches under precision $P$. Before another precision transition, those caches are cleared again.

Module teardown performs

$$
(P,C_P)\longrightarrow(P,\varnothing)\longrightarrow(E_{\mathrm{original}},\varnothing).
$$

Thus, for every test requesting precision $P$:

1. `TRITON_F32_DEFAULT` equals $P$ before its Wall kernel launches;
2. no Wall autotune selection or compiled kernel from another precision mode remains reusable;
3. the incoming environment is restored without retaining kernels compiled under the module's precision settings.

## Test plan

- `pre-commit run --files tests/ops/test_wall_attn.py`
- `python3 -m py_compile tests/ops/test_wall_attn.py`
- `python3 -m pytest --collect-only -q tests/ops/test_wall_attn.py` — 30 tests collected
- `FLA_TEST_WALL_STRUCTURAL_F32_PRECISION=tf32x3 python3 -m pytest --collect-only -q tests/ops/test_wall_attn.py` — the same 30 tests collected
- invalid-precision check — collection fails with the intended error
- cache-transition and environment-restoration sentinel check

## Runtime status

The branch adds a guarded candidate mode but keeps IEEE as the default. It does not claim that the `tf32x3` candidate already passes the complete runtime suite.

## Breaking changes

None. This is a test-only change; production APIs and kernel precision defaults are unchanged.

### DIFF
--- a/tests/ops/test_wall_attn.py
+++ b/tests/ops/test_wall_attn.py
@@ -14,13 +14,78 @@ import torch
 from fla.ops.utils.constant import RCP_LN2
 from fla.ops.utils.cumsum import chunk_global_cumsum
 from fla.ops.wall_attn import build_wall_kv_cache, naive_wall_attn, parallel_wall_attn, parallel_wall_attn_decode
+from fla.ops.wall_attn.decode import parallel_wall_attn_decode_kernel
+from fla.ops.wall_attn.parallel import (
+    parallel_wall_attn_bwd_kernel_dkv,
+    parallel_wall_attn_bwd_kernel_dq,
+    parallel_wall_attn_fwd_kernel,
+)
 from fla.utils import assert_close, device
 
-# Wall's log-space `R` factoring and the gate-gradient reverse-cumsum are sensitive
-# to TF32 matmuls (catastrophic cancellation for small gates), so force IEEE fp32
-# dots for these correctness checks -- matching the convention in `test_attn.py`.
-# Read by Triton at kernel-launch time, so setting it after imports is sufficient.
-os.environ['TRITON_F32_DEFAULT'] = 'ieee'
+# Keep IEEE as the default until the structural subset passes unchanged with tf32x3.
+_STRUCTURAL_F32_PRECISION_ENV = 'FLA_TEST_WALL_STRUCTURAL_F32_PRECISION'
+_STRUCTURAL_F32_PRECISION = os.environ.get(_STRUCTURAL_F32_PRECISION_ENV, 'ieee')
+_VALID_F32_PRECISIONS = {'ieee', 'tf32x3'}
+if _STRUCTURAL_F32_PRECISION not in _VALID_F32_PRECISIONS:
+    raise ValueError(
+        f'{_STRUCTURAL_F32_PRECISION_ENV} must be one of {sorted(_VALID_F32_PRECISIONS)}, '
+        f'but got {_STRUCTURAL_F32_PRECISION!r}',
+    )
+
+_WALL_KERNELS = (
+    parallel_wall_attn_fwd_kernel,
+    parallel_wall_attn_bwd_kernel_dq,
+    parallel_wall_attn_bwd_kernel_dkv,
+    parallel_wall_attn_decode_kernel,
+)
+
+
+def _clear_wall_kernel_caches():
+    """Drop precision-blind process-local caches before changing FP32 dot mode."""
+    for kernel in _WALL_KERNELS:
+        node = kernel
+        while node is not None:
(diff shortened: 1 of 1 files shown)


zhiyuan1i:
Thanks — the direction (IEEE only where numerics are asserted) is right, but closing for now: #1033 already fixed the compile stall, and this PR changes no default behavior. The tf32x3 path it enables currently emits illegal-instruction kernels on Triton 3.3 (verified while debugging #1033), so it can't run in CI until a Triton-side fix. Happy to revisit then.

morluto:
Thanks for the context, and for tracking this down while working on #1033. I took another look at the `tf32x3` path on a fresh H200 (SM90, driver 575.57.08, CUDA 12.8) using this PR head.

I could not reproduce the illegal-instruction failure in the cases I tried:

- With PyTorch 2.7.1 and Triton 3.3.1, both the FP32 MHA case and the split-value `V=320` case passed.
- After switching to Triton 3.3.0 in the same environment, the `V=320` case also passed, along with the FP32 GQA, small-shape, scalar-gate, and streaming-decode paths.
- The autotune sweeps exercised all runnable forward/decode configurations. Configurations that exceeded H200 shared-memory capacity were skipped normally, and I did not see an illegal instruction.

The structural comparisons were unchanged (`diff: 0`, `ratio: 0`) in these runs. This does not rule out the failure you saw, but it suggests it may depend on a particular test/configuration, driver or ptxas version, or cached kernel state rather than affecting every Triton 3.3 `tf32x3` kernel.

If you still have the exact failing test node or kernel configuration from the #1033 investigation, I would be happy to reproduce that case directly. Otherwise, would you be open to reopening #1040 and rerunning the full candidate suite before merge? IEEE would remain the default, so this would not change production behavior.

PR #1043, opened 2026-07-18
[Feature] support reusable final_state buffer in fused recurrent GDN

Bsgg1:


zhiyuan1i:
Thank you. I doubt the necessity of this, as FLA is primarily a training code repository.

lktr-wq:
Sharing one additional data point from an RTX 4060 Laptop GPU.

I tested the underlying input/output state-aliasing mechanism with preallocated buffers. At `T=1, B=16`, the cache-friendly standalone GDN benchmark improved by about 40%, with DRAM reads reduced by 77%. However, the gain fell to about 0.2% after cache interference, and the complete `GatedDeltaNetBlock` showed no stable improvement across two processes (-0.49% and +0.43%).

This suggests that buffer reuse can help substantially when the recurrent state remains cache-resident, but the model-level benefit is workload-dependent.

Setup, scripts, and results:
https://github.com/lktr-wq/fla-gdn-state-reuse-evaluation

I tested the mechanism against FLA commit `4d5515b`, not this PR branch's API implementation.

PR #1045, opened 2026-07-18
[Perf] Add opt-in TileLang RWKV6 intra kernel

markovchain-builder:
## Summary

Add an opt-in TileLang backend for the dense CUDA bf16 `D=64`, `chunk_size=64` RWKV6 forward-intra path. Unsupported dtypes, dimensions, layouts, sequence shapes, and varlen inputs are rejected by the backend verifier and continue through the existing Triton implementation.

The implementation, optimization loop, correctness validation, and performance evidence were completed autonomously by Argus. The target and H100 hardware constraint were selected by the operator.

## Test plan

- `python -m pytest tests/ops/test_backends.py -q`: 14 passed
- `FLA_TILELANG=1 python -m benchmarks.ops.verify --op chunk_rwkv6 --base b328e7c6`: full frozen gate, 13 passed
- The candidate run logged TileLang compilation for the accepted D64 path.

## Benchmark / NCU

- Hardware: NVIDIA H100 NVL
- Software: CUDA 12.6, PyTorch 2.7.1+cu126, Triton 3.3.1
- Workload: `B=8, T=1024, H=8, D=64`, bf16
- Forward: 0.199 ms -> 0.168 ms (`1.18x`)
- Forward + backward: 0.900 ms -> 0.747 ms (`1.21x`)

The backward-specific kernels remain Triton; the end-to-end forward + backward improvement comes from the faster TileLang forward-intra stage. NCU hardware counters were unavailable because the host denied profiling permissions, so the optimization was supported by CUDA timing and PyTorch profiler evidence instead.

Several larger fallback-only benchmark rows encountered OOM because other processes occupied most of the shared GPU memory. This did not affect the full correctness gate or the reported D64 row.

## Breaking changes

None. The backend is disabled by default and requires `FLA_TILELANG=1`.

### DIFF
--- a/fla/ops/rwkv6/backends/__init__.py
+++ b/fla/ops/rwkv6/backends/__init__.py
@@ -1,0 +1,17 @@
+# Copyright (c) 2023-2026, Songlin Yang, Yu Zhang, Zhiyuan Li
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+# For a list of all contributors, visit:
+#   https://github.com/fla-org/flash-linear-attention/graphs/contributors
+
+"""RWKV6 backends."""
+
+from fla.ops.backends import BackendRegistry, dispatch
+from fla.ops.rwkv6.backends.tilelang import RWKV6TileLangBackend
+
+rwkv6_registry = BackendRegistry("rwkv6")
+rwkv6_registry.register(RWKV6TileLangBackend())
+
+
+__all__ = ['dispatch', 'rwkv6_registry']
--- a/fla/ops/rwkv6/backends/tilelang/__init__.py
+++ b/fla/ops/rwkv6/backends/tilelang/__init__.py
@@ -1,0 +1,94 @@
+# Copyright (c) 2023-2026, Songlin Yang, Yu Zhang, Zhiyuan Li
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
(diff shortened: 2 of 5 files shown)


zhiyuan1i:
Thanks for the kernel — the dispatch/verifier gating is clean, and the generated code is race-free (I compiled it with tilelang 0.1.9 and reviewed the generated CUDA: all shared-memory RAW/WAR pairs are properly separated by barriers, no async-copy hazards).

One numerical issue needs fixing before merge, though.

In the off-diagonal path the decay is applied as two independent factors:
- `q_frag = q * exp2(ge) * scale` (chunk_fwd_intra.py:81-84)
- `k_shared = k * exp2(-gi)` (chunk_fwd_intra.py:87-90)

`gi`/`ge` are sequence-global cumulative log2 decays, not chunk-local. For long sequences with strong gates, `exp2(-gi)` overflows fp32 while `exp2(ge)` underflows, giving `0 * inf = NaN` — even though the intended product `exp2(ge - gi)` is small and well-behaved.

The Triton reference avoids this by centering both factors at a block-local reference point `gn = gi[bos + i_t*BT + i_i*BC - 1]` (fla/ops/rwkv6/chunk.py:177-188): `q * exp2(ge - gn)` × `k * exp2(gn - gi)`, which keeps every factor bounded by the intra-block decay range. Please apply the same centering here — it is a small change (load `gi` at the position just before the row sub-block and subtract it from both exponents) and matches the reference semantics exactly.

PR #1048, opened 2026-07-20
[Perf] Fuse ds in Ascend chunk_bwd_dqkwg to avoid recomputing do@v.T

sunyi0505:
# Summary
- Refactor Ascend Triton `chunk_bwd_kernel_dqkwg_npu` so each BC-tiled `(r, c)` block computes `ds = do @ v.T` once and reuses it for both `dq` (`ds @ k`) and `dk` (`ds.T @ q`).
- Split `dw = -dv @ h` into an independent V-loop (no longer tied to the old shared `h`/`dh` pass), then finalize `dk` as gated inter-chunk `v @ dh` plus the fused intra-chunk contribution from `dk_f32` scratch.
- Drop unused kernel arg `B` from the NPU dqkwg launch path.
## Motivation
Previously the backward path computed `do @ v.T` separately in the `dk` and `dq` loops. Fusing that matmul cuts redundant Cube work and should improve Ascend NPU throughput for chunk backward dq/dk/(dw).

gdn:
9c96b1cfdaf7c854cfdf5c1962274832

### DIFF
--- a/fla/ops/common/backends/triton_ascend/chunk_o.py
+++ b/fla/ops/common/backends/triton_ascend/chunk_o.py
@@ -775,7 +775,6 @@ def chunk_bwd_kernel_dqkwg_npu(
     cu_seqlens,
     chunk_indices,
     scale,
-    B: tl.constexpr,
     T,
     H: tl.constexpr,
     HV: tl.constexpr,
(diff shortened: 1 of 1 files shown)


sunyi0505:

PR #1051, opened 2026-07-21
[Attn] Support for heterogeneous hybrid-attention configurations

Michael-RDev:
## Summary

This PR generalizes FLA’s hybrid-attention configuration so different groups of layers can use different standard-attention settings.

The existing single-dictionary form remains supported, while `config.attn` can now also accept a list of attention specifications.

This also centralizes hybrid-attention normalization, validation, defaults, and per-layer lookup that were previously duplicated across model families.

## Problem

FLA currently uses one `attn` dictionary for both layer selection and attention settings:

  ```python
  attn = {
      "layers": [3, 7, 11],
      "num_heads": 16,
      "num_kv_heads": 4,
      "window_size": 2048,
  } 
```

Every selected layer therefore receives the same setting. Several model configuration classes also repeat the same validation and default-handling logic.

This prevents configurations such as sliding-window attention at some depths and full attention at others without model-specific changes.

## Design

You add a shared module at `fla/models/hybrid.py` containing:

- HybridAttentionSpec
- HybridAttentionConfig
- normalize_hybrid_attention_config
- get_hybrid_attention_spec

The accepted forms are:

```python
# Form 1: None
attn = None

# Form 2: dictionary (single config)
attn = {
    "layers": [3, 7],
    "num_heads": 16,
}

# Form 3: list of dictionaries (multi configs)
attn = [
    {
        "layers": [3, 7],
        "num_heads": 16,
        "num_kv_heads": 4,
        "window_size": 2048,
    },
    {
        "layers": [11],
        "num_heads": 16,
        "num_kv_heads": 4,
        "window_size": None,
    },
]
```

Here are the defaults that are applied independently to every specification:
- num_kv_heads = num_heads
- qkv_bias = False
- window_size = None
- rope_theta = 10000.0

Normalization validates:
- required layers and num_heads fields.
- layer index types and bounds.
- duplicate layers inside one specification.
- overlapping layers across specifications.
- positive head counts
- bool qkv_bias
- positive or None window sizes.
- positive finite RoPE theta values.

Unknown keys are preserved for forward compatibility

The original outer representation is retained:
- dictionary input remains a dictionary
- list input remains a list
- None remains None.

Per-layer resolution occurs only during block construction:

```python
attn_spec = get_hybrid_attention_spec(
    config.attn,
    layer_idx=layer_idx,
)
```

The existing Attention class is then constructed from that specification. No lookup is added to model forward methods.

## Model Integration

The shared configuration schema is now used by 21 configuration classes:

- ABC
- Comba
- DeltaNet
- DeltaFormer
- Gated DeltaNet
- Gated DeltaProduct
- GLA
- GSA
- HGRN
- HGRN2
- KDA
- LightNet
- Linear Attention
- MesaNet
- Mom
- Raven
- RetNet
- Rodimus
- RWKV6
- RWKV7
- Samba

All existing hybrid-selection block constructors now use shared per-layer lookup.

DeltaFormer is configuration-only because its modeling implementation doesn't currently select through `config.attn`

## Compatability

- The existing dictionary API remains supported.
- The public field remains `config.attn`
- Dictionary configurations remain  in saved JSON
- List configurations remain in shape of list.
- Existing block attributes such as `self.attn` and `self.mixer` are unchanged.
- Attention constructor arguments sets are unchanged.
- State-dict parameter names are unchanged
- No forward methods were changed
- No cache or generation code was changed.
- No attention kernels or operator mathematics were changed.

Samba includes a narrow compatibility path for historical shallow-model JSON that stored the complete depth-18 default attention plan.

Explicit plans remain strictly range-validated

### General Token-Mixer Plugin Framework

A general token-mixer registry or class hierarchy was considered. This would allow each layer to select and configure its mixer through a shared plugin interface.

This approach was rejected because it would:

- substantially expand the public API.
- increase configuration complexity.
- introduce additional checkpoint compatibility and migration risk.
- exceed the scope required to support heterogeneous standard-attention settings.

### JSON-Serializable Attention Specifications

A list of plain, JSON-serializable attention specifications was selected.

This approach supports different standard-attention settings across groups of layers while preserving the existing `config.attn` field, serialization behavior, model structure, and checkpoint compatibility.

## Tests

Added tests/models/test_hybrid_attention.py covering:

- None, dictionary, and list normalization.
- independent default handling
- preservation of unknown keys
- invalid fields and useful error messages.
- duplicate and overlapping layer detection
- empty plans and empty layer collections.
- per-layer lookup
- all 21 compatible configuration classes
- dictionary/list serialization round trips
- legacy Samba configuration loading.
- GLA, Gated DeltaNet, and Samba block construction
- local and full attention in one model
- dictionary versus one-item-list architecture equivalence
- state-dict key and tensor-shape equivalence.
- deterministic forward and gradient equivalence on supported GPU hardware.

Local validation:

- Pre-commit hooks passed for the feature files.
- Python compilation passed.
- git diff --check passed.
- Standalone normalization and serialization checks passed for all 21configuration classes.
- Source compatibility checks confirmed unchanged constructor parameter order, forward methods, module assignments, and Attention argument sets.

The pytest suite could not be collected locally because my macOS M1 environment does not support Triton :( 

Therefore, GPU numerical and gradient tests were therefore not run locally.

## Breaking changes

None.

## Limitations
This change supports multiple explicit attention specifications but doesn't add:

- periodic selector syntax
- learned routing
- parallel mixer composition
- new attention implementations
- cache formats
- generation changes
- Titans-or Hamba-specific modules
- a token-mixer registry or plugin framework

GQA divisibility validation remains outside this change while PR#1032 is still active.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,7 @@ All of the pretrained models currently available can be found in [`fla-hub`](htt
 
 `fla` provides a flexible method to incorporate standard attention layers into existing linear attention models.
 This is easily achieved by specifying the `attn` argument in the model configuration.
+The original dictionary form applies one shared attention specification to every listed layer.
 
 For example, to create a 2-layer Samba model with one Mamba layer followed by one local attention layer, using a sliding window size of 2048:
 
@@ -420,6 +421,35 @@ SambaForCausalLM(
 ```
 
 </details>
+
+To use different attention settings at different depths, pass a list of specifications. For example, this six-layer Samba model uses local attention at layers 1 and 3, full attention at layer 5, and the native Mamba mixer at layers 0, 2, and 4:
+
+```py
+>>> config = SambaConfig(
+...   num_hidden_layers=6,
+...   attn=[
+...     {
+...       'layers': [1, 3],
+...       'num_heads': 18,
+...       'num_kv_heads': 18,
+...       'qkv_bias': False,
+...       'rope_theta': 10000.,
+...       'window_size': 2048,
+...     },
+...     {
+...       'layers': [5],
+...       'num_heads': 18,
+...       'num_kv_heads': 18,
+...       'qkv_bias': False,
+...       'rope_theta': 10000.,
+...       'window_size': None,
+...     },
+...   ],
+... )
+>>> model = AutoModelForCausalLM.from_config(config)
+```
+
+Each specification is normalized independently. Layers omitted from the plan retain the model's native linear-attention, recurrent, or state-space mixer.
 
 During inference, you **DO NOT** need to revise anything for generation!
 The model will produce output as-is, without any need for additional configurations or modifications.
--- a/fla/layers/gdn2.py
+++ b/fla/layers/gdn2.py
@@ -131,11 +131,18 @@ class GatedDeltaNet2(nn.Module):
         # maps each value head to its qk head via `i_h = i_hv // (HV // H)`. num_v_heads <
         # num_heads would divide by zero there.
         if self.num_v_heads < self.num_heads:
-            raise ValueError(f"num_v_heads={self.num_v_heads} must be >= num_heads={self.num_heads}.")
+            raise ValueError(
+                f"num_v_heads={self.num_v_heads} must be >= num_heads={self.num_heads}."
+            )
         if self.num_v_heads % self.num_heads != 0:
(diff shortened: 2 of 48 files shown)


Michael-RDev:
Apologies for some files being changed due to my neovim formatter again

PR #1053, opened 2026-07-22
[KDA] Fuse chunk_kda forward kernels for ~30% D64 speedup on B200

nssmd:
## Summary

This PR fuses several kernels in the **`chunk_kda`** (Kimi Delta Attention) forward
pipeline to remove redundant kernel launches and HBM round-trips, giving a
correctness-preserving, memory-neutral speedup on NVIDIA B200 (sm_100).

Each change follows the same mechanism: an intermediate tensor that the baseline
stores to HBM and reloads in a separate launch is instead produced inside the
upstream kernel and consumed directly.

## Changes

1. **Paired q/k L2-norm fusion (forward + backward)** — `fla/modules/l2norm.py`, `fla/ops/kda/chunk.py`
   The baseline issues two separate `l2norm` launches (for `q` and `k`) in each
   direction. `l2norm_fwd_pair` / `l2norm_bwd_pair` fuse each pair into one kernel,
   halving launch count and normalization HBM traffic.

2. **Forward inter-solve recompute-epilogue fusion** — `fla/ops/kda/chunk_intra.py`
   `chunk_kda_fwd_kernel_inter_solve_fused` already holds the solved `Akk` in
   registers; computing `w/u/kg` in an epilogue removes the separate
   `recompute_w_u_fwd_kda_kernel` launch and its `Akk` HBM reload.

3. **Cumsum-into-intra producer fusion** — `fla/ops/kda/chunk_fwd.py`, `fla/ops/kda/chunk_intra.py`
   The chunk-local cumulative gate `g` is computed inside the intra-subchunk
   producer instead of a standalone `chunk_local_cumsum_vector_kernel` launch, and
   stored once for reuse.

4. **Blackwell autotune guard** — `fla/ops/kda/chunk_bwd.py`
   Restricts the `chunk_kda_bwd_kernel_dAv` autotune space on sm_100 to avoid an
   illegal-memory-access observed with `BK==32` and `num_warps != 2`.

## Benchmark

- **Op / shape:** `chunk_kda`, `B8_T1024_H8_D64` (D64), bf16, forward and forward+backward.
- **Hardware:** NVIDIA B200 (sm_100).
- **Method:** paired baseline-vs-candidate measurement on the same GPU
  (`geomean_speedup = baseline_latency / candidate_latency`); atol 1e-2 / rtol 2e-2;
  peak memory tracked as a ratio.

| Optimization | Speedup | Notes |
|---|---|---|
[...]

PR #1054, opened 2026-07-22
[KDA] Fuse chunk_kda forward kernels for ~30% D64 speedup on B200

miniH6:
## Summary

This PR fuses several kernels in the **`chunk_kda`** (Kimi Delta Attention) forward
pipeline to remove redundant kernel launches and HBM round-trips, giving a
correctness-preserving, memory-neutral speedup on NVIDIA B200 (sm_100).

Each change follows the same mechanism: an intermediate tensor that the baseline
stores to HBM and reloads in a separate launch is instead produced inside the
upstream kernel and consumed directly.

## Changes

1. **Paired q/k L2-norm fusion (forward + backward)** — `fla/modules/l2norm.py`, `fla/ops/kda/chunk.py`
   The baseline issues two separate `l2norm` launches (for `q` and `k`) in each
   direction. `l2norm_fwd_pair` / `l2norm_bwd_pair` fuse each pair into one kernel,
   halving launch count and normalization HBM traffic.

2. **Forward inter-solve recompute-epilogue fusion** — `fla/ops/kda/chunk_intra.py`
   `chunk_kda_fwd_kernel_inter_solve_fused` already holds the solved `Akk` in
   registers; computing `w/u/kg` in an epilogue removes the separate
   `recompute_w_u_fwd_kda_kernel` launch and its `Akk` HBM reload.

3. **Cumsum-into-intra producer fusion** — `fla/ops/kda/chunk_fwd.py`, `fla/ops/kda/chunk_intra.py`
   The chunk-local cumulative gate `g` is computed inside the intra-subchunk
   producer instead of a standalone `chunk_local_cumsum_vector_kernel` launch, and
   stored once for reuse.

4. **Blackwell (sm_100) autotune workaround** — `fla/ops/kda/chunk_bwd.py`
   On Blackwell, autotuning `chunk_kda_bwd_kernel_dAv` hits an illegal memory access when it
   explores `BK==32` with `num_warps != 2`. This prunes those configs **on Blackwell only**,
   so autotuning stays safe. It is a **workaround, not a root-cause fix** for the OOB, and the
   Hopper autotune space is left exactly as upstream (`num_warps == 2` guard untouched). Happy
   to drop this from the PR or replace it with a proper fix once the OOB is root-caused.

## Benchmark

- **Op / shape:** `chunk_kda`, `B8_T1024_H8_D64` (D64), bf16, forward and forward+backward.
- **Hardware:** NVIDIA B200 (sm_100).
- **Method:** paired baseline-vs-candidate measurement on the same GPU
  (`geomean_speedup = baseline_latency / candidate_latency`); atol 1e-2 / rtol 2e-2;
  peak memory tracked as a ratio.

| Optimization | Speedup | Notes |
|---|---|---|
| Paired q/k L2-norm fusion | **+5.6%** | N=10 median (independent re-verify +8.1%) |
| Forward inter-solve recompute-epilogue fusion | **+7.27%** | N=10 median |
| Cumsum-into-intra producer fusion | **+10.4% increment** | cumulative with the above = **+17.66%**, N=10 |
| **All three combined** | **+29.93% geomean** | single paired run, correctness PASS, `max_mem_ratio` 1.00, 0 CUDA errors |

Each individual optimization is **N≥10** certified (correctness PASS, memory-neutral,
median ≥ 1.05, cold-cache repeat). The **+29.93% combined** figure is from a single
paired verification run (fwd + fwd+bwd); a full N≥10 certification of the combined
stack is expected to land around ~+25–30%.

## Notes

- Branched from `ccb0ff94`; the diff touches only the 5 files above (+432/-32).
- Correctness is checked against the current PyTorch reference for `chunk_kda`.
- Happy to rebase onto the latest `main` and split into smaller PRs if preferred.

### DIFF
--- a/fla/modules/l2norm.py
+++ b/fla/modules/l2norm.py
@@ -106,6 +106,91 @@ def l2norm_fwd_kernel(
 
 
 @fla_cache_autotune(
+    configs=[triton.Config({"BT": BT}, num_warps=num_warps) for num_warps in [1, 2, 4] for BT in [16, 32, 64]],
+    key=["D", "NB"],
+    **autotune_cache_kwargs,
+)
+@triton.jit(do_not_specialize=["T"])
+def l2norm_fwd_pair_kernel(
+    x0,
+    x1,
+    y0,
+    y1,
+    rstd0,
+    rstd1,
+    eps,
+    T,
+    D: tl.constexpr,
+    BD: tl.constexpr,
+    NB: tl.constexpr,
+    BT: tl.constexpr,
+):
+    i_t = tl.program_id(0).to(tl.int64)
+    rows = i_t * BT + tl.arange(0, BT).to(tl.int64)
+    cols = tl.arange(0, BD).to(tl.int64)
+    row_mask = rows < T
+    col_mask = cols < D
+    mask = row_mask[:, None] & col_mask[None, :]
+    offsets = rows[:, None] * D + cols[None, :]
+
+    b_x0 = tl.load(x0 + offsets, mask=mask, other=0.0).to(tl.float32)
+    b_x1 = tl.load(x1 + offsets, mask=mask, other=0.0).to(tl.float32)
+    b_rstd0 = 1 / tl.sqrt(tl.sum(b_x0 * b_x0, 1) + eps)
+    b_rstd1 = 1 / tl.sqrt(tl.sum(b_x1 * b_x1, 1) + eps)
+    b_y0 = b_x0 * b_rstd0[:, None]
+    b_y1 = b_x1 * b_rstd1[:, None]
+
+    tl.store(y0 + offsets, b_y0.to(y0.dtype.element_ty), mask=mask)
+    tl.store(y1 + offsets, b_y1.to(y1.dtype.element_ty), mask=mask)
+    tl.store(rstd0 + rows, b_rstd0, mask=row_mask)
+    tl.store(rstd1 + rows, b_rstd1, mask=row_mask)
+
+
+@fla_cache_autotune(
+    configs=[triton.Config({"BT": BT}, num_warps=num_warps) for num_warps in [1, 2, 4] for BT in [16, 32, 64]],
+    key=["D", "NB"],
+    **autotune_cache_kwargs,
+)
+@triton.jit(do_not_specialize=["T"])
+def l2norm_bwd_pair_kernel(
+    y0,
+    rstd0,
+    dy0,
+    dx0,
+    y1,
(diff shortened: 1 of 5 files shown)


zhiyuan1i:
Thanks for the PR — the fusion strategy looks sound. One request on the benchmark side:

Could you add numbers for **D=128 shapes (e.g. H32 / H64, D128)**? D=64 has very limited practical use for KDA, so the ~30% geomean at B8_T1024_H8_D64 is hard to extrapolate: at D128 each chunk does substantially more compute, so kernel-launch overhead and HBM round-trips weigh much less, and the register pressure of the solve-epilogue fusion also grows. D128 numbers would make the real-world benefit much clearer.

If possible, **Hopper (H100) numbers** would also be valuable — FLA's CI runs on H100, so that's the platform where most users will actually validate and run this.

lbx154:
I helped re-evaluate this against the requested D128/Hopper cases and current `main`.

The existing CI artifact is from an NVIDIA H200 (despite the runner name). Two independent runs (#1053 and this PR) agree: across the four D128 shapes, the combined stack gives only 1.055–1.061x forward geomean and 1.001–1.002x forward+backward geomean. For B4/T4096/H64/D128 specifically, forward is 0.987–0.990x and forward+backward is 0.998–1.000x.

I also isolated the smaller pieces on a B200 after porting to current `main`:

- paired q/k L2 norm: D128 1.015x forward, 1.001x forward+backward; H64 is effectively unchanged;
- cumsum producer fusion: same-node D128 0.999x forward, 1.000x forward+backward; also effectively unchanged.

So I don't think the 432-line performance stack is justified for practical D128 workloads as-is, especially after the intervening kernel refactors. The one independently reproducible correctness issue is the SM100 backward-autotune illegal memory access. I extracted that into the focused #1109, narrowed the condition to SM100, and verified the full KDA test file on B200 (76 passed, 7 skipped).

My recommendation is to supersede this PR with #1109 rather than rebase the performance fusions without a practical-D128 win.

miniH6:
Thanks for the feedback. After re-evaluating the requested D128 workloads and the available Hopper-family CI results, I agree that the full performance-fusion stack is not justified without a meaningful practical-D128 training win, especially given the size of the diff and the intervening KDA refactors.

The independently reproducible SM100 backward-autotune issue has been extracted into the focused #1109. I’m closing this PR in favor of that fix. Thanks again for the review.

ExtReMLapin:
Ran a similar experiment but for SM80 on A100, same conclusion, not worth it.

PR #1055, opened 2026-07-22
[Perf] Rewrite Ascend chunk_delta_h fwd

OsirisDuan:
## Summary

1D core grid, inner V tiling autotune BV.

### DIFF
--- a/fla/ops/common/backends/triton_ascend/chunk_delta_h.py
+++ b/fla/ops/common/backends/triton_ascend/chunk_delta_h.py
@@ -12,8 +12,10 @@ from __future__ import annotations
 import torch
 import triton
 import triton.language as tl
+import triton.runtime.driver as driver
 
 from fla.ops.utils import prepare_chunk_indices, prepare_chunk_offsets
+from fla.ops.utils.cache import fla_cache_autotune
 from fla.ops.utils.op import exp2
 from fla.utils import input_guard
 from fla.utils.ascend_ub_manager import (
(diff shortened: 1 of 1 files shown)


zhiyuan1i:
Please check ascend CI, failed :)

OsirisDuan:
> Please check ascend CI, failed :)

Yeah, I am trying to fix that bug.

zhiyuan1i:
Thanks for the rewrite — the 1D persistent grid + inner V tiling direction looks good, and the core math matches the GPU reference. But the current A2 CI failure is a real bug that needs fixing before merge; a few more issues below.

**1. (Blocking) `T_max` is loop-carried and gets corrupted across tasks — this is the root cause of the red A2 CI.**

In the task loop, `T_max = 1 * T` (chunk_delta_h.py:116) sits inside the `tl.range` body while `T` is reassigned per-sequence at :122. Under Triton semantics, variables reassigned in the loop body are loop-carried: from the 2nd task onward, `T_max` holds the *previous* task's sequence length, so the gate head stride at :256 (`g_ptr = g + bos + i_h * T_max`) reads the wrong location for every varlen task after the first. Evidence: the only failing case is `test_chunk_varlen[...cu_seqlens[0,256,500,1000]-fp16]` with task_num = 3×8 = 24 > 20 (A2 cores) — every varlen case with task_num ≤ 20 passes, on both the 07-23 and 07-26 runs with the identical `o: NaN detected in tri` signature, while the same test is green on main. The wrong stride produces large positive `exp2(b_g_last - b_g)` exponents, which is very likely the actual source of the "overflow" symptoms addressed in the last three commits. Minimal fix: hoist `T_max = T` out of the task loop (or keep the per-sequence length in a fresh variable instead of reassigning `T`).

**2. (Should fix) int64 pointer casts were lost in the rewrite.**

The old code did `(boh * HV + i_h).to(tl.int64) * K * V` (cast before multiply); the rewrite reverted to int32-then-cast at :190, :219, :248, :265, :286, :316 — the same class of overflow we just fixed in bwd_dhu (#1060). At K=V=128 these overflow around 1–2M tokens. Please restore cast-before-multiply.

**3. (Should fix) `v_new_base` is computed outside the `SAVE_NEW_VALUE` guard** (:248).

With `save_new_value=False`, `v_new` is None and this fails at compile time. All current callers pass True so it's latent, but the old code had it inside the guard — please move it back or assert in the wrapper.

**4. (Perf evidence) The PR is titled [Perf] but carries no benchmark data** — please add before/after numbers for the rewrite (same shapes, note the baseline commit). The A2 benchmark CI runs with `--base ''` so it only proves it runs.

**5. (Note) Rounding points now deviate from the GPU reference on purpose?** GPU casts `b_h` to fp16 before the w·h dot and `b_v` to input dtype before the k·v outer product; the rewrite keeps everything fp32 and drops `allow_tf32=False`. Same formulas and arguably better precision — if intentional, a short comment would help. Also a reminder that the g=None (delta_rule) path still has no A2 CI coverage (test_delta.py isn't in the suite) — a small use_g=False case for fwd_h would close that.

zhiyuan1i (approved):
The NaN fix checks out — for the record: bugfix1 (T_total) fixed the varlen NaN but crashed bishengir-compile on two dense cases (SIGSEGV, specialization hint); bugfix2's hoisted `T_max = T` (chunk_delta_h.py:113) is the correct minimal fix and the 07-28 run is fully green (918 passed), including the previously failing H2-HV8 varlen case. The rewrite direction (1D core-grid, inner V tiling, G_T_CONTIG) matches our Ascend conventions, and cross-run benchmark CI comparison shows fwd ~1.14-1.31x / fwd+bwd ~1.22-1.31x on chunk_gdn dense shapes.

Two non-blocking follow-ups from my earlier review, please pick up in a follow-up PR:
1. int64-before-multiply casts are still missing at chunk_delta_h.py:190/:219/:248/:265/:286/:316 and the g_ptr offsets at :256/:258 — the old NPU code had `.to(tl.int64)` on exactly these offsets (h base overflows int32 at ~131K varlen tokens for HV=64, K=V=128). Same class as #1023/#1060.
2. Please add the official before/after benchmark table to the PR body (dense + varlen, per mr-readiness).

Note: main's #1060 also touched this file (bwd_dhu tiling) — no textual conflict, but worth a rebase-and-rerun to be safe.

PR #1057, opened 2026-07-23
[Ops] Add triton-ascend backend for attn_res kernel

zheliuyu:
## Summary

Add a Triton-Ascend AttnRes backend for Ascend NPU and follow up with memory fixes so the largest benchmark shape runs fwdbwd without OOM.

- Implement `triton_ascend` fused AttnRes kernels (L-chunk stacked sources, fp32 `o_mix` accumulation, D/N tiling via `ascend_ub_manager`)
- Register the NPU backend ahead of Gluon; keep GPU `fla/ops/attnres/fused.py` unchanged
- Add routing/parity tests, Ascend CI coverage, and `fused_attnres` benchmark registration
- Fix fwdbwd OOM at `L64 × T8192 × D8192` by detaching stacked benchmark residual views and lazy-allocating per-source `dv` gradients in L-chunks

## Test plan

- [x] `pytest tests/ops/test_attnres.py` — 16/16 passed on NPU
- [x] `python benchmarks/ops/run.py --op fused_attnres` — all fwd + fwdbwd shapes passed (incl. max shape)
- [x] Ascend A2 CI (`ascend-a2-ci.yml`)
- [x] Ascend A2 benchmark CI (`ascend-a2-benchmark-ci.yml`)

### DIFF
--- a/.github/workflows/ascend-a2-benchmark-ci.yml
+++ b/.github/workflows/ascend-a2-benchmark-ci.yml
@@ -119,7 +119,7 @@ jobs:
           print("matplotlib savefig smoke test: ok")
           PY
 
-      - name: Benchmark chunk_gdn, chunk_kda, and conv1d
+      - name: Benchmark chunk_gdn, chunk_kda, fused_attnres, and conv1d
         shell: bash
         env:
           FLA_BENCH_WARMUP_MS: "25"
@@ -137,19 +137,26 @@ jobs:
             --op chunk_kda --base '' --json benchmark_chunk_kda.json \
             > chunk_kda.log 2>&1 &
           pid_kda=$!
-          ASCEND_RT_VISIBLE_DEVICES=2 python benchmarks/modules/benchmark_conv.py \
+          ASCEND_RT_VISIBLE_DEVICES=2 python -m benchmarks.ops.run \
+            --op fused_attnres --base '' --json benchmark_fused_attnres.json \
+            > fused_attnres.log 2>&1 &
+          pid_attnres=$!
+          ASCEND_RT_VISIBLE_DEVICES=3 python benchmarks/modules/benchmark_conv.py \
             > conv1d.log 2>&1 &
           pid_conv1d=$!
 
           status=0
           wait "$pid_gdn" || status=1
           wait "$pid_kda" || status=1
+          wait "$pid_attnres" || status=1
           wait "$pid_conv1d" || status=1
 
           echo "========== chunk_gdn =========="
           cat chunk_gdn.log
           echo "========== chunk_kda =========="
           cat chunk_kda.log
+          echo "========== fused_attnres =========="
+          cat fused_attnres.log
           echo "========== conv1d =========="
           cat conv1d.log
           exit "$status"
@@ -162,6 +169,7 @@ jobs:
           path: |
             benchmark_chunk_gdn.json
             benchmark_chunk_kda.json
+            benchmark_fused_attnres.json
             conv_benchmark/
           if-no-files-found: ignore
           retention-days: 10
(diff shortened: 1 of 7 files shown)


zheliuyu:
@zhiyuan1i This pr is ready for review. Thanks.

zhiyuan1i:
Thanks for the solid work — I checked the kernel math against the GPU Triton reference formula-by-formula (fp32 accumulation points, checkpoint_level semantics, the softmax/rmsnorm backward identities) and everything matches. A few questions:

1. **`_flat_residual` silently detaches view inputs** (triton_ascend/fused.py): for any residual that is itself a view (e.g. slices of a stacked `[L,B,T,D]` buffer — exactly what the benchmark feeds), `r._base is not None` triggers `detach().requires_grad_(True)`, so backward writes dvs into detached temporaries and the caller's `.grad` silently stays empty. The GPU path keeps the graph via reshape+contiguous, so semantics diverge. If the detach is only to avoid materializing the stacked buffer's grad in the benchmark (the OOM fix), could it live in the benchmark's `post_init` instead of the op? In real usage residuals are upstream layer outputs, so silently dropping grads on view inputs could corrupt training.

2. **`L_OFFSET: tl.constexpr` recompiles per L-chunk**: each chunk launch passes a new `l0` (0, l_chunk, ...), and Triton caches one binary per constexpr value, so large shapes compile each kernel once per chunk. It only participates in pointer arithmetic — would a runtime scalar (e.g. via `do_not_specialize`) work, or is there a triton-ascend constraint?

3. **`attnres_bwd_kernel_dqdw_npu` reduces over N serially**: the grid is only `cdiv(D, BD)` programs, each looping over the full N=B*T. The GPU version tiles this with BN=1024-4096. Curious how much this shows up in the benchmark numbers — is a tiled reduction planned?

zheliuyu:
Thanks for the thorough review.

1. `_flat_residual` silently detaches view inputs

You're right that silently detaching view inputs in the op would diverge from the GPU path and could drop gradients on real training inputs (e.g. upstream layer outputs that are views).

**Changes made:**

- Removed the `detach().requires_grad_(True)` logic from `_flat_residual` in `fla/ops/attnres/backends/triton_ascend/fused.py`.
- Restored GPU-aligned semantics: `r.reshape(-1, D).contiguous()`, so the autograd graph is preserved for view inputs.
- Moved the benchmark OOM workaround into `_attnres_post_init` in `benchmarks/ops/registry.py`: each stacked `[L, B, T, D]` slice is split via `res[i].detach().clone().requires_grad_(True)`, so fwdbwd no longer materializes one giant grad buffer for the parent stack, while the op itself stays semantically correct.

2. `L_OFFSET: tl.constexpr` recompiles per L-chunk

Good catch. Since `L_OFFSET` only participates in pointer arithmetic (`g_l = L_OFFSET + o_l`), it does not need to be a compile-time constant.

**Changes made:**

- Changed `L_OFFSET` from `tl.constexpr` to a runtime scalar in the three L-chunk kernels (`attnres_fwd_p1_chunk_kernel`, `attnres_fwd_p2_chunk_kernel`, `attnres_bwd_dv_chunk_kernel`).
- Added `L_OFFSET` to `do_not_specialize`, following the same pattern as KDA's `NT_OFFSET` in `chunk_intra.py`.

This should avoid one kernel binary per L-chunk on first run for large `L`.

3. `attnres_bwd_kernel_dqdw_npu` reduces over N serially

Acknowledged. The current NPU kernel uses a grid of `cdiv(D, BD)` programs with a serial loop over `N = B * T`, while the GPU path tiles this dimension with `BN = 1024–4096`.

For this PR I kept the serial reduction as-is. On Ascend A2, the largest benchmark shape (`L64 × T8192 × D8192`) fwdbwd completes in ~10.5s with correct numerics. BN tiling for the `dqdw` reduction is planned as a follow-up optimization once this lands.

PR #1060, opened 2026-07-24
[Perf] Optimize Ascend chunk_delta_h bwd_dhu with dynamic BK/BV tiling and host gate precomp

sunyi0505:
## Summary

Add _select_bwd_dhu_tiles to pick (BK, BV) under a soft UB budget (tighter when gates are computed in-kernel), replacing the fixed BK=64 + _get_bv path for non–STATE_V_FIRST layouts.
Host-precompute fp32 g_exp / g_ratio / g_last_exp when T % BT == 0 (non-varlen); fall back to transposed g + in-kernel exp2 for unaligned T / varlen.
Fold gate×scale into do once per chunk so K-slabs skip repeated gating; update dv in place and drop the extra dv2 buffer.
Walk dh via chunk-stride pointer arithmetic and parameterize K-slab offsets by BK instead of hardcoded 64/128/192.

GDN:
d7d78f152a9cb0530bb0e3ae0d38f05a

### DIFF
--- a/fla/ops/common/backends/triton_ascend/chunk_delta_h.py
+++ b/fla/ops/common/backends/triton_ascend/chunk_delta_h.py
@@ -19,6 +19,7 @@ from fla.utils import input_guard
 from fla.utils.ascend_ub_manager import (
     ASCEND_MAX_GRID_DIM,
     compute_row_tile_block_size,
+    get_ub_manager,
     max_grid_axis_chunks,
 )
 
@@ -28,6 +29,11 @@ _FWD_H_MEM_MULT = 8.0
 _SAFETY_MARGIN = 0.80
 _FALLBACK_BV = 16
 _MAX_BV = 64
+_DHU_BT = 64
+# Soft UB over-admit for host-precomputed gates (live-range can beat peak estimate).
+_DHU_UB_SOFT = 1.15
+# Tighter fraction when in-kernel exp2(g) inflates live UB (~1.7× analytical).
+_DHU_UB_GATE_INLINE = 0.60
 
 
 def _get_bv(K: int, V: int) -> int:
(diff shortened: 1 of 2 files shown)


sunyi0505:

zhiyuan1i:
Nice optimization — I verified the math against the previous implementation (host-precomputed fp32 exp2 gates ≡ inline exp2, folding g_exp*scale into do is exact by dot-product linearity, in-place dv is safe since both callers rebind the returned buffer). Two requests before merge:

1. **int32 overflow in the new dh chunk addressing (should fix)**. `dh_chunk = dh + ((NT - 1) * DH_CS).to(tl.int64)` multiplies in int32 *before* the cast: T is `do_not_specialize`, so NT is a runtime int32 and `(NT-1) * HV * K * V` wraps past 2^31. With K=V=128, HV=64 a single sequence of T>131K tokens already overflows — long context is exactly where this kernel matters. The old code was safe (`i_t.to(tl.int64) * HV * K * V`). One-line fix: `dh + (NT - 1).to(tl.int64) * DH_CS`.

2. **Please cover the g=None path of this kernel (test gap)**. This PR puts new code on the g=None path too (in-place dv, the `b_do * scale` fold, the new tiling), but nothing exercises it: `test_chunk_gated_delta_rule_bwd_dhu` always passes a gate (no use_g parametrization), and the A2 CI does not run test_delta.py, so the delta_rule end-to-end path that routes here with g=None is untested. The lightweight fix: add a `use_g` parameter (True/False) to the existing `test_chunk_gated_delta_rule_bwd_dhu` with a g=None branch in the reference — no need to add whole new test files to CI.

sunyi0505:
> Nice optimization — I verified the math against the previous implementation (host-precomputed fp32 exp2 gates ≡ inline exp2, folding g_exp*scale into do is exact by dot-product linearity, in-place dv is safe since both callers rebind the returned buffer). Two requests before merge:
> 
> 1. **int32 overflow in the new dh chunk addressing (should fix)**. `dh_chunk = dh + ((NT - 1) * DH_CS).to(tl.int64)` multiplies in int32 _before_ the cast: T is `do_not_specialize`, so NT is a runtime int32 and `(NT-1) * HV * K * V` wraps past 2^31. With K=V=128, HV=64 a single sequence of T>131K tokens already overflows — long context is exactly where this kernel matters. The old code was safe (`i_t.to(tl.int64) * HV * K * V`). One-line fix: `dh + (NT - 1).to(tl.int64) * DH_CS`.
> 2. **Please cover the g=None path of this kernel (test gap)**. This PR puts new code on the g=None path too (in-place dv, the `b_do * scale` fold, the new tiling), but nothing exercises it: `test_chunk_gated_delta_rule_bwd_dhu` always passes a gate (no use_g parametrization), and the A2 CI does not run test_delta.py, so the delta_rule end-to-end path that routes here with g=None is untested. The lightweight fix: add a `use_g` parameter (True/False) to the existing `test_chunk_gated_delta_rule_bwd_dhu` with a g=None branch in the reference — no need to add whole new test files to CI.

Fixed. Thanks for your suggestions!

PR #1061, opened 2026-07-24
[Perf] Rewrite Ascend local_cumsum

OsirisDuan:
# Summary

Rewrite `chunk_local_cumsum_scalar_kernel_npu` for the Ascend NPU backend, replacing the 2D grid / offset-based launch with a 1D vector-core ring dispatch and adding UB-safe block size computation.

# Benchmark

python benchmarks/ops/run.py --op chunk_gdn

```
===========================================================================================================================
  Machine: Ascend910B3 | NPU: CANN 9.0.0-beta.2 | PyTorch 2.7.1+cpu
===========================================================================================================================
  fwd        B      T    H    D  op                                   main[cf0242f9](ms) 260724/l...[37ddc123](ms)  speedup
          -----------------------------------------------------------------------------------------------------------------
             1   8192   96  128  chunk_gdn                                       323.119                   321.845    1.00x
          -----------------------------------------------------------------------------------------------------------------
             2  16384   16  128  chunk_gdn                                       222.010                   220.702    1.01x
          -----------------------------------------------------------------------------------------------------------------
             4   2048   16  128  chunk_gdn                                        53.552                    53.430    1.00x
          -----------------------------------------------------------------------------------------------------------------
             4   4096   64  128  chunk_gdn                                       427.493                   425.630    1.00x
          -----------------------------------------------------------------------------------------------------------------
             8   1024    8   64  chunk_gdn                                        24.159                    23.963    1.01x
          -----------------------------------------------------------------------------------------------------------------
             8   2048   32  256  chunk_gdn                                       269.993                   268.865    1.00x
===========================================================================================================================
  fwdbwd     B      T    H    D  op                                   main[cf0242f9](ms) 260724/l...[37ddc123](ms)  speedup
          -----------------------------------------------------------------------------------------------------------------
             1   8192   96  128  chunk_gdn                                      1367.492                  1365.899    1.00x
          -----------------------------------------------------------------------------------------------------------------
             2  16384   16  128  chunk_gdn                                       986.000                   986.203    1.00x
          -----------------------------------------------------------------------------------------------------------------
             4   2048   16  128  chunk_gdn                                       235.600                   234.618    1.00x
          -----------------------------------------------------------------------------------------------------------------
             4   4096   64  128  chunk_gdn                                      1845.429                  1845.896    1.00x
          -----------------------------------------------------------------------------------------------------------------
             8   1024    8   64  chunk_gdn                                        72.681                    73.731    0.99x
          -----------------------------------------------------------------------------------------------------------------
             8   2048   32  256  chunk_gdn                                      2138.920                  2131.672    1.00x
===========================================================================================================================
```


OsirisDuan:
@zhiyuan1i Please review this PR.

fla/ops/utils/backends/triton_ascend/cumsum.py
```diff
@@ -185,49 +164,69 @@ def _launch_local_cumsum_vector(
...
+        if HEAD_FIRST:
+            ptr_s = tl.make_block_ptr(s + bos * H, (H, T), (T, 1), (0, i_block * BLOCK_T), (H, BLOCK_T), (1, 0))
+            ptr_o = tl.make_block_ptr(o + bos * H, (H, T), (T, 1), (0, i_block * BLOCK_T), (H, BLOCK_T), (1, 0))
+            b_s = tl.load(ptr_s, boundary_check=(0,)).to(tl.float32)
+            b_s = tl.reshape(b_s, (H, N_CHUNKS, CHUNK_SIZE))
+            b_s = tl.trans(b_s, (2, 0, 1))
+            b_o = tl.cumsum(b_s, axis=0, reverse=REVERSE)
+            if HAS_SCALE:
+                b_o *= scale
+            b_o = tl.trans(b_o, (2, 0, 1))
```
zhiyuan1i: **(Blocking) Wrong inverse permutation — `head_first` output is scrambled for any H>1.** After `tl.cumsum`, `b_o` is `(CHUNK_SIZE, H, N_CHUNKS)`; transposing again with `(2, 0, 1)` yields `(N_CHUNKS, CHUNK_SIZE, H)`, and the following `tl.reshape(b_o, (H, BLOCK_T))` then flattens head/chunk/position in the wrong order. The correct inverse is `tl.trans(b_o, (1, 2, 0))`. The old NPU kernel handled `head_first` correctly, so this is a regression on a public API flag that the verifier unconditionally accepts (no test covers it — that's why A2 CI is green).

fla/ops/utils/backends/triton_ascend/cumsum.py
```diff
@@ -185,49 +164,69 @@ def _launch_local_cumsum_vector(
...
-    if REVERSE:
-        b_z = tl.sum(b_s, axis=0)
-        b_o = -b_o + b_z[None] + b_s
-    if HAS_SCALE:
-        b_o *= scale
-    tl.store(p_o, b_o.to(p_o.dtype.element_ty), boundary_check=(0,))
+        if HEAD_FIRST:
+            ptr_s = tl.make_block_ptr(s + bos * H, (H, T), (T, 1), (0, i_block * BLOCK_T), (H, BLOCK_T), (1, 0))
+            ptr_o = tl.make_block_ptr(o + bos * H, (H, T), (T, 1), (0, i_block * BLOCK_T), (H, BLOCK_T), (1, 0))
+            b_s = tl.load(ptr_s, boundary_check=(0,)).to(tl.float32)
```
zhiyuan1i: **(Blocking) Wrong `boundary_check` axis in the head_first branch.** The tile here is `(H, BLOCK_T)` with offset `(0, i_block*BLOCK_T)`, so axis 0 (H) is always fully covered and axis 1 (T) is the one that can overrun when `T % BLOCK_T != 0`. `(0,)` is a no-op; this needs `(1,)`. As written, tail blocks read OOB — and since stride is `(T, 1)`, the overrun aliases the next head's rows.

fla/ops/utils/backends/triton_ascend/cumsum.py
```diff
@@ -185,49 +164,69 @@ def _launch_local_cumsum_vector(
...
+            b_s = tl.reshape(b_s, (N_CHUNKS, CHUNK_SIZE, H))
+            b_s = tl.trans(b_s, (1, 0, 2))
+            b_o = tl.cumsum(b_s, axis=0, reverse=REVERSE)
+            if HAS_SCALE:
+                b_o *= scale
+            b_o = tl.trans(b_o, (1, 0, 2))
+            b_o = tl.reshape(b_o, (BLOCK_T, H))
+
+        tl.store(ptr_o, b_o.to(ptr_o.dtype.element_ty), boundary_check=(0,))
```
zhiyuan1i: **(Blocking) Same axis issue on the store side — this is the memory-corruption half.** The shared store also uses `boundary_check=(0,)`, so the head_first branch writes out of bounds into the next head's output rows. Since the BTH branch genuinely needs axis 0 checked, this line should use `(0, 1)` (the extra check is a no-op for BTH).

fla/ops/utils/backends/triton_ascend/cumsum.py
```diff
@@ -31,9 +32,32 @@
...
+        safety_margin=_CUMSUM_SAFETY_MARGIN,
+        dtype_size=4,
+        fallback=_FALLBACK_BT_LOCAL,
+        min_block=chunk_size,
+    )
+    BT = (BT // chunk_size) * chunk_size
+    return max(BT, chunk_size)
+
+
+def get_npu_properties():
```
zhiyuan1i: Nit: third copy of `get_npu_properties` in the repo (`chunk_o.py`, `wy_fast.py` have the same), and it's re-queried on every call. `fla.utils._device.get_multiprocessor_count` is already `@cache`d and returns `num_vectorcore` on NPU — please reuse it or hoist a shared helper.

zhiyuan1i (changes requested):
Thanks for the rewrite — the direction matches our Ascend playbook (1D vector-core ring dispatch + contiguous `(BLOCK_T, H)` tiles instead of stride-H gathers, single launch instead of host grid-chunking), and the BTH dense/varlen paths check out with A2 CI green. Two blocking correctness regressions in the `head_first` branch (inline), plus a few asks:

1. The two `head_first` regressions (scrambled output for H>1, and out-of-bounds read/write from a wrong `boundary_check` axis) — please apply the fixes suggested inline and add a `head_first=True` case to `tests/ops/utils/test_cumsum.py` (current tests never exercise it, which is how this slipped through). Alternatively, reject `head_first` in the verifier so it falls back to the default impl.
2. Perf evidence: the chunk_gdn e2e numbers are all ~1.00x with no varlen and no kernel-level before/after. Per the repo's MR-readiness conventions, please add op-level evidence (`kernel_details`/`op_statistic` Duration for this kernel, dense + varlen) — that's where the contiguous-tile and single-launch wins should show up.
3. Caller-passed `chunk_indices` (chunk_size granularity) is reinterpreted as BLOCK_T-block indices; it only works because BLOCK_T is always a multiple of chunk_size. Please recompute indices with BLOCK_T explicitly, or document + assert this contract.

The H100 test-ops cancellation is unrelated to this diff (NPU-only file, never imported on CUDA; the new NPU-only CI gate will skip it after a rebase) — please rebase to make it green.

zheliuyu:
@OsirisDuan Hi — this has been sitting for a while. Are you still planning to continue work on this kernel?

fla/ops/utils/backends/triton_ascend/cumsum.py
OsirisDuan: Used `boundary_check=(1,) if HEAD_FIRST else (0,)` instead of the suggested `(0, 1)` to avoid redundant no-op checks — the two are functionally equivalent since axis 0 (H) is always fully covered in HEAD_FIRST and axis 1 (H) is always fully covered in non-HEAD_FIRST. Keeps load/store symmetry.

fla/ops/utils/backends/triton_ascend/cumsum.py
OsirisDuan: Fixed.

fla/ops/utils/backends/triton_ascend/cumsum.py
OsirisDuan: Fixed.

OsirisDuan:
> @OsirisDuan Hi — this has been sitting for a while. Are you still planning to continue work on this kernel?

Yes, this PR is being updated.

PR #1069, opened 2026-07-28
[Perf] A Tilelang implementation of DPLR

sablin39:
## Summary

A TileLang implementation of the DPLR chunk pipeline, registered as a
dispatch backend on `chunk_dplr_delta_rule` (`chunk_rwkv7` is an alias).
A verifier accepts a call only where the TileLang path is correct and
measurably faster; everything else falls back to the default Triton
implementation unchanged. On by default when `tilelang` is present;
`FLA_TILELANG=0` forces Triton. Context parallelism (`cp_context`) works
through FLA's existing boundary-state scheme.

## 1. Differences from the Triton implementation

**Structure**

- Forward: one fused kernel produces the chunk states `h` and the output
  `o` in a single sweep, replacing Triton's separate `fwd_h` / `fwd_o`
  passes and their global-memory round trip of `h`/`v_new`.
- Backward: one fused kernel runs the reverse `dh` scan together with its
  consumers (`dq`/`dk`/`dw`/`db`/`dv2`/`dgk_last`/`dh0`); `dh` is never
  materialized. Triton runs separate `bwd_dhu` / `dv` / `dqkwg` kernels.
- Gate cumsum: computed in-CTA from raw `gk` at `K = 64`; at `K = 128` a
  segmented fp32 scan kernel replaces a PyTorch cumsum chain that
  profiled at ~2.3–2.8 ms/iter of glue (h4096).
- Autograd: `torch.library.custom_op` + `register_fake` +
  `register_autograd` — opaque to `torch.compile`, no graph break.
  Triton uses `autograd.Function` under `@torch.compiler.disable`.
  Dead zero-grad fills for the 13 saved intermediates are disabled via
  `set_materialize_grads(False)` (~2.8 GB/step at h4096).
- `disable_recompute` keeps the Triton semantics (save vs. recompute);
  both policies are supported and benchmarked.

**Numerics** — same conventions as Triton (log2 gates, mid-chunk
gate-centering for tensor-core operands, fp32 state accumulation, fp32
`A_ab_inv`), except `A_ab`/`A_ak` are stored in fp16 (Triton: fp32),
validated within the existing tolerances. Context parallelism uses the
exact same boundary-state exchange as the Triton CP path.

Accuracy vs. the fp32 per-token recurrence (max err ratio over o, final
state, and all input grads; B8 T256 H32, `initial_state` + `dht`
backward, measured on sm_120):

| Config | Triton | TileLang |
| ------ | ------ | -------- |
| cs16 D64 bf16 | 0.0061 | 0.0053 |
| cs32 D64 bf16 | 0.0081 | 0.0053 |
| cs32 D128 bf16 | 0.0104 | 0.0058 |
| cs32 D64 fp16 | 0.0021 | 0.0019 |
| cs32 D128 fp16 | 0.0021 | 0.0018 |
| cs64 D64 bf16 (RWKV7 gates) | 0.0068 | 0.0060 |
| cs64 D128 bf16 (RWKV7 gates) | 0.0084 | 0.0084 |

TileLang is at or below Triton's error in every config (up to ~1.8×
closer at cs32 bf16); the fp16 `A_ab`/`A_ak` storage costs nothing
measurable (fp16 rows). Under sustained −5 gate saturation both stay
finite, but Triton's `dgk` err ratio is ~3× worse (0.51 vs. 0.17).

**Routing**

- Requires a caller-asserted gate bound — `safe_gate=True`, or a
  `lower_bound` that fits the chunk size (GDN2's convention): the
  A-stage's mid-chunk-centered `exp2` operands must stay in fp32 range.
  Unbounded calls stay on Triton, which handles arbitrary gates.
- The verifier rejects unsupported or measurably slower configurations
  (one log line per reason). For `chunk_size=64` it shares schedule
  arithmetic with the launcher, so acceptance implies launchability.

## 2. Benchmark results

bf16, `triton.testing.do_bench` (warmup 10, rep 50), `FLA_TILELANG=0`
vs. `=1` through the dispatched public op, idle GPUs: RTX PRO 6000
(sm_120, 99KB smem) and H800 (sm_90, 228KB smem). `dr` =
`disable_recompute`. † = verifier reroutes to Triton (~1.00× by
construction). Headline cells were additionally re-measured with
CUDA-event wall timing and profiler attribution — same ratios. Varlen
harness: each row splits 8192 tokens into N sequences with lengths
∝ `rand(N) + 0.35` (seed = 17 + N), `[1, 8192]` layout, `gk` drawn as
`-0.61·sigmoid(5x)` in fp32; the timing script is the same do_bench
harness for every table and is included in the PR discussion.

### PRO 6000 (sm_120)

Rectangular, cs32, T=4096 (`hidden` = H·D):

| Workload (hidden, H, D, B) | Mode    | Triton dr=0 | Triton dr=1 | TileLang dr=0 | TileLang dr=1 | Speedup dr=0 | Speedup dr=1 |
| -------------------------- | ------- | ----------- | ----------- | ------------- | ------------- | ------------ | ------------ |
| 1024, 16, 64, 8            | fwd     | 2.26        | 2.26        | 1.67          | 1.66          | 1.36×        | 1.36×        |
|                            | fwd+bwd | 7.79        | 5.87        | 5.85          | 4.26          | 1.33×        | 1.38×        |
| 1024, 8, 128, 8 †          | fwd     | 3.64        | 3.60        | 3.63          | 3.62          | 1.00×        | 1.00×        |
|                            | fwd+bwd | 10.56       | 7.30        | 10.53         | 7.33          | 1.00×        | 1.00×        |
| 2560, 40, 64, 8            | fwd     | 6.23        | 6.25        | 4.30          | 4.31          | 1.45×        | 1.45×        |
|                            | fwd+bwd | 21.26       | 15.91       | 15.14         | 11.18         | 1.40×        | 1.42×        |
| 2560, 20, 128, 8           | fwd     | 9.57        | 9.54        | 4.90          | 4.90          | 1.95×        | 1.95×        |
|                            | fwd+bwd | 27.90       | 19.19       | 19.66         | 12.95         | 1.42×        | 1.48×        |
| 4096, 64, 64, 8            | fwd     | 8.99        | 9.09        | 6.41          | 6.44          | 1.40×        | 1.41×        |
|                            | fwd+bwd | 31.37       | 23.69       | 22.95         | 17.09         | 1.37×        | 1.39×        |
| 4096, 32, 128, 8           | fwd     | 14.66       | 14.63       | 7.46          | 7.47          | 1.96×        | 1.96×        |
|                            | fwd+bwd | 42.78       | 29.53       | 31.02         | 20.51         | 1.38×        | 1.44×        |
| 2560, 40, 64, 16           | fwd     | 12.63       | 12.57       | 9.19          | 9.25          | 1.37×        | 1.36×        |
|                            | fwd+bwd | 42.53       | 31.82       | 32.08         | 23.62         | 1.33×        | 1.35×        |
| 2560, 20, 128, 16          | fwd     | 19.30       | 19.33       | 9.54          | 9.93          | 2.02×        | 1.95×        |
|                            | fwd+bwd | 56.34       | 39.07       | 38.55         | 26.22         | 1.46×        | 1.49×        |
| 4096, 64, 64, 16           | fwd     | 18.75       | 18.69       | 13.92         | 13.87         | 1.35×        | 1.35×        |
|                            | fwd+bwd | 64.11       | 48.07       | 48.82         | 35.95         | 1.31×        | 1.34×        |
| 4096, 32, 128, 16          | fwd     | 30.24       | 29.82       | 15.49         | 15.59         | 1.95×        | 1.91×        |
|                            | fwd+bwd | 87.02       | 60.32       | 59.58         | 40.07         | 1.46×        | 1.51×        |

Varlen, cs32, 8192 tokens total (`[1, 8192]` layout; N = sequence count):

| Workload (hidden, H, D, N) | Mode    | Triton dr=0 | Triton dr=1 | TileLang dr=0 | TileLang dr=1 | Speedup dr=0 | Speedup dr=1 |
| -------------------------- | ------- | ----------- | ----------- | ------------- | ------------- | ------------ | ------------ |
| 2560, 40, 64, 3            | fwd     | 1.56        | 1.55        | 1.21          | 1.20          | 1.29×        | 1.29×        |
|                            | fwd+bwd | 5.34        | 4.01        | 4.12          | 2.96          | 1.30×        | 1.35×        |
| 2560, 20, 128, 5           | fwd     | 2.37        | 2.37        | 1.27          | 1.27          | 1.87×        | 1.87×        |
|                            | fwd+bwd | 6.95        | 4.83        | 5.67          | 3.87          | 1.23×        | 1.25×        |
| 4096, 64, 64, 2            | fwd     | 2.38        | 2.39        | 1.85          | 1.83          | 1.29×        | 1.31×        |
|                            | fwd+bwd | 8.23        | 6.22        | 6.38          | 4.54          | 1.29×        | 1.37×        |
| 4096, 32, 128, 3           | fwd     | 3.69        | 3.69        | 1.98          | 1.98          | 1.87×        | 1.86×        |
|                            | fwd+bwd | 10.79       | 7.49        | 8.97          | 6.02          | 1.20×        | 1.24×        |
| 1024, 16, 64, 6            | fwd     | 0.57        | 0.57        | 0.54          | 0.54          | 1.07×        | 1.05×        |
|                            | fwd+bwd | 1.96        | 1.45        | 1.82          | 1.31          | 1.08×        | 1.11×        |
| 1024, 8, 128, 6 †          | fwd     | 0.91        | 0.92        | 0.91          | 0.92          | 1.00×        | 1.00×        |
|                            | fwd+bwd | 2.65        | 1.82        | 2.64          | 1.82          | 1.00×        | 1.00×        |

### H800 (sm_90)

Rectangular, cs32, T=4096:

| Workload (hidden, H, D, B) | Mode    | Triton dr=0 | Triton dr=1 | TileLang dr=0 | TileLang dr=1 | Speedup dr=0 | Speedup dr=1 |
| -------------------------- | ------- | ----------- | ----------- | ------------- | ------------- | ------------ | ------------ |
| 1024, 16, 64, 8            | fwd     | 1.91        | 1.89        | 1.47          | 1.47          | 1.29×        | 1.29×        |
|                            | fwd+bwd | 5.84        | 4.11        | 5.27          | 4.01          | 1.11×        | 1.02×        |
| 1024, 8, 128, 8            | fwd     | 3.83        | 3.82        | 1.78          | 1.77          | 2.15×        | 2.16×        |
|                            | fwd+bwd | 9.78        | 6.13        | 6.47          | 4.98          | 1.51×        | 1.23×        |
| 2560, 40, 64, 8            | fwd     | 4.24        | 4.26        | 3.52          | 3.52          | 1.21×        | 1.21×        |
|                            | fwd+bwd | 13.65       | 9.84        | 13.01         | 10.23         | 1.05×        | 0.96×        |
| 2560, 20, 128, 8           | fwd     | 9.13        | 9.13        | 4.80          | 4.81          | 1.90×        | 1.90×        |
|                            | fwd+bwd | 23.84       | 15.24       | 16.67         | 12.62         | 1.43×        | 1.21×        |
| 4096, 64, 64, 8            | fwd     | 6.93        | 6.94        | 5.03          | 4.97          | 1.38×        | 1.40×        |
|                            | fwd+bwd | 21.74       | 15.47       | 19.02         | 15.10         | 1.14×        | 1.02×        |
| 4096, 32, 128, 8           | fwd     | 15.14       | 15.15       | 6.90          | 6.90          | 2.19×        | 2.20×        |
|                            | fwd+bwd | 38.79       | 24.43       | 22.80         | 17.29         | 1.70×        | 1.41×        |
| 2560, 40, 64, 16           | fwd     | 8.43        | 8.43        | 6.37          | 6.37          | 1.32×        | 1.32×        |
|                            | fwd+bwd | 26.93       | 19.34       | 23.95         | 18.96         | 1.12×        | 1.02×        |
| 2560, 20, 128, 16          | fwd     | 17.84       | 17.84       | 8.62          | 8.66          | 2.07×        | 2.06×        |
|                            | fwd+bwd | 46.56       | 29.77       | 30.27         | 23.26         | 1.54×        | 1.28×        |
| 4096, 64, 64, 16           | fwd     | 13.46       | 13.46       | 9.87          | 9.87          | 1.36×        | 1.36×        |
|                            | fwd+bwd | 42.66       | 30.49       | 37.72         | 30.09         | 1.13×        | 1.01×        |
| 4096, 32, 128, 16          | fwd     | 30.24       | 30.23       | 13.48         | 13.48         | 2.24×        | 2.24×        |
|                            | fwd+bwd | 77.47       | 48.75       | 45.27         | 34.21         | 1.71×        | 1.42×        |

Varlen, cs32, 8192 tokens total:

| Workload (hidden, H, D, N) | Mode    | Triton dr=0 | Triton dr=1 | TileLang dr=0 | TileLang dr=1 | Speedup dr=0 | Speedup dr=1 |
| -------------------------- | ------- | ----------- | ----------- | ------------- | ------------- | ------------ | ------------ |
| 2560, 40, 64, 3            | fwd     | 1.31        | 1.32        | 1.11          | 1.10          | 1.18×        | 1.20×        |
|                            | fwd+bwd | 4.02        | 2.84        | 3.87          | 2.94          | 1.04×        | 0.96×        |
| 2560, 20, 128, 5           | fwd     | 8.00        | 8.00        | 1.34          | 1.34          | 5.95×        | 5.96×        |
|                            | fwd+bwd | 17.48       | 9.61        | 4.63          | 3.39          | 3.77×        | 2.84×        |
| 4096, 64, 64, 2            | fwd     | 2.07        | 2.06        | 1.69          | 1.68          | 1.22×        | 1.23×        |
|                            | fwd+bwd | 6.35        | 4.47        | 5.92          | 4.50          | 1.07×        | 0.99×        |
| 4096, 32, 128, 3           | fwd     | 12.22       | 12.22       | 2.18          | 2.13          | 5.60×        | 5.74×        |
|                            | fwd+bwd | 26.68       | 14.66       | 7.21          | 5.20          | 3.70×        | 2.82×        |
| 1024, 16, 64, 6            | fwd     | 0.62        | 0.61        | 0.57          | 0.56          | 1.08×        | 1.08×        |
|                            | fwd+bwd | 1.84        | 1.29        | 1.91          | 1.42          | 0.96×        | 0.91×        |
| 1024, 8, 128, 6            | fwd     | 3.19        | 3.19        | 0.72          | 0.71          | 4.45×        | 4.48×        |
|                            | fwd+bwd | 7.03        | 3.89        | 2.55          | 1.87          | 2.76×        | 2.08×        |

Reading the numbers:

- fwd+bwd is the weighted mix of the two passes, so it always lands
  below the fwd-only ratio: the forward carries the large wins, the
  backward (bwd-only, derived as fb − fwd) is 1.0–1.5× at dr=0 on both
  devices. At dr=1 on sm_90 the TileLang backward alone loses to
  Triton's dr=1 backward (0.6–0.9× — Triton skips its recompute too),
  and the fwd win carries the total; dr=0 is the default and the
  recommended mode.
- The H800 varlen D128 rows are genuine routes: profiling shows Triton's
  intra tensorcore kernel degrading on ragged varlen on sm_90 (14.4 of
  17.7 ms over fwd+bwd at 2560/D128/N5 — two calls, ~7.2 ms each). On
  sm_120 the same kernel is healthy (2.7 of 7.5 ms) and the fused
  TileLang A-stage still wins ~1.9× fwd there.
- Honest weak cells: small varlen D64 at dr=1 is 0.91–0.99× on sm_90
  (1.11–1.37× on sm_120); varlen cs16 measured 0.83–1.02× on sm_90 (no
  win at any size) and is now rerouted to Triton below cc12 — cc12x
  keeps it (1.08–1.23× at h2560/h4096), with the smallest cell (h1024,
  N6) at 0.92–0.98× there; one rect D64 dr=1 cell (h2560, B8) is 0.96×
  on sm_90. Grids below half the SMs reroute to Triton at ~1.00×.
- An earlier small-model/varlen collapse (0.16–0.7×) was host-side
  launch overhead in the tvm_ffi adapter (dynamic output-shape
  resolution per launch), fixed by caller-allocated outputs — not a
  kernel effect.
- Outlier note: the 7.08× PRO 6000 varlen cell from the previous table
  does not reproduce on an idle GPU — it was measured while two sweep
  shards shared one device, which inflated the Triton baseline; all
  tables above were re-run one shard per GPU. The current maximum ratio
  (5.96×, H800 varlen D128) is profiler-attributed above.

### chunk_size=64 (PRO 6000 only)

Two footprint fixes (staging-pair aliasing in the A-backward,
global-`gk` reads in the A-forward) made every BT=64 stage launchable in
99KB, bit-exact against their predecessors. The routed sweep:

| Workload                    | dr=0 fwd | dr=0 fwd+bwd | dr=1 fwd | dr=1 fwd+bwd |
| --------------------------- | -------- | ------------ | -------- | ------------ |
| rect h2560 D64 B8           | 2.11×    | 2.03×        | 2.10×    | 1.97×        |
| rect h4096 D64 B8           | 2.22×    | 2.12×        | 2.23×    | 2.05×        |
| rect h4096 D64 B16          | 2.21×    | 2.07×        | 2.22×    | 1.99×        |
| varlen h2560 D64 N3         | 1.94×    | 1.83×        | 1.93×    | 1.82×        |
| varlen h4096 D64 N2         | 2.10×    | 1.92×        | 2.10×    | 1.93×        |
| rect/varlen D128 (fallback) | 1.00×    | 1.00×        | 1.00×    | 1.00×        |

(h1024 D64 spot cells: 1.93–2.20× rect, 1.66–1.92× varlen.) On H800
every cs64 schedule launches but measured 0.68–1.05× vs. Triton, so the
verifier accepts cs64 only on cc12x+; below that it falls back at 1.00×.

Why cs64 loses on sm_90 (per-kernel attribution via `torch.profiler`
on both GPUs at the same rect h4096/h1024 D64/D128 dr1 configs,
generated code inspected from the TileLang/Triton kernel caches):

- The deficit is confined to the two serial state-scan stages — the
  fused h+o forward and the stream backward — which walk T/64 chunks
  serially per block on grids of N·H·(V/BV) and N·H blocks. At h1024
  D128 that is 256/64 blocks against 132 SMs, so the scans run long
  underfilled chains while Triton's split passes (dhu + dqkwg + o)
  re-parallelize over chunks and V-blocks (stream 1.13 ms vs. 0.65 ms
  combined; h+o 0.74 ms vs. 0.53 ms — together most of the 0.92×
  total, while the fused intra/A stage itself wins 0.61 ms vs.
  0.92 ms). Once the grid fills, the scan stages reach parity
  (stream 2.50 ms vs. 2.64 ms at h4096 D64 dr1) and the residual
  delta is the intra/A stage (2.96 ms vs. 2.24 ms), where Triton's
  sm_90a schedule is at its strongest.
- IKET warp-level tracing of the stream kernel (TileLang's in-kernel
  event tool, `nvidia-cutlass-dsl`) confirms the mechanism at 64
  blocks / 8 warps on 132 SMs (h1024 D64, cs64): per chunk iteration
  (~8 µs of dependent phases), global staging dominates — input
  loads 33% + output row stores 21% = 54% of warp time vs. 37% for
  all mma phases (dv2 11% + state gemms 23% + dh update 4%), with
  dgk reduction at 8%. The next chunk's loads are not prefetched
  ahead of the dh dependency chain, and 1 block/SM leaves ~12% warp
  occupancy to hide load latency. That points at next-chunk
  prefetch/pipelining as the follow-up optimization, on all
  architectures.
- Triton's sm_90a intra kernels are an exceptionally tuned baseline:
  the same stage takes 2.2 ms on H800 vs. 12.3 ms on the PRO 6000 —
  a 5.5× regression of the *Triton reference itself* on sm_120 (2.4×
  for the backward intra). Our kernels' absolute times are flat across
  architectures (≤1.3×, tracking the 1980 vs. 3090 MHz clocks). Both
  backends emit the same instruction class on both GPUs (mma.sync
  m16n8k16 + ldmatrix; no wgmma/TMA on either side), so the gap is
  scheduling, not tensor-core ISA.
- The intra stages are latency-bound everywhere (~3% of peak MMA
  FLOPS, ~15–18% of peak bandwidth), which is why fusion pays once
  the GPU is filled, and why sm_120 — where the reference's intra
  path collapses — flips the same h4096 config from 0.89× to 1.88×.
- The same serial structure is not a liability on cc12x: stage-for-
  stage the scans hold up there (stream 3.21 ms vs. Triton's combined
  dh-side passes 5.01 ms = 1.56×; fused h+o at parity, 2.42 vs.
  2.54 ms) — Triton's split passes are just as latency-bound and
  gain nothing from the extra 56 SMs. What flips the total is the
  intra/A stage turning a 0.76× deficit into a 3.3× win (3.72 ms
  vs. 12.30 ms): the stages the design is weak in stay at or above
  parity, and the stage it is strong in decides the total
  (24.85 ms vs. 46.69 ms = 1.88×).

Correctness gates: route-parity vs. Triton within `assert_close`
0.008–0.01 on all outputs and gradients (rect + varlen, fwd + bwd,
fp16/bf16, fp32 gates, K=V∈{64,128}, chunk 16/32/64); parity vs. the
per-token fp32 recurrence at the same 0.007/0.008 ratios
`test_dplr_delta` holds the Triton path to (worst observed 0.0051);
exact 0.0000 match for bf16 + fp32 gates on rectangular inputs; the
2-GPU CP test against the naive recurrence. The full gate
(`tests/ops/test_dplr_tilelang.py` + `tests/ops/test_dplr_delta.py`,
99 tests) passes on both devices: 67 passed + 2 skipped in the backend
file, and `test_dplr_delta.py` apart from the one `D100-chunk_size64`
fp16 failure that reproduces identically on `main` (pre-existing; head
dim 100 never routes to TileLang). With cs64 gated to cc12x+, the sm_90
run skips the cs64 parity rows and exercises the rejection unit test
instead.

## 3. Current incompatibilities

The verifier rejects these and the call falls back to Triton (one log
line per reason):

| Configuration | Reason |
| ------------- | ------ |
| `safe_gate=False` without a fitting `lower_bound` | The mid-chunk-centered A-stage needs a caller-asserted gate bound to keep `exp2` operands in fp32 range; unbounded calls stay on Triton. RWKV7's clamp (w ∈ (−0.61, 0)) qualifies as either. |
| fp32 activations | No launchable fp32 stream-backward schedule on ≤99KB smem; fp32 is also second-class in the Triton path. |
| `gk` dtype ∉ {fp16, bf16, fp32}; `k/v/a/b` dtype ≠ `q` dtype | Gate math needs a real dtype; intra kernels load at one dtype boundary. |
| `K != V` | The streaming backward targets square states. |
| head dim ∉ {64, 128}; `chunk_size` ∉ {16, 32, 64} | Not validated. |
| `chunk_size` 16 + head dim 128; varlen `chunk_size` 16 below cc12 | ~0.5× measured (non-vectorized A-stage, 2-warp h+o at BT=16); varlen cs16 measures 0.83–1.02× on sm_90 — cc12x keeps it (1.08–1.23× at h2560/h4096). |
| `chunk_size` 64 | Two gates: launchability (K=128 needs ≥167936B smem — rejected on 99KB caps and A100; K=64 fits via 98432B/82944B schedules) and measurement (0.68–1.05× on sm_90, so accepted only on cc12x+, where K=64 is 1.66–2.23×). |
| Small grids (`N·H·(V/BV) < SMs/2`) | The fused serial state pass underfills the GPU (up to 10× slower at B2 H8). |
| Small varlen workloads | Fixed (host launch overhead; PRO 6000 low-smem stream-bwd vectorization). Now 1.05–1.87× on sm_120, 1.08–5.96× fwd / 0.91–3.77× fwd+bwd on sm_90; residual dr=1 D64 cells 0.91–0.99× on sm_90 only. |
| Non-CUDA devices | TileLang kernels are CUDA-only. |
| GQA / MQA | Shared-head ABI (same constraint as the Triton DPLR). |

`cp_context` is **supported** (same boundary-state exchange and the same
contract as the Triton entry: no user `initial_state` /
`output_final_state` under CP, loud failure on a user `dht`).

One stress-test note: under sustained saturation at the −5 gate bound
(every gate pinned, cs32) the Triton reference's `dgk` diverges from the
true recurrence far beyond parity tolerance (measured err ratio ~0.5 vs.
TileLang's ~0.17 on sm_120; both stay finite), so stress rows pin ~1% of
gates. The `lower_bound` kwarg is wired through both routes.

## Test plan

- `tests/ops/test_dplr_tilelang.py` (new): verifier accept/reject per
  branch (incl. `safe_gate=False`, `lower_bound` licensing, cs64 device
  gates, small-grid guard), route parity vs. Triton (fwd + all grads +
  `dh0`, rect + varlen, bf16/fp16, fp32 gates, chunk 16/32/64, both dr
  policies), gate-stress rows, PyTorch-recurrence baseline rows, and
  `opcheck` / `torch.compile` fullgraph / `torch.utils.checkpoint`
  smokes.
- `tests/ops/test_backends.py`: nvcc/tilelang gating parametrization
  extended to the new backend.
- `tests/context_parallel/test_cp_dplr.py`: new 2-GPU
  `test_cp2_tilelang_route` with per-rank route assertion.
- Regression: `test_dplr_delta.py`, `test_rwkv7.py`,
  `test_modeling_rwkv7.py`, `test_layer_cache_layer_idx.py` pass (the 4
  `channel_mixing_gradients` and 1 `D100-chunk_size64` failures
  reproduce identically on `main` — pre-existing).

## Breaking changes

None. The change is additive: where the verifier does not accept a call,
the default Triton implementation is used unchanged.

## Checklist

---

*Developed on `tilelang-dplr-backend`; benchmarks on idle RTX PRO 6000
(sm_120) and H800 (sm_90). Correctness gates pass on both.*


fla/ops/generalized_delta_rule/dplr/backends/tilelang/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        v: torch.Tensor,
+        a: torch.Tensor,
+        b: torch.Tensor,
+        gk: torch.Tensor,
+        scale: float | None = None,
+        initial_state: torch.Tensor | None = None,
+        output_final_state: bool = False,
+        cu_seqlens: torch.LongTensor | None = None,
+        cu_seqlens_cpu: torch.LongTensor | None = None,
+        safe_gate: bool = False,
```
zhiyuan1i, line 57: `safe_gate` is accepted here but never checked. Please route **only** when `safe_gate=True and chunk_size == 64`: the kernels' mid-chunk-centered tensorcore scheme overflows fp32 when `(chunk_size/2) * max|gk|` ≳ 127 log2. `safe_gate=False` callers make no gate-range assertion — Triton's `sub_intra` stays correct for arbitrary gates, so those calls must not be rerouted. And `safe_gate=True` alone is not sufficient for BT=64: the documented [-5, 0) range gives a 160-log2 half-range at BT=64 (it only licenses BT≤32). RWKV7 satisfies both conditions (`w` architecturally clamped to (-0.61, 0), explicit `chunk_size=64`).

fla/ops/generalized_delta_rule/dplr/backends/tilelang/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+            return False, f"TileLang backend supports head dim 64 or 128 (got {k.shape[-1]}); fall back to Triton"
+        chunk_size = 16 if chunk_size is None else chunk_size
+        if chunk_size == 16 and k.shape[-1] == 128:
+            # measured ~0.5x vs Triton (the non-vectorized A-stage and the
+            # 2-warp h+o path at BT=16 do not pay off at K=128)
+            return False, "TileLang backend is slower than Triton at chunk_size 16 with head dim 128; fall back to Triton"
+        if chunk_size == 64:
+            # the intra backward at BT=64 needs 131200B (K=64) or 148480B
+            # (K=128) of shared memory per block; smaller caps cannot run it
+            smem_need = 131200 if k.shape[-1] <= 64 else 148480
```
zhiyuan1i: This gate is derived from the intra-bwd requirement, but the stream-bwd schedules need more: at (K=128, BT=64) `high`=295424B and `low`=167936B, both above A100's 166912B optin — so on A100 the call passes this check and then `_select_stream_bwd_schedule` raises at backward time instead of falling back. A100 is a target platform for us; please derive the gate from the schedule's own arithmetic (or reject the combo here).

fla/ops/generalized_delta_rule/dplr/backends/tilelang/chunk.py
```diff
@@ -0,0 +1,1324 @@
...
+    if cp_context is not None:
+        assert initial_state is None, "Initial state is not supported for CP"
+        assert output_final_state is False, "Output final state is not supported for CP"
+        assert cp_context.cu_seqlens is not None, "cu_seqlens is required for CP"
+        cu_seqlens = cp_context.cu_seqlens
+    chunk_size = 16 if chunk_size is None else int(chunk_size)
+    scale_f = float(q.shape[-1] ** -0.5 if scale is None else scale)
+    n_seqs = len(cu_seqlens) - 1 if cu_seqlens is not None else q.shape[0]
+    if initial_state is not None:
+        h0 = initial_state
```
zhiyuan1i: `initial_state` needs a contiguity guard: a non-contiguous fp32 `h0` would be read with wrong strides by the kernels (the Triton path is protected by `input_guard`, which this backend bypasses). Please normalize here or check contiguity in the verifier.

fla/ops/generalized_delta_rule/dplr/backends/tilelang/chunk_ho_fwd.py
```diff
@@ -0,0 +1,769 @@
...
+
+
+def _ho_fwd_configs(
+    H, K, V, BT,
+    in_dtype, state_dtype, USE_INITIAL_STATE, STORE_FINAL_STATE,
+    BV: int = 64, threads: int = 128,
+):
+    if not torch.cuda.is_available():
+        return [{"BV": 16, "threads": 64}]
+    major = torch.cuda.get_device_capability()[0]
```
zhiyuan1i: This queries the *current* device rather than `q.device` — on multi-GPU with a different current device the configs come from the wrong GPU while the Triton path (via `input_guard`'s device ctx) works. Same pattern in `_ho_fragment_merge_flags` and `_chunk_h_fwd_configs`. Also consider moving these raw `torch.cuda` queries into `fla.utils` helpers, per repo guidelines.

fla/ops/generalized_delta_rule/dplr/backends/tilelang/chunk_A_bwd.py
```diff
@@ -0,0 +1,446 @@
...
+    dbg_f = dbg.reshape(N_tokens, H, K).contiguous()
+    dgk_last_f = dgk_last.reshape(n_chunks, H, K).contiguous()
+    dq_f = dq_out.reshape(N_tokens, H, K).contiguous()
+    dk_f = dk_out.reshape(N_tokens, H, K).contiguous()
+    da_f = da_out.reshape(N_tokens, H, K).contiguous()
+    db_f = db_out.reshape(N_tokens, H, K).contiguous()
+    dgk_out_f = dgk_out.reshape(N_tokens, H, K).contiguous()
+    # Temporary ABI shim: the fused kernel specialization ignores dAqk/dAqb,
+    # but the shared JIT signature still requires same-shape bf16 inputs.
+    dummy_dA = q.new_empty((N_tokens, H, BT))
```
zhiyuan1i: Nit: two full-size `[N_tokens, H, BT]` allocations that the FUSE_QSIDE_DA kernel never reads (~128MB dead memory at T=8192/H=64/BT=64). Consider 0-elem tensors or splitting the kernel signature.

fla/ops/generalized_delta_rule/dplr/backends/tilelang/wy_fast_bwd.py
```diff
@@ -0,0 +1,322 @@
...
+                if t < eos:
+                    if r > c:
+                        dA_ab[t, i_h, c] = T.Cast(in_dtype, dA_ab_frag[r, c])
+                    else:
+                        dA_ab[t, i_h, c] = T.Cast(in_dtype, 0.0)
+
+    return wy_fast_bwd_tl
+
+
+def chunk_dplr_bwd_wy(
```
zhiyuan1i: Nit: dead code — only the `_into` variant is used.

fla/ops/generalized_delta_rule/dplr/backends/tilelang/wy_fast_bwd.py
```diff
@@ -0,0 +1,322 @@
...
+    in_dtype = str(dw.dtype).split(".")[-1]
+
+    A_ab_inv_f = A_ab_inv.reshape(N_tokens, H, BT).contiguous()
+    A_ak_f = A_ak.reshape(N_tokens, H, BT).contiguous()
+    ag_f = ag.reshape(N_tokens, H, K).contiguous()
+    v_f = v.reshape(N_tokens, H, V).contiguous()
+    dw_f = dw.reshape(N_tokens, H, K).contiguous()
+    du_f = du.reshape(N_tokens, H, V).contiguous()
+    dv0_f = dv0.reshape(N_tokens, H, V).contiguous()
+    dA_ab_f = dA_ab_out.reshape(N_tokens, H, BT).contiguous()
```
zhiyuan1i, line 305: Nit: `reshape(...).contiguous()` on an out-param silently writes into a temporary copy if a caller ever passes a non-contiguous tensor — the kernel's writes would be dropped. Please `assert out.is_contiguous()` here and in the other `_into` wrappers.

tests/context_parallel/test_cp_dplr.py
```diff
@@ -441,6 +464,26 @@ def test_cp2_safe_gate():
     )
 
 
+def test_cp2_tilelang_route():
```
zhiyuan1i, line 468: Nit: this should skip when `DPLRTileLangBackend.is_available()` is False — on a 2-GPU machine without tilelang/nvcc it fails instead of skipping (the `requires_tilelang_route` pattern in `test_dplr_tilelang.py` covers this). Also please restore `FLA_TILELANG` to its previous value rather than `del`.

fla/ops/generalized_delta_rule/dplr/backends/tilelang/chunk.py
```diff
@@ -0,0 +1,1324 @@
...
+        scale_f,
+        initial_state is not None or cp_context is not None,
+        output_final_state,
+        cu_seqlens is not None,
+        chunk_size,
+    )
+    token = _DPLR_CP_CONTEXT.set(cp_context)
+    try:
+        if disable_recompute:
+            if not is_compiling and checkpoint_phase == _CHECKPOINT_PHASE_FORWARD or (
```
zhiyuan1i: Nit: please parenthesize — `A and B or (C and D and E)` happens to parse correctly here but is fragile.

zhiyuan1i (changes requested):
Thanks — impressive work. We checked the generated CUDA and the headline claims hold: the fused fwd h+o single sweep (no h/v_new global round-trip) and the never-materialized bwd dh are real, and the stage mapping onto the Triton pipeline is complete. A few items to address before merge:

1. **Verifier routing must require `safe_gate=True and chunk_size == 64`** (inline). The TileLang A-stage always uses the mid-chunk-centered tensorcore scheme, which is only valid while `(chunk_size/2) * max|gk|` stays in fp32 range. Default `safe_gate=False` callers assert nothing about gate range (Triton's `sub_intra` is correct for arbitrary gates — those calls must stay on Triton), and the documented safe range [-5, 0) only covers BT≤32: at BT=64 the centered half-range is 160 log2 and overflows fp32. RWKV7 satisfies both conditions (bounded `w`, explicit `chunk_size=64`). Longer term, consider GDN2's explicit `lower_bound` contract.
2. **Please add route-parity tests for `initial_state=None` + `output_final_state=False`** — every parity case currently passes h0 and requires the final state, but RWKV7 training hits exactly those uncovered branches. A checkpoint-elision test and a torch.compile/opcheck smoke test would also be worthwhile, since both are headline features.
3. **Robustness** (details inline): `initial_state` needs a contiguity guard; exceptions from the impl (JIT failure, schedule selection) escape instead of falling back, contrary to the PR description; and the (K=128, chunk64) smem gate does not match the stream-bwd requirement — on A100 this passes the verifier and then raises. A100 is a target platform for us.
4. **Benchmark evidence**: per the repo's MR-readiness conventions, kernel PRs need before/after numbers on the same hardware, dense + varlen. If Hopper access is the blocker, we can run them on our H100 — happy to help.
5. **Style**: please move the raw `torch.cuda.get_device_capability/properties` queries into small `fla.utils` helpers, per the repo guidelines.

sablin39:
Thanks for the thorough review — all five items are now addressed in the pushed commits. Details per item:

**1. Gate-bound routing.** The verifier now requires a caller-asserted gate bound, exactly as you framed it. `safe_gate=False` calls with no bound stay on Triton unconditionally. For cs64 we took your longer-term suggestion rather than the `chunk_size == 64` carve-out: a new `lower_bound` kwarg (GDN2's convention) is wired through `chunk_dplr_delta_rule`/`chunk_rwkv7` — it canonicalizes to `safe_gate=True` on the Triton path (so both routes validate the bound identically, including `ValueError` on non-negative bounds), and the TileLang verifier accepts when `|lower_bound|·(chunk_size/2)·log2e ≤ 120` (`gate_bound_is_safe` in `fla/ops/generalized_delta_rule/dplr/chunk.py`). RWKV7's architectural clamp (w ∈ (−0.61, 0) → 28 log2 at BT=64) is what licenses cs64; the [-5, 0) safe_gate range licenses cs16/32. The flag/bound is a pure contract — kernels run the same centered scheme either way, so there is no second code path to validate. Tests cover accept/reject per branch, `ValueError` parity across routes, and a `lower_bound` route-parity row asserting the TileLang route is actually taken.

**2. Coverage of the RWKV7-training branches.** `tests/ops/test_dplr_tilelang.py` now has route-parity rows (fwd + all six input grads + `dh0`) with `initial_state=None` and `output_final_state=False`, in both rect and varlen layouts. Also added, since both are headline features: `torch.library.opcheck` on the forward custom op, a `torch.compile` full-graph smoke on the backend entry, and a `torch.utils.checkpoint` test that asserts the context-function elision route is taken. PyTorch-baseline rows compare the routed op against the per-token fp32 recurrence at the same 0.007/0.008 ratios `test_dplr_delta` holds the Triton path to, including gate-stress rows at the safe_gate envelope.

**3. Robustness.** (a) `initial_state` is `.contiguous()`-guarded in the backend entry. (b) The TileLang route is wrapped so that any exception from the impl (JIT failure, schedule selection, launch failure) falls back to Triton with a logged warning — behavior now matches the PR description. (c) The smem gate is fixed properly rather than patched: the verifier and the launcher now share one arithmetic module (`schedules.py`) that computes every stage's shared-memory footprint from (K, BT, dtype, device optin), so acceptance implies every stage launches on the current device. A100's 166912B optin rejects K=128 cs64 (all three stream-backward schedules exceed it) and accepts K=64; the test suite includes A100-shaped cases via monkeypatched device properties. We'd appreciate an A100 spot-check from your side if convenient — our hardware coverage is PRO 6000 (sm_120) and H800 (sm_90).

**4. Benchmarks.** Same-hardware before/after, dense + varlen, now in the PR body: PRO 6000 (sm_120) and H800 (sm_90), bf16, routed `FLA_TILELANG=0` vs. `=1`, both `disable_recompute` policies. Headline: all 64 H800 cells ≥1.02×; on the PRO 6000 every accepted cs32 cell is ≥0.96× (small-grid cells reroute to Triton at ~1.00× by design) and cs64-D64 is 1.6–2.2×. Two anomalies are footnoted with profiling evidence (one genuine 7.1× varlen win, one rerouted fallback measuring do_bench noise). If you can rerun the sweep on your H100 we'd gladly take the cross-check — the H800 numbers suggest nothing is architecture-specific to sm_120.

**5. Style.** Device queries now go through small `fla.utils` helpers (commit `0e4f1e19`); the raw `torch.cuda.get_device_capability/properties` calls are gone from the backend.

Inline nits (dummy 1-elem alloc in `chunk_A_bwd`, `_into`-only wy_fast_bwd, and/or parenthesization, `FLA_TILELANG` save/restore in the CP test) are all fixed in place.

Rebased onto current `main` (includes your DPLR safe-gate midpoint fix — our centering math was unaffected, and it's nice to see the same midpoint analysis land upstream).

fla/ops/generalized_delta_rule/dplr/backends/tilelang/__init__.py
```diff
@@ -0,0 +1,239 @@
...
+            return False, (
+                f"TileLang backend requires K == V (got K={k.shape[-1]}, V={v.shape[-1]}); "
+                "fall back to Triton"
+            )
+        if k.shape[-1] not in (64, 128):
+            return False, f"TileLang backend supports head dim 64 or 128 (got {k.shape[-1]}); fall back to Triton"
+        chunk_size = 16 if chunk_size is None else chunk_size
+        if chunk_size not in (16, 32, 64):
+            return False, f"TileLang backend supports chunk_size 16/32/64, got {chunk_size}; fall back to Triton"
+        if not safe_gate:
```
zhiyuan1i: Blocking: when `safe_gate=True`, the gate-bound check is skipped entirely, so `safe_gate=True, chunk_size=64` enters the mid-chunk-centered A-stage unconditionally. The public docstring still describes safe_gate gates as roughly [-5, 0), which at cs64 reaches 5×32×log2e ≈ 231 log2 — well past the fp32 exp2 range (~127). safe_gate=True callers assert *a* bound, but nothing here enforces that the asserted bound fits the chunk size. The verifier should apply `gate_bound_is_safe` to both paths: for safe_gate=True use the documented default bound (-5), and reject cs64 for it, same as the lower_bound path.

fla/ops/generalized_delta_rule/dplr/chunk.py
```diff
@@ -452,6 +470,7 @@ def chunk_dplr_delta_rule(
     cu_seqlens: torch.LongTensor | None = None,
     cu_seqlens_cpu: torch.LongTensor | None = None,
     safe_gate: bool = False,
+    lower_bound: float | None = None,
```
zhiyuan1i: Blocking: inserting `lower_bound` between `safe_gate` and `chunk_size` breaks every existing positional caller — a positional `chunk_size` now binds into `lower_bound` (a float), silently changing behavior instead of raising. Please append new parameters after the existing public ones (or make it keyword-only), per the GDN2 convention.

zhiyuan1i (changes requested):
Thanks for the thorough follow-up — the A100 smem gate now correctly rejects (K=128, cs64) and accepts (K=64, cs64), the device queries are properly routed through `fla.utils` helpers, and the new parity/opcheck/compile/checkpoint tests are real additions. Four items still block merge:

1. **Numerical-safety hole (inline):** the verifier skips the gate-bound check entirely when `safe_gate=True`, so `safe_gate=True, chunk_size=64` — a combination the public docstring still advertises as ~[-5, 0) — enters the centered TileLang A-stage with worst-case exponents of ~231–238 log2, past fp32 range. The new tests even assert this unsafe combination is *accepted*. The RWKV7 built-in caller is safe (clamped w → ~28 log2), but the public path is not.

2. **Positional-argument break (inline):** `lower_bound` is inserted between `safe_gate` and `chunk_size` in the public signature, so existing positional callers silently bind their chunk_size into lower_bound. Please append it after the existing parameters (keyword-style, per the GDN2 convention).

3. **CP rank-local fallback:** the runtime TileLang→Triton fallback is decided per rank. Under context parallelism, one rank failing over while its peers stay on TileLang can diverge the collective sequence or hang. The fallback decision needs to be collective-uniform (e.g. all-reduce the verifier result), or disabled under CP.

4. **H800 cs64 regression still routed:** your own table shows H800 cs64 fwd+bwd at 0.54–0.81x, yet the verifier accepts cs64 on sm90 — contradicting the "only accepts when measurably faster" claim. Please either reject cs64 on sm90 (fall back to Triton) or provide per-shape data justifying the accepted subset; the Impact section's "cs64 up to ~2x on both architectures" also needs correcting against your H800 numbers.

Two smaller items:

- `gate_bound_is_safe` is off by one: it uses `chunk_size/2` rows, but `centered_ge` spans `chunk_size/2 + 1` rows. At cs16 the helper accepts lower_bound≈-10.4, where the real exponent reaches ~135 log2 and overflows. Please tighten the formula.
- The varlen benchmark numbers aren't reproducible: only total tokens and sequence count N are given — no cu_seqlens, generation rule, seed, or script. Please add the harness details (and the 7.08x PRO 6000 outlier deserves a sanity note).

Once the verifier gates both safe_gate paths on the (corrected) bound formula, the signature keeps positional compatibility, the CP fallback is made uniform, and sm90 cs64 stops routing to a slower backend, this should be in good shape.

sablin39:
Thanks for the careful review. Items 1–3 and the two minors are addressed in d24c7e9a, fd7ee74c, 43beb652 (pushed); item 4 needs a Hopper box, plan below.

**1. safe_gate + cs64 overflow** — confirmed and fixed. The verifier now maps `safe_gate=True` to its documented [-5, 0) bound and runs `gate_bound_is_safe` on both gate paths, so `safe_gate=True` + cs64 falls back to Triton (the cs64 half-range at -5 is 231 log2, past the 128 fp32 limit). RWKV7 keeps its cs64 TileLang route by asserting its architectural w clamp through `lower_bound=-exp(-0.5)` in `fla/layers/rwkv7.py`. New sustained-pin stress tests clamp every gate at the licensed bound for the whole sequence (cs16/32 at -5, cs64 at -2.0 and -0.61) and assert finite outputs tracking the fp32 per-token recurrence; a routing regression test asserts safe_gate+cs64 stays on Triton. One finding worth noting: at the -5 pin the Triton reference itself is the less accurate backend — `dgk` RMS err vs the fp32 recurrence is 0.34–0.48 for Triton vs 0.12–0.17 for TileLang — so those rows assert parity against the baseline rather than cross-backend.

**5. Off-by-one in `gate_bound_is_safe`** — measured the actual onset to check this; the formula is correct as-is. With all gates pinned, the NaN onset matches a cs/2-row half-range, not cs/2+1: at cs64 it lands between 2.75 and 2.80 (cs/2 predicts 128/(32·log2e) = 2.77; cs/2+1 predicts 2.69, but 2.70 is still finite); at cs16 between 11.0 and 11.5 (cs/2 predicts 11.09). The extra row does not participate in the max exponent in this kernel, and the 120 threshold keeps real margin to the 128 limit.

**2. Positional break** — `lower_bound` moved after `cp_context` in the public op and the backend protocol methods; all in-tree callers use keywords.

**3. CP fallback** — the forward now re-raises under `cp_context` instead of falling back rank-locally.

**6. Varlen repro** — generation rule added to the benchmark preamble: 8192 tokens flattened to `[1, 8192]`, N sequences from `torch.Generator(seed=17+N)`, lengths proportional to `rand(N)+0.35`, last sequence absorbing the remainder; inputs are bf16 `randn` with `a = normalize(rand)`, `b = -a`, fp32 gates `-0.61·sigmoid(5·randn)`. Note the cs64 cells now pass `lower_bound=-0.61` to license the route post-fix.

**4. H800 cs64** — agree with the data (0.54–0.81×). Plan: gate cs64 to devices with measured wins (cc12x) in the verifier, one cc check next to the existing schedule arithmetic — unless a Hopper re-measurement justifies keeping sm_90. I'll run the Hopper pass first (machine access incoming) and follow up either way.

sablin39:
Thanks for the detailed review — all four blockers and both smaller items are addressed in the current head.

**1. Gate-bound hole.** Both `safe_gate` and `lower_bound` now go through `gate_bound_is_safe`; `safe_gate=True + cs64` (33-row span, ~238 log2, past fp32 range) is rejected and falls back to Triton, with tests asserting the rejection and the Triton fallback results. cs64 is exercised via `lower_bound`; RWKV7's clamped w (~29 log2) still routes.

**2. Signature order.** `lower_bound` is appended after `cp_context`, keyword-style per GDN2 — positional callers are unaffected.

**3. CP fallback.** Under `cp_context` the forward re-raises instead of failing over per rank (`43beb65`, pushed after your review), so ranks can't diverge on the collective sequence; a new 2-GPU test asserts the route per rank.

**4. sm90 cs64.** cs64 now routes only on cc12x+ (acceptance still implies launchability via shared schedule arithmetic); varlen cs16 measured the same and got the same gate:

| Route | H800 (sm_90) | PRO 6000 (sm_120) | Now routes on |
| --- | --- | --- | --- |
| cs64, K=64 | 0.68–1.05× | 1.66–2.23× | cc12x+ only |
| varlen cs16 | 0.83–1.02× | 1.08–1.23× (h2560/h4096) | cc12x+ only |
| rect cs16 | 1.06–1.35× | 1.06–1.35× | both |

Rejected configs fall back at 1.00×; the Impact section is corrected.

Why it lags on Hopper (per-kernel attribution + IKET warp-level tracing): the deficit is confined to the two serial state-scan stages (fused h+o forward, stream backward), which walk T/64 chunks serially on grids of N·H·(V/BV) and N·H blocks — at h1024 D128 that's 256/64 blocks against 132 SMs, while Triton's split passes re-parallelize over chunks and V-blocks (stream 1.13 vs. 0.65 ms; h+o 0.74 vs. 0.53), even though our fused intra stage wins there (0.61 vs. 0.92 ms). IKET confirms it: per ~8 µs chunk chain, 54% of warp time is global staging vs. 37% for all mma phases, with no next-chunk prefetch and ~12% occupancy — prefetch/pipelining is the follow-up. Once the grid fills, the scans reach parity and the residual is the intra stage (2.96 vs. 2.24 ms), Triton's strongest.

On cc12x the same structure holds up: the stream still wins (3.21 vs. 5.01 ms) and h+o is at parity — Triton's split passes are just as latency-bound and gain nothing from the extra 56 SMs. The total flips because Triton's sm_90a-tuned intra regresses 5.5× on sm_120 (2.2 → 12.3 ms) while ours stays flat (≤1.3×, tracking clocks): 0.76× deficit → 3.3× win, deciding the 1.88× total. Same ISA class on both GPUs (mma.sync m16n8k16, no wgmma/TMA) — scheduling, not tensor cores. The cumsum/stream fixes also moved the worst rect cell from 0.80× to 0.92× on H800; what remains is the structural underfill, hence the gate.

**Off-by-one.** `gate_bound_is_safe` now uses the exact `chunk_size/2 + 1`-row span with a 124 log2 threshold (128 limit − 4 headroom); your −10.4/cs16 example (135 log2) is rejected, boundary cases unit-tested.

**Varlen reproducibility.** The harness is documented: 8192 tokens split into N sequences with lengths ∝ `rand(N) + 0.35` (seed = 17 + N), `[1, 8192]` layout, fp32 `gk = −0.61·sigmoid(5x)`, and the same do_bench script (warmup 10, rep 50, `FLA_TILELANG=0/1` through the dispatched op) drives every table. The 7.08× cell was two sweep shards sharing a GPU inflating the Triton baseline; all tables re-ran one shard per GPU, headline cells cross-checked with CUDA-event timing and profiler attribution (current max 5.96×, H800 varlen D128).

**Accuracy.** Pinned by the new tests: route parity holds TileLang to Triton's own err-ratio bars vs. the fp32 recurrence (0.007/0.008) across rect+varlen, bf16/fp16, fp32 gates, cs16/32/64, both `dr` policies — with consistently tighter margins (0.0058 vs. 0.0104 at cs32/D128/bf16). Under sustained gate pinning both stay finite, but Triton's dgk err lands ~0.5 vs. ours ~0.17 — why the stress rows pin only ~1% of gates. Rect bf16 + fp32-gates forward is bit-exact.

Suite: `test_dplr_tilelang.py` 70 passed + 2 skipped; `test_dplr_delta.py` passes (the D100-cs64 fp16 failure reproduces on `main`). Branch merged with current `main`, audits re-ran clean.

sablin39:
Follow-up round (`852bb3f4`, branch rebased onto current main for a clean merge): TileLang 0.1.13 compatibility + vectorization. Codegen and schedules only — no verifier, routing, or API changes since the last review round; no new tests needed (existing gate re-run plus the audits below).

Environment for all numbers below: tilelang 0.1.13 @ `3abbef4e` (`v0.1.13-61-g3abbef4e`), same source on both GPUs. PRO 6000 = RTX PRO 6000 Blackwell (sm_120, 188 SMs), H800 (sm_90, 132 SMs).

**How routing works (unchanged this round, restated for the tables).** The verifier accepts or rejects the *whole call* — fwd + all bwd stages run on one backend, no per-stage mixing. Rejection means the default Triton implementation runs unchanged, with one log line per reason. Gates, in order:

- *Numeric safety*: dtypes (fp16/bf16 activations, fp32-capable gates), K=V∈{64,128}, chunk∈{16,32,64}, and the gate-bound rule `(chunk_size/2+1)·|bound|·log2e ≤ 124` applied to both `safe_gate` (bound −5) and explicit `lower_bound` — this keeps the centered A-stage's exp2 operands inside fp32 range.
- *Launchability*: acceptance requires a kernel schedule that fits the device's smem opt-in cap, using arithmetic shared with the launcher — so an accepted call cannot raise "schedule not found" (the A100 cs64-K128 case).
- *Measured performance gates*: cs16+K128 rejected (~0.5×); varlen cs16 rejected below cc12 (0.83–1.02× on sm_90); cs64 rejected below cc12 (0.68–1.05× on sm_90); the fused serial state pass needs a grid of `N·H·⌈V/BV⌉ ≥ SMs/2` (BV = 64 at V=64, 32 at V=128) — it parallelizes only over (V-blocks, seqs, heads) and walks chunks serially, so below half the SMs it loses to Triton's split kernels; the low-smem (97KB) stream backward additionally needs `N·H ≥ SMs/2` since its footprint leaves no prefetch headroom.
- *Runtime*: a JIT/launch failure after acceptance still falls back to Triton — except under CP, where it re-raises so ranks can't diverge on the collective sequence.

**64k-token validation.** Same harness as the previous tables, T=65536 (rect; varlen = `[1, 65536]` split into N sequences; per-row B/N chosen to clear the SMs/2 grid guard on both GPUs). Every routed cell is bit-exact against the Triton route (worst rel. diff 0.0 over `o`, final state, all grads) — 2048-chunk serial chains at cs32 do not drift. All times in ms, fwd+bwd, dr=0/dr=1:

PRO 6000, cs32 (fwd-only ratios: 1.27–1.41× D64, 1.85–1.96× D128):

| hidden | H | D | B/N | layout | Triton | TileLang | speedup | route |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1024 | 16 | 64 | 8 | rect | 135.5/101.8 | 101.6/72.6 | 1.33/1.40 | TileLang |
| 2560 | 40 | 64 | 4 | rect | 177.3/132.5 | 130.1/94.9 | 1.36/1.40 | TileLang |
| 4096 | 64 | 64 | 2 | rect | 135.9/102.6 | 101.4/72.7 | 1.34/1.41 | TileLang |
| 1024 | 8 | 128 | 12 | rect | 277.1/201.0 | 208.2/140.1 | 1.33/1.43 | TileLang |
| 2560 | 20 | 128 | 5 | rect | 300.5/207.4 | 220.8/145.2 | 1.36/1.43 | TileLang |
| 4096 | 32 | 128 | 3 | rect | 279.7/192.7 | 211.3/138.0 | 1.32/1.40 | TileLang |
| 1024 | 16 | 64 | N8 | varlen | 17.7/13.2 | 14.0/9.9 | 1.27/1.34 | TileLang |
| 2560 | 40 | 64 | N4 | varlen | 45.4/33.9 | 34.2/24.3 | 1.33/1.39 | TileLang |
| 4096 | 64 | 64 | N2 | varlen | 70.7/52.9 | 54.5/37.8 | 1.30/1.40 | TileLang |
| 2560 | 20 | 128 | N8 | varlen | 58.1/39.9 | 43.3/29.1 | 1.34/1.37 | TileLang |
| 4096 | 32 | 128 | N4 | varlen | 90.0/62.0 | 70.1/46.7 | 1.28/1.33 | TileLang |

PRO 6000, cs64 native (fwd-only ratios: 1.88–2.16×):

| hidden | H | D | B/N | layout | Triton | TileLang | speedup | route |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1024 | 16 | 64 | 8 | rect | 247.3/174.3 | 121.5/86.9 | 2.04/2.01 | TileLang |
| 2560 | 40 | 64 | 4 | rect | 318.0/223.9 | 154.7/111.5 | 2.06/2.01 | TileLang |
| 4096 | 64 | 64 | 2 | rect | 249.4/174.4 | 122.9/87.8 | 2.03/1.99 | TileLang |
| 1024 | 16 | 64 | N8 | varlen | 31.6/22.2 | 16.8/11.8 | 1.88/1.88 | TileLang |
| 2560 | 40 | 64 | N4 | varlen | 78.6/55.6 | 41.5/29.7 | 1.89/1.88 | TileLang |
| 4096 | 64 | 64 | N2 | varlen | 134.0/93.4 | 66.8/47.6 | 2.01/1.96 | TileLang |

H800, cs32 (fwd-only ratios: 1.19–1.53× D64, 2.07–2.13× D128 rect, 6.7–7.3× D128 varlen):

| hidden | H | D | B/N | layout | Triton | TileLang | speedup | route |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1024 | 16 | 64 | 8 | rect | 92.1/65.0 | 70.1/51.5 | 1.31/1.26 | TileLang |
| 2560 | 40 | 64 | 4 | rect | 113.5/80.9 | 97.3/73.6 | 1.17/1.10 | TileLang |
| 4096 | 64 | 64 | 2 | rect | 93.4/66.0 | 70.3/51.8 | 1.33/1.28 | TileLang |
| 1024 | 8 | 128 | 4 | rect | 84.6/53.0 | 65.0/48.6 | 1.30/1.09 | TileLang |
| 2560 | 20 | 128 | 2 | rect | 98.5/62.8 | 74.3/56.4 | 1.33/1.11 | TileLang |
| 4096 | 32 | 128 | 1 | rect | 85.1/53.4 | 65.1/48.8 | 1.31/1.09 | TileLang |
| 1024 | 16 | 64 | N8 | varlen | 12.9/9.1 | 10.7/7.9 | 1.21/1.16 | TileLang |
| 2560 | 40 | 64 | N4 | varlen | 30.4/21.6 | 27.6/20.6 | 1.10/1.05 | TileLang |
| 4096 | 64 | 64 | N2 | varlen | 50.2/35.5 | 40.1/29.5 | 1.25/1.20 | TileLang |
| 2560 | 20 | 128 | N4 | varlen | 137.3/75.2 | 35.9/26.7 | 3.83/2.81 | TileLang |
| 4096 | 32 | 128 | N2 | varlen | 213.1/117.0 | 53.6/40.6 | 3.97/2.88 | TileLang |

The varlen-D128 outliers are a Triton-side collapse, visible inside the table: Triton's varlen D128 runs far above its own rect D128 at equal tokens (fwd 63–98 vs. 33–38 ms), while the TileLang fused passes stay proportional (fwd 9–13 ms).

H800, cs64 — the perf gate doing its job (dispatched rows measure the Triton path, hence 1.00):

| hidden | H | D | B/N | layout | Triton | dispatched | speedup | route |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| 1024 | 16 | 64 | 8 | rect | 96.6/70.5 | 96.8/70.5 | 1.00 | Triton (cs64 gated <cc12) |
| 2560 | 40 | 64 | 4 | rect | 131.0/95.6 | 131.1/95.5 | 1.00 | Triton (cs64 gated <cc12) |
| 4096 | 64 | 64 | 2 | rect | 98.9/72.4 | 98.9/72.4 | 1.00 | Triton (cs64 gated <cc12) |
| 1024 | 16 | 64 | N8 | varlen | 13.0/9.4 | 13.0/9.4 | 1.00 | Triton (cs64 gated <cc12) |
| 2560 | 40 | 64 | N4 | varlen | 32.1/23.5 | 32.1/23.5 | 1.00 | Triton (cs64 gated <cc12) |
| 4096 | 64 | 64 | N2 | varlen | 51.3/37.2 | 51.3/37.2 | 1.00 | Triton (cs64 gated <cc12) |

**0.1.13 codegen audit.** The 0.1.12→0.1.13 vectorization-plan refactor (tilelang #2935) scalarized exactly one site: the stream backward's `dgk` row-dot (8× uint4 staged loads + FMA chain → 64 scalar bf16 smem loads/row). Fixed by routing the `h·dh` product through a dedicated fp32 fragment + `T.reduce_sum(dim=1)` — the idiom KDA's `chunk_bwd_dqkg` already uses — which re-vectorizes the loads and lowers the reduction to an `AllReduce` under a named barrier. Same-config generated-CUDA audit of every other kernel: no further drift. Post-fix H800 per-kernel attribution vs. the 0.1.12 baseline: all kernels within ±0.7%; the stream backward recovers fully at cs32-D128/cs64-D64, keeping a documented +1.6% e2e residual at cs32-D64 (0.1.12's zero-communication replicated reduction is not re-expressible through 0.1.13's reducer lowering). PRO 6000 spot checks are flat vs. the pre-upgrade tree.

**Vectorization / schedules.** Wide copies restored pipeline-wide (gate cumsum, WY fwd/bwd, fused h+o, stream bwd); the bulk-copy WY backward path now also covers sm_90 at BT=32 with 128 threads (measured win), alongside the existing cc120 BT≥64 / 256-thread gate.

**Known follow-up (not in this commit).** The small-grid guard was calibrated on 128-chunk chains and is conservative at the 2048-chunk chains CP creates: two rejected CP shapes measure fwd 1.83–2.04×, fwd+bwd 1.15–1.32× with the verifier bypassed, bit-exact. Worst case today is a Triton fallback. I'd recalibrate it with a chain-length-aware boundary sweep as a small separate change.

Gates: backend + CP test files green on both devices, re-run after the rebase (`test_dplr_tilelang.py` 70 passed/2 skipped, `test_cp_dplr.py` 5 passed/2 skipped); the only skips are the device-gated cs64 rows on sm_90.

zhiyuan1i (approved):

sablin39:
Follow-up cleanup on the latest commit (81fab6c0), no behavior or numerics change:

- merged the plain/ctx h+o forward kernel factories into one with `STORE_H_CTX`/`STORE_V_NEW_CTX` constexprs (they were byte-identical apart from the guarded context stores) — net -280 lines;
- renamed `utils.py` → `layout.py` (it only holds `ChunkLayout` + the layout builders) and `chunk_dplr_bwd_wy_into` → `chunk_dplr_bwd_wy_repr_into` for Triton-side naming parity;
- corrected a stale comment: the BT≤16 novec path is a warp-partition constraint of M=16 GEMM tiles, not a tilelang-version bug.

`test_dplr_tilelang.py` (70 passed, 2 skipped) and `test_dplr_delta.py` pass on sm_120 after the merge with current main.

PR #1070, opened 2026-07-29
[Fix] Zero-init Ascend chunk_scaled_dot_kkt output and add solve_tril to CI

zheliuyu:
## Summary

Fix incorrect `chunk_scaled_dot_kkt_fwd` output on Ascend NPU for `chunk_size >= 32`.

The NPU kernel writes only lower-triangular 16×16 sub-blocks into each BT×BT tile and does not cover the upper triangle (e.g. rows 0–15, cols 16–31 when BT=32). Using `torch.empty()` left uninitialized values there, so downstream `solve_tril` varlen tests failed for BT=32/64.

- Initialize the output buffer with `torch.zeros()` instead of `torch.empty()`.
- Add `tests/ops/test_solve_tril.py` to Ascend A2 CI to catch regressions in the GDN/KDA intra-chunk pipeline.

## Test plan

- [x] `pytest tests/ops/test_solve_tril.py` — all tests pass on NPU
- [x] Ascend A2 CI includes `test_solve_tril.py`

### DIFF
--- a/.github/workflows/ascend-a2-ci.yml
+++ b/.github/workflows/ascend-a2-ci.yml
@@ -100,7 +100,7 @@ jobs:
           assert IS_NPU, "Expected NPU backend (IS_NPU=True)"
           PY
 
-      - name: Run fla/modules and fla/ops/utils, gdn, kda, attnres tests
+      - name: Run tests/modules and tests/ops/utils, gdn, kda, attnres, solve_tril
         shell: bash
         env:
           FLA_NPU_XDIST: "1"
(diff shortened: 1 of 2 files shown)


zhiyuan1i (approved):
LGTM. Verified the NPU kernel only stores lower-triangular 16×16 sub-blocks (`for c in range(s + 1)`), so the upper triangle is uninitialized with `torch.empty` for BT≥32; `solve_tril` itself never reads the upper triangle, but `test_solve_tril_varlen`'s `torch.inverse` reference consumes the full tile — consistent with the reported failure. Two notes: (1) heads-up that #1023 rewrites this kernel to store the full tile (upper zeroed in-kernel), so whoever merges second will need a trivial rebase, and the `zeros` init becomes redundant there; (2) optional follow-up: writing zeros to the upper sub-blocks in-kernel would avoid the extra memset, but fine as-is given #1023 is in flight.

PR #1076, opened 2026-07-30
[Fix] Correct GRPO loss with old log probabilities

### LINKED ISSUE #1073: [Bug] grpo_loss_with_old_logps double-negates the loss (wrong sign)
`grpo_loss_with_old_logps` in `fla/modules/grpo.py` negates the loss twice, so the returned loss has the wrong sign.

```python
# fla/modules/grpo.py:415-419
token_loss = -(torch.min(advantages * importance_weights, advantages *
               importance_weights_clipped) - beta * per_token_kl) * completion_mask

loss = -token_loss.sum() / completion_mask.sum()
```

`token_loss` is already the per-token GRPO loss; the second `-` flips it back, giving `loss = +(min(A*w, A*w_clip) - beta*KL)` averaged over valid tokens. Minimizing that pushes the policy away from high-advantage actions and increases KL to the reference.

`grpo_loss_torch` (line 351) and the TRL snippet in the module docstring (line 45) both apply the negation only once.

Fix:

```python
loss = token_loss.sum() / completion_mask.sum()
```

The function is currently unused and untested (not exported from `fla/modules/__init__.py`, no references elsewhere), so nothing is affected today.

Two smaller issues in the same function:

- Line 405: `torch.arange(logits_to_keep)[None, :] >= 0` is `True` everywhere, so the mask reduces to `pad_mask` and `logits_to_keep` has no effect.
- Line 390: `rewards.view(-1, B)` with `B = logps.shape[0]` and `rewards` of shape `[B]` yields `(1, B)`, so the advantage mean/std are taken over the whole batch rather than per-prompt generation groups — despite the `# B,num_generations` comment.

Happy to send a PR if useful.

dongchany:
Partially fixes #1073.

## What changed

- Remove the second negation when reducing `token_loss` in `grpo_loss_with_old_logps`.
- Add a pure PyTorch reference implementation and coverage for unclipped, clipped, KL, padding-mask, float32, and bfloat16 paths.
- Add an explicit gradient-direction regression test.

`token_loss` is already the negative PPO-style objective. Negating it again at reduction time turns minimization into maximization and reverses the policy-gradient direction.

## Compatibility

- No public API or function signature changes.
- `logits_to_keep` and reward-grouping semantics are intentionally left unchanged and should be discussed separately.

## Validation

- `pytest -q tests/modules/test_grpo.py`: 17 passed, 4 existing hardware skips, rerun on upstream `3adcb3c5`.
- New `grpo_loss_with_old_logps` cases: 9 passed.
- `pre-commit run --files fla/modules/grpo.py tests/modules/test_grpo.py`: passed.

### DIFF
--- a/fla/modules/grpo.py
+++ b/fla/modules/grpo.py
@@ -416,6 +416,6 @@ def grpo_loss_with_old_logps(
                    importance_weights_clipped) - beta * per_token_kl) * completion_mask
 
     # Calculate the final loss by summing the token losses and normalizing by the number of valid tokens
-    loss = -token_loss.sum() / completion_mask.sum()
+    loss = token_loss.sum() / completion_mask.sum()
 
     return loss
--- a/tests/modules/test_grpo.py
+++ b/tests/modules/test_grpo.py
@@ -8,8 +8,111 @@
 import pytest
 import torch
 
-from fla.modules.grpo import fused_grpo_loss, grpo_loss_torch
+from fla.modules.grpo import fused_grpo_loss, grpo_loss_torch, grpo_loss_with_old_logps
 from fla.utils import IS_NVIDIA_HOPPER, assert_close, device, device_torch_lib
+
+
+def grpo_loss_with_old_logps_torch(
+    logps: torch.Tensor,
+    ref_logps: torch.Tensor,
+    old_logps: torch.Tensor,
+    pad_mask: torch.Tensor,
+    logits_to_keep: int,
+    rewards: torch.Tensor,
+    beta: float,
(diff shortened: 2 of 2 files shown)


zhiyuan1i (approved):

zhiyuan1i (approved):
Verified against TRL's sign convention: `token_loss` is already the negated objective, the extra negation in the reduction flipped the gradient direction. Fix is correct, tests cover both values and gradient signs. (Edited the body to 'Partially fixes #1073' since the other two points there are still open.)

PR #1085, opened 2026-07-31
[Fix] Use the full token stride for DPLR safe-gate offsets

### LINKED ISSUE #1084: [Bug] Safe-gate DPLR kernels use the wrong midpoint offset for multi-head inputs
## Problem

The `safe_gate=True` TensorCore path in the DPLR chunk operator uses the wrong address for the midpoint gate offset when the input has more than one head. The tensors use `[B, T, H, K]` storage, so advancing one token requires an `H * K` element stride. Both the forward and backward intra-chunk kernels currently advance by `K` after anchoring at the current head.

The affected expressions are:

- `fla/ops/generalized_delta_rule/dplr/chunk_A_fwd.py:229`
- `fla/ops/generalized_delta_rule/dplr/chunk_A_bwd.py:293`

```text
gi[b, token, head, k]
          token stride = H * K

current head base + midpoint * K     -> another token/head
current head base + midpoint * H * K -> same head at the midpoint
```

With `H > 1`, the short stride can select a different head's cumulative gate. That makes the numerical stabilization offset head-dependent in the wrong way and can corrupt the intra-chunk scores and their gradients. Single-head inputs do not expose the addressing error.

## Reproduction

The following regression test uses `B=1, T=16, H=3, K=V=16`, distinct per-head inputs, and distinct per-head gates. It exercises the forward intra-chunk score, the backward query gradient, and the public final-state gradient path:

```bash
TMPDIR=/home/morluto/.cache/fla-audit-tmp \
TRITON_CACHE_DIR=/home/morluto/.cache/fla-audit-triton-cache \
/home/morluto/.cache/fla-audit-venv/bin/python -m pytest \
  tests/ops/test_dplr_delta.py::test_chunk_safe_gate_multihead_gradients -q
```

The corrected implementation passes. Running the same test against the current `* K` expressions fails in the backward assertion: the first query gradient becomes `NaN` and another head's gradient becomes zero. The forward score also moves away from the analytical value for the selected midpoint pair.

## Expected behavior

For each head, the midpoint offset must address `gi[b, midpoint, head, :]`. The forward and backward safe-gate kernels should agree with the corresponding analytical intra-chunk score and gradient, and the public operator should return finite outputs and finite gradients to its inputs and initial state.

## Suggested fix

Use the full token stride, `H * K`, in both midpoint-offset expressions and retain the multi-head forward/backward regression test.

morluto:
Fixes #1084.

## Problem

The DPLR `safe_gate=True` TensorCore kernels center gate values around a midpoint vector. Inputs are stored as `[B, T, H, K]`, so the stride from one token to the next is `H * K`. The midpoint offset in both kernels advanced by only `K`, which can select a different token/head whenever `H > 1`.

```text
gi[b, token, head, k]

current head base + midpoint * K     -> wrong token/head
current head base + midpoint * H * K -> same head at midpoint
```

The bad offset changes the numerical stabilization used by the intra-chunk score and its backward computation. Single-head inputs do not expose the addressing error.

## Change

- Use the full `H * K` token stride in the forward TensorCore midpoint load.
- Apply the same correction in the backward TensorCore midpoint load.
- Add a multi-head regression with distinct per-head inputs and gates. It checks the analytical midpoint score, the corresponding query gradient, and finite gradients through the public final-state path.

## Validation

```text
pytest tests/ops/test_dplr_delta.py::test_chunk_safe_gate_multihead_gradients -q
1 passed

pytest tests/ops/test_dplr_delta.py::test_chunk_safe_gate_multihead_gradients tests/ops/test_dplr_delta.py::test_chunk -q
12 passed
```

The regression was also run against the pre-fix `* K` expressions. It failed in the backward assertion because one head produced `NaN` and another produced a zero query gradient. The same test passes with the corrected stride.

The change is limited to multi-head `safe_gate=True` DPLR TensorCore intra-chunk paths; single-head and non-safe-gate paths are unchanged.

### DIFF
--- a/fla/ops/generalized_delta_rule/dplr/chunk_A_bwd.py
+++ b/fla/ops/generalized_delta_rule/dplr/chunk_A_bwd.py
@@ -290,7 +290,7 @@ def chunk_dplr_bwd_kernel_intra_tensorcore(
     valid_len = min(T_len - i_t * BT, BT)
     mid_idx = valid_len // 2
     m_k = tl.arange(0, BK) + i_k * BK < K
-    p_offset = gi + offset_base_k + (i_t * BT + mid_idx) * K + tl.arange(0, BK) + i_k * BK
+    p_offset = gi + offset_base_k + (i_t * BT + mid_idx) * H * K + tl.arange(0, BK) + i_k * BK
     b_offset = tl.load(p_offset, mask=m_k, other=0.0).to(tl.float32)
 
     # Q, K, A, B, Gates: [BT, BK]
--- a/fla/ops/generalized_delta_rule/dplr/chunk_A_fwd.py
+++ b/fla/ops/generalized_delta_rule/dplr/chunk_A_fwd.py
@@ -226,7 +226,7 @@ def chunk_dplr_fwd_A_kernel_intra_tensorcore(
     # Load the offset vector from Global Memory
     # p_offset points to gi[i_t*BT + mid_idx, :]
     m_k = tl.arange(0, BK) < K
-    p_offset = gi + offset_base + (i_t * BT + mid_idx) * K + tl.arange(0, BK)
+    p_offset = gi + offset_base + (i_t * BT + mid_idx) * H * K + tl.arange(0, BK)
     b_offset = tl.load(p_offset, mask=m_k, other=0.0).to(tl.float32)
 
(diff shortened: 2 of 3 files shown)


zhiyuan1i (approved):
Thanks!

PR #1090, opened 2026-08-01
[Feature] Add Ascend NPU backend for KDA fused recurrent forward

sunyi0505:
### To fix the error: FAILED tests/ops/test_kda.py::test_fused_recurrent_state_v_first[B2-T512-H3-D60-scale1-gate_logit_normalizer1-torch.float32] - AssertionError: o: NaN detected in tri

Root cause:
        Fused_recurrent is not implemented on the NPU, so the CUDA kernel will be invoked. The test_fused_recurrent_state_v_first test fails when D equals 60. This occurs because K=60 is padded to BK=64, and the loading of g is performed without a mask, resulting in the reading of incorrect decay values from adjacent memory addresses. This can lead to the generation of infinity values that subsequently turn into NaNs along the execution path where state_v_first is set to True.

solution:
        Added fla/ops/kda/backends/triton_ascend/fused_recurrent.py to implement the Triton-Ascend kernel of fused_recurrent_kda_fwd for Ascend NPU, with functions consistent with the generic Triton implementation (including varlen, continuous batching, spec decoding, in-kernel gate/QK L2 norm, beta sigmoid, state_v_first and other features).
        Registered fused_recurrent_kda_fwd_verifier and fused_recurrent_kda_fwd within TritonAscendKDABackend to enable the backend to invoke the Ascend execution path.
        Modified the decorator of fused_recurrent_kda_fwd in fla/ops/kda/fused_recurrent.py from @torch.compiler.disable to @dispatch("kda"), which automatically dispatches execution to the Ascend implementation on NPUs while retaining the original Triton kernel for GPUs.

### To fix fused_layernorm_gate error

When recompute_output=True is enabled for the backward computation of layer_norm_gated on Ascend, the kernel has to retain the pre-gated b_y value to execute tl.store(y), leading to higher UB usage compared with standard backward passes. The original implementation adopted the default memory multiplier for backward computation, which results in an excessively large BT value under large dimension D scenarios and may trigger UB overflow.

This PR calibrates the UB budget independently for the recompute execution path and passes the recompute_output flag into the launch configuration, allowing the backward computation to automatically adopt a smaller row tile size (BT).

### DIFF
--- a/.github/workflows/nvidia-h100.yml
+++ b/.github/workflows/nvidia-h100.yml
@@ -53,8 +53,9 @@ jobs:
               owner, repo, pull_number: pr.number, per_page: 100,
             });
             // Safe set: triton-ascend backend files (except __init__.py, which
-            // GPU imports unconditionally) and Ascend workflow definitions.
-            const safePattern = /^fla\/(.+\/)?backends\/triton_ascend\/(?!.*__init__\.py$)|^\.github\/workflows\/ascend-/;
+            // GPU imports unconditionally), Ascend workflow definitions, and
+            // fla-ascend-performance agent skill (NPU profiling/optimization docs).
+            const safePattern = /^fla\/(.+\/)?backends\/triton_ascend\/(?!.*__init__\.py$)|^\.github\/workflows\/ascend-|^\.agents\/skills\/fla-ascend-performance\//;
             const unsafe = files.filter(f => !safePattern.test(f.filename)).map(f => f.filename);
             if (unsafe.length > 0) {
               core.info('GPU-affecting files changed: ' + unsafe.join(', '));
--- a/fla/modules/backends/triton_ascend/fused_norm_gate.py
+++ b/fla/modules/backends/triton_ascend/fused_norm_gate.py
@@ -29,9 +29,12 @@ from fla.utils.ascend_ub_manager import (
 # Bwd stores dg early before the dx path. Calibrated on Ascend910 (192 KiB UB,
 # 0.85 margin): fwd BT=8 @ D=1024 (mult=3); bwd BT=4 @ D=1024 (mult=8).
 # Do not lower bwd mult below ~6 without re-validating (BT=8 bwd overflows).
+# Recompute-output (norm+linear bwd) keeps b_y live for tl.store(y);
+# bwd BT=2 @ D=1024 (mult=11). Do not lower without re-validating.
 _BD_MEM_MULT = 6.0
 _FWD_MEM_MULT = 3.0
 _BWD_MEM_MULT = 8.0
+_BWD_RECOMPUTE_MEM_MULT = 11.0
 _UB_SAFETY_MARGIN = 0.85
 # Legacy byte cap when UB capacity cannot be detected (65536 // fp32).
 _FALLBACK_MAX_BD = 65536 // 4
@@ -40,6 +43,7 @@ _MAX_BT = 128
 _LARGE_BD = 2048
 _LARGE_BD_FWD_MEM_MULT = 4.0
 _LARGE_BD_BWD_MEM_MULT = 12.0
+_LARGE_BD_BWD_RECOMPUTE_MEM_MULT = 16.5
 
 # ACTIVATION constexpr: 0 = swish/silu, 1 = sigmoid
 _ACTIVATION_SWISH = 0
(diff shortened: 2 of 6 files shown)


sunyi0505:

zhiyuan1i (changes requested):
Thanks @sunyi0505 — feature parity with the GPU version is impressively complete (varlen, continuous batching, spec decoding, in-kernel gate / QK L2 norm, beta sigmoid, state_v_first), the int64 offset handling is clean, and the grid >65535 chunking is a nice touch. Two things I'd like addressed before merge:

1. **`@input_guard` silently copies `initial_state` / `out`** — a blocker for the continuous-batching scenario this backend targets. `input_guard` calls `.contiguous()` on every tensor argument. With `inplace_final_state=True` (the default) and a non-contiguous `initial_state` (e.g. a strided view into a vLLM-style state pool), the guard silently copies it, the kernel writes the new state into the copy, and the caller's pool is never updated — silent wrong result. The GPU wrapper deliberately has no guard. Please switch to `@input_guard(no_guard_contiguous={'initial_state', 'out'})` and explicitly assert contiguity for these two instead (fail loudly rather than silently copying).

2. **The new op is not covered by the backend routing test.** `fused_recurrent_kda_fwd` is missing from `_TRITON_ASCEND_KDA_OPS` in `tests/ops/test_kda.py`, so `test_triton_ascend_backend_routing` never asserts the Ascend kernel actually executes. If the verifier ever silently falls back to the GPU kernel (which runs fine on NPU — the pre-PR behavior), CI stays green and the new kernel becomes dead code. Please add it to the spy list and assert a hit.

Minor:
- The NPU wrapper drops `**kwargs` from the GPU signature; legacy kwargs now raise TypeError on NPU while GPU silently ignores them. Please keep the signatures aligned.
- The body attributes the `test_fused_recurrent_state_v_first` NaN to the fused_norm_gate UB fix, but that KDA test does not import fused_norm_gate — cross-test device-state pollution in the same pytest worker seems more likely. Worth clarifying so nobody chases the wrong lead.
- No perf numbers yet; consider retitling `[Perf]` → `[Feature]`, or add a short old-vs-new path comparison on NPU.

The fused_norm_gate UB fix itself looks correct (the budget math checks out) — fine to keep here, though a separate PR would have been cleaner.

sunyi0505:
> Thanks @sunyi0505 — feature parity with the GPU version is impressively complete (varlen, continuous batching, spec decoding, in-kernel gate / QK L2 norm, beta sigmoid, state_v_first), the int64 offset handling is clean, and the grid >65535 chunking is a nice touch. Two things I'd like addressed before merge:
> 
> 1. **`@input_guard` silently copies `initial_state` / `out`** — a blocker for the continuous-batching scenario this backend targets. `input_guard` calls `.contiguous()` on every tensor argument. With `inplace_final_state=True` (the default) and a non-contiguous `initial_state` (e.g. a strided view into a vLLM-style state pool), the guard silently copies it, the kernel writes the new state into the copy, and the caller's pool is never updated — silent wrong result. The GPU wrapper deliberately has no guard. Please switch to `@input_guard(no_guard_contiguous={'initial_state', 'out'})` and explicitly assert contiguity for these two instead (fail loudly rather than silently copying).
> 2. **The new op is not covered by the backend routing test.** `fused_recurrent_kda_fwd` is missing from `_TRITON_ASCEND_KDA_OPS` in `tests/ops/test_kda.py`, so `test_triton_ascend_backend_routing` never asserts the Ascend kernel actually executes. If the verifier ever silently falls back to the GPU kernel (which runs fine on NPU — the pre-PR behavior), CI stays green and the new kernel becomes dead code. Please add it to the spy list and assert a hit.
> 
> Minor:
> 
> * The NPU wrapper drops `**kwargs` from the GPU signature; legacy kwargs now raise TypeError on NPU while GPU silently ignores them. Please keep the signatures aligned.
> * The body attributes the `test_fused_recurrent_state_v_first` NaN to the fused_norm_gate UB fix, but that KDA test does not import fused_norm_gate — cross-test device-state pollution in the same pytest worker seems more likely. Worth clarifying so nobody chases the wrong lead.
> * No perf numbers yet; consider retitling `[Perf]` → `[Feature]`, or add a short old-vs-new path comparison on NPU.
> 
> The fused_norm_gate UB fix itself looks correct (the budget math checks out) — fine to keep here, though a separate PR would have been cleaner.

All issues have addressed. If the bug fixes of fused_norm_gate and fused_recurrent are split into two individual PRs, the CI checks for both PRs will fail. For this reason, we submit all fixes within one unified PR. Thanks for your suggestions!

zhiyuan1i (approved):
All comments addressed — the `no_guard_contiguous` + explicit contiguity checks, the routing-test coverage, and the signature alignment all look right. CI is green on both A2 and H100. Thanks for the quick turnaround!

PR #1091, opened 2026-08-01
[CI] Run Ascend A2 CI on NPU-related file changes

zheliuyu:
## Summary

Both Ascend A2 workflows now run for PRs that touch NPU-related files, detected by a ~5s `changes` job listing PR files via the GitHub API, instead of relying solely on `npu`/`ascend` title keywords. 

The file set: `fla/**/backends/triton_ascend/`, `.github/workflows/ascend-*`, `pyproject.toml`, `fla/utils/`, `tests/conftest.py`; plus `benchmarks/` for the benchmark workflow. Title gating, push, and `workflow_dispatch` behavior are unchanged. Mirrors #1063's file-based detection in the opposite direction. Also adds `timeout-minutes: 150` to both A2 jobs.

## Test plan

- [x] Run .github/workflows/ascend-a2-ci.yml

- [x] .github/workflows/ascend-a2-benchmark-ci.yml

## Breaking changes

N/A

## Checklist

### DIFF
--- a/.github/workflows/ascend-a2-benchmark-ci.yml
+++ b/.github/workflows/ascend-a2-benchmark-ci.yml
@@ -22,14 +22,53 @@ on:
       - ".github/workflows/*.yml"
 
 jobs:
+  changes:
+    name: Detect NPU-related changes
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    outputs:
+      npu_related: ${{ steps.detect.outputs.npu_related }}
+    steps:
+      - name: Detect NPU-related diff
+        id: detect
+        uses: actions/github-script@v9.0.0
+        with:
+          script: |
+            if (context.eventName !== 'pull_request') {
+              core.setOutput('npu_related', 'false');
+              return;
+            }
+            const pr = context.payload.pull_request;
+            const { owner, repo } = context.repo;
+            const files = await github.paginate(github.rest.pulls.listFiles, {
+              owner, repo, pull_number: pr.number, per_page: 100,
+            });
+            // NPU-related set: triton-ascend backend files, Ascend workflow
+            // definitions, the files that define the NPU CI stack itself:
+            //   pyproject.toml    - pins the [npu] extra (torch_npu / triton-ascend)
+            //   fla/utils/        - IS_NPU / device_platform detection, ascend_ub_manager
+            //   tests/conftest.py - FLA_NPU_XDIST xdist worker -> device mapping
+            // and the benchmark harness invoked by this workflow (benchmarks/).
+            // Unlike the H100 'npu_only' detection, __init__.py is NOT excluded
+            // here: a change to it affects the NPU backend too, so it should
+            // trigger this CI.
+            const npuPattern = /^fla\/(.+\/)?backends\/triton_ascend\/|^\.github\/workflows\/ascend-|^pyproject\.toml$|^fla\/utils\/|^tests\/conftest\.py$|^benchmarks\//;
+            const related = files.filter(f => npuPattern.test(f.filename)).map(f => f.filename);
+            if (related.length > 0) {
+              core.info('NPU-related files changed: ' + related.join(', '));
+            }
+            core.setOutput('npu_related', related.length > 0 ? 'true' : 'false');
+
   benchmark-a2-npu-pytorch-2-7:
     name: Benchmark A2 NPU (PyTorch 2.7.1)
+    needs: changes
     if: |
       github.event_name == 'workflow_dispatch' ||
       github.event_name == 'push' ||
       (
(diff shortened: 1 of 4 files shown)


zheliuyu:
NPU CI failures are caused by the recent changes in #1071 not triggering the necessary checks, which is exactly the reason why this PR was opened. 

We need to wait for #1090 to be merged.

zheliuyu:
## For "[Fix] GRPO old-logps NPU compile bypass and GDN bwd varlen skip"

### Summary

Two unrelated Ascend gaps surfaced after recent upstream changes; this PR unblocks local/CI runs without changing public APIs.

**GRPO (`fla/modules/backends/triton_ascend/__init__.py`)**  
After #1076, `grpo_loss_with_old_logps` in `fla/modules/grpo.py` is correct in eager mode, but `@torch.compile(fullgraph=True)` on NPU goes through torch_npu Inductor: loss/grad tests fail numerically and the gradient-direction case fails to compile (symbolic reductions). When the triton_ascend backend loads (existing `grpo` → `backends` import order), install an NPU-only `torch.compile` shim that leaves `grpo_loss_with_old_logps` eager; all other `torch.compile` calls are unchanged.

**GDN (`tests/ops/test_gdn_kernels.py`)**  
#1077 added `test_chunk_gated_delta_rule_bwd_dhu_varlen_many_documents` (2050 packed documents). CUDA grid limits were fixed in `fla/ops/common/chunk_delta_h.py`; Ascend still uses `triton_ascend` `bwd_dhu`, which AICore-times out on this stress test. Skip that test on NPU until a proper backend launch/gate fix is designed; the matching fwd varlen test remains enabled on NPU.

### Test plan

- [x] `pytest tests/modules/test_grpo.py` on Ascend NPU
- [x] `pytest tests/ops/test_gdn_kernels.py` on Ascend NPU (54 passed, 1 skipped, bwd varlen stress)

zheliuyu:
@zhiyuan1i Maybe we need a better CI trigger strategy. For now, this is our temp fix. PTAL. Thanks!

zhiyuan1i:
Thanks @zheliuyu — direction looks good, happy to land this as the temp fix. One hole: the allowlist misses `fla/modules/`, but pure-torch modules do run on NPU and the A2 suite tests them — #1071 (`layer_norm_gated`) and #1076 (`grpo.py`) broke NPU CI exactly this way and still wouldn't trigger it under this scheme. Please add `fla/modules/` (or note explicitly the gate is best-effort with main as the backstop).

Longer term, dependency-based test selection (like `scripts/find_dependent_tests.py` on the H100 side) beats a curated allowlist — worth a follow-up issue.

zheliuyu:
@zhiyuan1i Thanks for the review — one thing I’d like to align on before we expand the allowlist.

Adding paths like `fla/modules/` would catch more regressions on PRs, but not every contributor has Ascend hardware to reproduce A2 failures locally. For many `fla/modules/` changes, `tests/modules/` on CUDA can pass while A2 still fails (e.g. NPU `torch.compile` / `Inductor`). If we trigger A2 on those PRs, authors without an NPU may be blocked for a long time with no realistic way to fix red CI on their own.

Mitigations I can think of: get help from folks with A2 runners, or case-by-case maintainer judgment on whether to merge despite a known NPU-only failure (with a follow-up tracked). That’s why I’d like to agree on the workflow before landing `fla/modules/` in the allowlist — what do you prefer? Should we still add `fla/modules/` in this PR, or wait until we have something like H100-style dependent test selection or a clearer merge policy for NPU CI?

zheliuyu:
> @zhiyuan1i Thanks for the review — one thing I’d like to align on before we expand the allowlist.
> 
> Adding paths like `fla/modules/` would catch more regressions on PRs, but not every contributor has Ascend hardware to reproduce A2 failures locally. For many `fla/modules/` changes, `tests/modules/` on CUDA can pass while A2 still fails (e.g. NPU `torch.compile` / `Inductor`). If we trigger A2 on those PRs, authors without an NPU may be blocked for a long time with no realistic way to fix red CI on their own.
> 
> Mitigations I can think of: get help from folks with A2 runners, or case-by-case maintainer judgment on whether to merge despite a known NPU-only failure (with a follow-up tracked). That’s why I’d like to agree on the workflow before landing `fla/modules/` in the allowlist — what do you prefer? Should we still add `fla/modules/` in this PR, or wait until we have something like H100-style dependent test selection or a clearer merge policy for NPU CI?


@zhiyuan1i After discussing this, we’ve added fla/modules/ to the NPU CI path allowlist in this PR. We’ll follow up soon with a concrete plan for the contributor workflow issues raised above (e.g. fixing A2 failures without local Ascend hardware).

zhiyuan1i (approved):

PR #1094, opened 2026-08-02
[GDN] Enable FlashQLA forward dispatch on SM120

Akrao9:
## Motivation

[FlashQLA v0.1.2](https://github.com/QwenLM/FlashQLA) added an SM120 (consumer/workstation Blackwell)
forward kernel and its README now lists `SM90, SM100, SM103 or SM120` as supported. FLA already treats
SM120 as Blackwell (`IS_NVIDIA_BLACKWELL` covers capability 10 and 12), but `FlashQLABackend`'s verifier
still gates on `IS_NVIDIA_HOPPER or IS_NVIDIA_SM100`, so every SM120 call is rejected with
`"FlashQLA requires NVIDIA SM90 or SM100"` and silently falls back to the Triton path — including
pure-inference and frozen-GDN workloads that FlashQLA can serve today.

## What this does

Adds `IS_NVIDIA_SM120` (capability major == 12) to `fla/utils/_device.py` and accepts SM120 in the
FlashQLA verifier **only for the subset that actually works there**:

1. **Forward only.** Backward is unimplemented upstream — `flash_qla` raises `NotImplementedError` from
   `chunk_gated_delta_rule_bwd` and `intra_card_cp_preprocess_bwd` on SM120. The verifier accepts SM120
   only when no backward can follow: `torch.is_grad_enabled()` is `False`, or none of
   `q/k/v/g/beta/initial_state` requires grad. This is why the gate cannot simply become
   `IS_NVIDIA_BLACKWELL` — that would dispatch trainable calls into FlashQLA and fail in backward.

2. **bfloat16 only.** FlashQLA's `blackwell_sm120` forward kernel emits `tl::pack_float16x4` on
   `cutlass::half_t` in its masked tail-store, which does not compile:

   ```
   tvm_kernels.cu(626): error: no suitable user-defined conversion
                        from "cutlass::half_t" to "half" (aka "__half") exists
       tl::store_global_256(... tl::pack_float16x4(broadcast_var_12, ...))
   nvcc --cubin -O3 -arch=sm_120a -std=c++17 ...
   ```

   float16 is therefore rejected on SM120 so those calls keep the working Triton path rather than
   hitting a codegen error. bf16 is unaffected, and float16 still dispatches on SM90/SM100/SM103.

Also updates the now-inaccurate `"SM90 or SM100"` rejection message. Hopper / SM100 / SM103 dispatch is
otherwise unchanged (SM103 is capability 10.3, already covered by `IS_NVIDIA_SM100`).

## Tests

- `test_flash_qla_verifier_rejects` — updated for the new message, with `IS_NVIDIA_SM120` pinned `False`
  so the pre-existing branches are exercised as before.
- `test_flash_qla_verifier_sm120_accepts_forward` — grad-free call accepted; requires-grad inputs under
  `torch.no_grad()` accepted (the frozen-model case); an `initial_state` requiring grad rejected.
- `test_flash_qla_verifier_sm120_rejects_backward` — trainable call rejected on SM120, still accepted on
  Hopper.
- `test_flash_qla_verifier_sm120_rejects_float16` — fp16 rejected on SM120, still accepted on Hopper.
- `test_flash_qla_chunk_forward_only` — forward parity against `naive_recurrent_gated_delta_rule` under
  `no_grad`, incl. a GVA (`HV > H`) shape, behind a new `_SKIP_FLASH_QLA_FWD` marker.

The existing fwd+bwd parity tests keep the original `_SKIP_FLASH_QLA` marker (SM90/SM100 only), so they
stay skipped on SM120 rather than intentionally crashing.

## Validation

```
GPU:   NVIDIA RTX PRO 6000 Blackwell Server Edition
SM:    12.0
Torch: 2.11.0+cu128
```

`pytest tests/ops/test_gdn.py -k flash_qla -q -rs`:

```
7 passed, 13 skipped, 79 deselected
SKIPPED [12] FlashQLA backward requires an SM90/SM100 GPU and the flash_qla package
SKIPPED  [1] FlashQLA's SM120 forward kernel does not compile for float16
```

Forward parity vs. the naive reference, with backend dispatch confirmed in the log by
`[FLA Backend] gated_delta_rule.chunk_gated_delta_rule -> flash_qla`:

| shape | dtype | `o` diff / ratio | `ht` diff / ratio |
|---|---|---|---|
| B1-T1024-H4-HV4-D128  | bf16 | 0.000551 / 0.005179 | 0.002821 / 0.003155 |
| B2-T2048-H8-HV8-D128  | bf16 | 0.000917 / 0.005095 | 0.005449 / 0.003784 |
| B1-T1024-H2-HV8-D128 (GVA) | bf16 | 0.000709 / 0.005005 | 0.002834 / 0.003674 |

The 12 existing fwd+bwd parity cases skip on SM120 rather than crashing, and the fp16 forward case
skips via the new dtype gate — both as intended.

## Forward benchmark (Qwen3-Next-shaped GDN, H=16 HV=32 D=128, bf16, `no_grad`)

Each shape asserts the backend actually flipped before timing, so a silent Triton-vs-Triton comparison
would raise rather than report a fake 1.00x.

| B | T | Triton (ms) | FlashQLA (ms) | speedup | max &#124;Δo&#124; | max &#124;Δht&#124; |
|---|---|---|---|---|---|---|
| 1 | 2048  | 0.197 | 0.148 | **1.33x** | 0.0010 | 0.0045 |
| 1 | 8192  | 0.872 | 0.409 | **2.13x** | 0.0010 | 0.0042 |
| 1 | 32768 | 3.539 | 1.351 | **2.62x** | 0.0010 | 0.0038 |
| 4 | 2048  | 0.881 | 0.375 | **2.35x** | 0.0010 | 0.0050 |
| 4 | 8192  | 3.512 | 1.419 | **2.48x** | 0.0010 | 0.0054 |

The gain grows with sequence length, which is the regime that matters for long-context prefill.

No kernel numerics and no public API change — this only restores the compatibility FlashQLA already
advertises for the subset that compiles and runs, so no RFC.


zhiyuan1i (approved):
Verified against upstream FlashQLA v0.1.2: the SM120 forward kernel and README arch list check out, backward indeed raises NotImplementedError on SM120 (chunk/__init__.py:107), and the fp16 codegen issue is consistent with the upstream dtype chain (and the gate is a safe conservative rejection regardless). The grad-based forward-only gating is sound. One note for a follow-up: `IS_NVIDIA_SM120` (major==12) also matches 12.1 (GB10), but upstream only accepts "12.0" and raises ValueError at import — on GB10 with flash_qla installed this turns a clean Triton fallback into a runtime error. Consider gating on `get_device_capability() == (12, 0)` later; not a blocker.

PR #1103, opened 2026-08-05
[Fix] Mask out-of-bounds accesses in RWKV7 channel mixing kernels

### LINKED ISSUE #1102: [Bug] RWKV7 channel mixing: three kernels read and write past the tensor end when the element count is not a multiple of the block size
### Checklist

### Describe the Bug

`fla/ops/rwkv7/channel_mixing.py` sizes its launch grids with `cdiv(numel, BLOCK_SIZE)` — anticipating a partial last block — but three of its kernels do not mask the tail lanes. Any shape whose element count is not a multiple of the block size reads and, worse, **writes** past the end of the tensors involved. All three sites are reachable from the public `channel_mixing_rwkv7` (the RWKV7 FFN):

1. `rwkv_channel_mixing_pow_and_relu` (forward): no mask on its load or store, `BLOCK_SIZE` hardcoded to 4096. A launch at `numel = 4900` stores 3292 floats past the end of the output.

https://github.com/fla-org/flash-linear-attention/blob/2501ac83fd2c1223f3f0393563f2d58255df9b5d/fla/ops/rwkv7/channel_mixing.py#L88-L94

2. `relu_square_bwd_kernel` (backward): same omission — no element count, no masks — writing past the end of `dk` inside `rwkv_channel_mixing_bwd`.

https://github.com/fla-org/flash-linear-attention/blob/2501ac83fd2c1223f3f0393563f2d58255df9b5d/fla/ops/rwkv7/channel_mixing.py#L180-L191

3. `rwkv_mix_bwd_kenel` (backward): this kernel *does* compute `is_valid` and applies it on three of its four accesses, but the `dx_prev` store is masked by `is_first_step` alone. `is_first_step` is a position predicate, not a bounds predicate: for dead tail lanes `seq_idx` cycles back to 0, so they pass it and store at `batch_idx >= batch_size`, past the end of `dx_prev` — which under `inplace=True` is the caller's own `x_prev` buffer.

https://github.com/fla-org/flash-linear-attention/blob/2501ac83fd2c1223f3f0393563f2d58255df9b5d/fla/ops/rwkv7/channel_mixing.py#L235-L242

The corruption is silent (no launch error) and lands on whatever the caching allocator placed after those tensors. In our runs, the first `backward()` on a fresh autotune cache corrupted gradients that had been computed just before by an unrelated reference path: max abs gradient error 6e+01 against the same computation with masks (8e-06). It goes unnoticed at common RWKV7 configurations because `n_embd` and `dim_ffn` are typically powers of two and every element count comes out block-aligned — e.g. `test_channel_mixing_gradients` uses (B, T, n_embd, dim_ffn) = (2, 1024, 1024, 4096), all numels multiples of 8192.

Sentinel-filled guard bands around the tensors each site writes through (element counts chosen so no block size in [128, 8192] divides them):

```python
import torch
import fla.ops.rwkv7.channel_mixing as cm

torch.manual_seed(42)
SENT = 777.0
B, T, n_embd, dim_ffn = 1, 7, 200, 700
N = B * T * dim_ffn                    # 4900
W = (N + 4095) // 4096 * 4096          # 8192, the window the 4096 grid covers

# site 1: forward relu-square, in place on a view of a sentinel-filled buffer
buf = torch.full((2 * W,), SENT, device='cuda')
cm.rwkv_relu_and_square_fwd(buf[:N].view(B, T, dim_ffn), inplace=True)
print('site 1  pow_and_relu     cells written past the end:', (buf[N:] != SENT).sum().item())

# site 2: backward relu-square, launched exactly as rwkv_channel_mixing_bwd launches it
buf = torch.full((2 * W,), SENT, device='cuda')
dk, k1_K = buf[:N], buf[W:W + N]
cm.relu_square_bwd_kernel[((N + 4095) // 4096,)](dk, k1_K, BLOCK_SIZE=4096)
print('site 2  relu_square_bwd  cells written past dk:', (buf[N:W] != SENT).sum().item())

# site 3: full backward with inplace=True, so dx_prev is the caller's x_prev
big = torch.full((1 << 20,), SENT, device='cuda')
x = torch.randn(B, T, n_embd, device='cuda')
x_prev = big[:B * n_embd].view(B, n_embd)
x_k = torch.randn(1, 1, n_embd, device='cuda')
Kw = torch.randn(n_embd, dim_ffn, device='cuda')
Vw = torch.randn(dim_ffn, n_embd, device='cuda')
k1 = cm.rwkv_mix_fwd(x, x_prev, x_k)
k1_K = k1 @ Kw
k = cm.rwkv_relu_and_square_fwd(k1_K, inplace=False)
go = torch.randn(B, T, n_embd, device='cuda')
cm.rwkv_channel_mixing_bwd(go, x, x_prev, x_k, Kw, Vw, k1, k1_K, k, inplace=True)
print('site 3  mix_bwd dx_prev  cells written past dx_prev:', (big[B * n_embd:] != SENT).sum().item())
```

```
site 1  pow_and_relu     cells written past the end: 3292
site 2  relu_square_bwd  cells written past dk: 3292
site 3  mix_bwd dx_prev  cells written past dx_prev: 8
```

Driving the public `channel_mixing_rwkv7` forward+backward at the same shape under `compute-sanitizer --tool memcheck` with `PYTORCH_NO_CUDA_MEMORY_CACHING=1` attributes every invalid access to the three sites:

```
========= Invalid __global__ read of size 16 bytes
=========     at rwkv_channel_mixing_pow_and_relu+0x110 in fla/ops/rwkv7/channel_mixing.py:91
=========     and is 1 bytes after the nearest allocation at 0x7fa6983c5e00 of size 19,600 bytes
...
error counts by source line, one public forward+backward:
  channel_mixing.py:91    142 invalid reads    (pow_and_relu load)
  channel_mixing.py:94     96 invalid writes   (pow_and_relu store)
  channel_mixing.py:184     23 invalid reads   (relu_square_bwd load)
  channel_mixing.py:185     96 invalid reads   (relu_square_bwd load)
  channel_mixing.py:240  3,248 invalid writes  (mix_bwd dx_prev store)
  ERROR SUMMARY: 3606 errors        -> 0 device errors with the masks in place
```

### Expected Behavior

No access outside the tensors: tail lanes masked, as this file's own `rwkv_seq_mix_kernel` already does on every load and store. PR with the fix and a regression test: #PR_NUMBER.

### Environment Information

1. Torch: 2.13.0+cu129
2. Triton: 3.7.1
3. Commit tested: 2501ac83 (current main), Python 3.12, NVIDIA L40S, driver CUDA 12.4

truong-v:
## Summary

Three kernels in `fla/ops/rwkv7/channel_mixing.py` launch over grids that round up to the block size but leave the tail lanes unmasked, so any shape whose element count is not a multiple of the block size reads and writes past the end of the tensors involved — silently, through the public `channel_mixing_rwkv7` forward and backward. `rwkv_channel_mixing_pow_and_relu` and `relu_square_bwd_kernel` had no bounds concept at all (no element-count argument, no masks, `BLOCK_SIZE` hardcoded to 4096); `rwkv_mix_bwd_kenel` computes `is_valid` and applies it on three of its four accesses but stores `dx_prev` under `is_first_step` alone, so dead tail lanes whose `seq_idx` wraps to 0 store past the end of `dx_prev` — the caller's own `x_prev` when `inplace=True`. The corruption lands on whatever the allocator placed after those tensors; in our runs a first (autotune-sweep) `backward()` corrupted previously computed gradients of an unrelated reference path (max abs grad error 6e+01, vs 8e-06 once masked). Common RWKV7 configurations hide the bug because power-of-two `n_embd`/`dim_ffn` make every element count block-aligned.

Fixes #1102.

The fix follows the pattern this file's `rwkv_seq_mix_kernel` already uses:

- `rwkv_channel_mixing_pow_and_relu` and `relu_square_bwd_kernel` gain an `n_elements` argument and mask every load and store; their two launch sites (same file) pass `output.numel()` / `dk.numel()`. Neither kernel is exported or called anywhere else in the repo.
- `rwkv_mix_bwd_kenel`'s `dx_prev` store mask becomes `is_first_step & is_valid`, matching the three accesses above it.

In-bounds results are unchanged: masked lanes never contributed to any in-bounds output.

## Test plan

- `python scripts/find_dependent_tests.py fla/ops/rwkv7/channel_mixing.py` → `tests/ops/test_rwkv7.py`; full file passes on an L40S: 38 passed, 4 skipped (the pre-existing Hopper-only T=131072 skips on this sm_89 card).
- New `test_channel_mixing_no_out_of_bounds_writes`: sentinel guard bands around the storage each site writes through, at element counts (1, 7, 200, 700) that no block size in [128, 8192] divides — a view fed to `rwkv_relu_and_square_fwd(inplace=True)`, the exact `relu_square_bwd_kernel` launch `rwkv_channel_mixing_bwd` performs, and a full `rwkv_channel_mixing_bwd(inplace=True)` whose `dx_prev` is a view of a larger buffer. The existing suite could not see the bug: its shapes are block-aligned, and in-bounds outputs are correct anyway.
- Reverting the kernel changes makes the new test fail at the first guard-band assert; reverting only the one-line `dx_prev` mask makes it fail at that site's assert (`rwkv_mix_bwd_kenel stored past the end of dx_prev`).
- Under `compute-sanitizer --tool memcheck` with `PYTORCH_NO_CUDA_MEMORY_CACHING=1`, the public forward+backward goes from 3,606 invalid global reads/writes attributed to `channel_mixing.py:91/94/184/185/240` to zero device errors.
- The new test's shapes force a fresh `torch.compile` of this file's `compute_x_k_grad` helper, which exposed a latent `tests/conftest.py` issue on the PyTorch 2.12 CI runner: inductor's one-time pattern-matcher init allocates `requires_grad=True` example tensors through the patched `torch.empty`, and `fill_(nan)` raises on a leaf tensor that requires grad. The conftest now skips poisoning for allocations that require grad — kernel scratch buffers never do, so no coverage is lost.

## Benchmark / NCU

do_bench per kernel on an NVIDIA L40S at (B, T, n_embd, dim_ffn) = (2, 4096, 1024, 4096), bf16 — element counts block-aligned so both versions do identical work (no tail): the mix numel is 8Mi, the FFN numel 32Mi.

| kernel | before | after |
|---|---|---|
| `rwkv_channel_mixing_pow_and_relu` | 210.3–211.3 us | 211.2–211.5 us |
| `relu_square_bwd_kernel` | 310.5–310.6 us | 311.0–311.5 us |
| `rwkv_mix_bwd_kenel` | 54.1 us | 54.7–55.1 us |

(three interleaved before/after runs each; `triton.testing.do_bench`, kernels driven exactly as their wrappers launch them)

Conclusion: neutral for the two fully-unmasked kernels (≤ 0.3%, within run-to-run spread); `rwkv_mix_bwd_kenel` pays up to ~1 us (~1.5%) at this shape for the corrected store predicate. Varlen: not applicable, these kernels have no varlen path.

## Breaking changes

None. In-bounds outputs are unchanged. The two previously unmasked kernels gain an `n_elements` parameter; they are internal launch targets of their same-file wrappers and are not exported from `fla.ops.rwkv7`.

## Checklist

## Environment

- Commit base: 2501ac83 (main)
- GPU: NVIDIA L40S (sm_89), driver CUDA 12.4
- torch 2.13.0+cu129, triton 3.7.1, Python 3.12

### DIFF
--- a/fla/ops/rwkv7/channel_mixing.py
+++ b/fla/ops/rwkv7/channel_mixing.py
@@ -82,16 +82,18 @@ def rwkv_seq_mix_kernel(
 def rwkv_channel_mixing_pow_and_relu(
     in_ptr,
     out_ptr,
+    n_elements,
     BLOCK_SIZE: tl.constexpr,
 ):
     """Fused ReLU and Power operation: x = ReLU(x)^2"""
     xoffset = tl.program_id(0).to(tl.int64) * BLOCK_SIZE
     xindex = xoffset + tl.arange(0, BLOCK_SIZE)
     x0 = xindex
-    x = tl.load(in_ptr + (x0), None)
+    xmask = xindex < n_elements
+    x = tl.load(in_ptr + (x0), mask=xmask, other=0.0)
     x = tl.maximum(x, 0.0).to(tl.float32)
     x = tl.cast(x * x, dtype=out_ptr.dtype.element_ty, fp_downcast_rounding='rtne')
-    tl.store(out_ptr + (x0), x, None)
+    tl.store(out_ptr + (x0), x, mask=xmask)
 
 
 def rwkv_mix_torch(x: torch.Tensor, x_prev: torch.Tensor, x_k: torch.Tensor):
@@ -162,6 +164,7 @@ def rwkv_relu_and_square_fwd(x: torch.Tensor, inplace: bool = True):
     rwkv_channel_mixing_pow_and_relu[grid](
         x,
         output,
+        output.numel(),
         BLOCK_SIZE=4096,
     )
 
@@ -172,6 +175,7 @@ def rwkv_relu_and_square_fwd(x: torch.Tensor, inplace: bool = True):
 def relu_square_bwd_kernel(
     out_ptr,
     forward_input_ptr,
+    n_elements,
     BLOCK_SIZE: tl.constexpr,
 ):
     """ReLU(x)^2 backward kernel
@@ -180,15 +184,16 @@ def relu_square_bwd_kernel(
     pid = tl.program_id(0).to(tl.int64)
     block_start = pid * BLOCK_SIZE
     offsets = block_start + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < n_elements
 
-    x = tl.load(forward_input_ptr + offsets).to(tl.float32)
-    grad = tl.load(out_ptr + offsets).to(tl.float32)
+    x = tl.load(forward_input_ptr + offsets, mask=mask, other=0.0).to(tl.float32)
+    grad = tl.load(out_ptr + offsets, mask=mask, other=0.0).to(tl.float32)
 
     x = tl.maximum(x, 0.0)
 
     grad_input = grad * 2 * x
 
-    tl.store(out_ptr + offsets, grad_input.to(out_ptr.dtype.element_ty))
+    tl.store(out_ptr + offsets, grad_input.to(out_ptr.dtype.element_ty), mask=mask)
 
 
 @triton.autotune(
@@ -238,7 +243,7 @@ def rwkv_mix_bwd_kenel(
     tl.store(
         dx_prev_ptr + dx_prev_offset,
         tl.cast(prod, dtype=dx_prev_ptr.dtype.element_ty),
-        mask=is_first_step,
+        mask=is_first_step & is_valid,
     )
 
 
@@ -272,6 +277,7 @@ def rwkv_channel_mixing_bwd(grad_output, x, x_prev, x_k, key_weight, value_weigh
     relu_square_bwd_kernel[grid](
         dk,
         k1_K,
+        dk.numel(),
         BLOCK_SIZE=BLOCK_SIZE,
     )
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -73,7 +73,9 @@ def _guarded_empty(*args, **kwargs):
     if not (dtype.is_floating_point or dtype.is_complex):
         return _ORIGINAL_EMPTY(*args, **kwargs)
 
-    if is_compiling() or not _is_called_from_fla():
+    # fill_ raises on leaf tensors that require grad; such allocations (e.g. inductor's
+    # pattern-matcher example inputs) are never kernel scratch buffers, so leave them unpoisoned
+    if is_compiling() or kwargs.get('requires_grad', False) or not _is_called_from_fla():
         return _ORIGINAL_EMPTY(*args, **kwargs)
 
     result = _ORIGINAL_EMPTY(*args, **kwargs)
@@ -88,7 +90,7 @@ def _guarded_empty(*args, **kwargs):
 
 def _guarded_empty_like(input, **kwargs):
     """Create a tensor filled with NaN instead of uninitialized values."""
-    if is_compiling() or not _is_called_from_fla():
+    if is_compiling() or kwargs.get('requires_grad', False) or not _is_called_from_fla():
         return _ORIGINAL_EMPTY_LIKE(input, **kwargs)
 
     if kwargs.get('dtype') is None:
@@ -110,7 +112,7 @@ def _guarded_empty_like(input, **kwargs):
 
 def _guarded_new_empty(self, *args, **kwargs):
     """Create a tensor filled with NaN instead of uninitialized values."""
-    if is_compiling() or not _is_called_from_fla():
+    if is_compiling() or kwargs.get('requires_grad', False) or not _is_called_from_fla():
         return _ORIGINAL_NEW_EMPTY(self, *args, **kwargs)
 
     if kwargs.get('dtype') is None:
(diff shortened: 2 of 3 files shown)


zhiyuan1i (approved):
The three kernel fixes look correct, and the guard-band tests do reproduce the original out-of-bounds writes — and they actually run in CI. The `dx_prev` case is the nasty one: with `inplace=True` that store lands directly in the caller's buffer.

One non-blocking ask: the sentinel tests prove nothing is written out of bounds, but they would not catch an off-by-one predicate that masks out the last valid element, since the existing numerical comparison only uses block-aligned shapes. Could you add one small non-block-aligned case through `channel_mixing_rwkv7` comparing the forward output and gradients against the torch reference? Fine as a follow-up.

Also, the new test's `SKIP_TEST_CHUNK_VARLEN` decorator does not seem related to this test (and its reason names a different variable) — worth dropping or fixing.

truong-v:
Thanks for the review.

- Dropped the stray `SKIP_TEST_CHUNK_VARLEN` decorator from the sentinel test (pushed).
- The numerical non-block-aligned case I'll send as a follow-up PR as suggested. One finding from preparing it: adding such a case to `test_channel_mixing_gradients` at bf16 flakes, because the 5e-3 ratio tolerance has no headroom there — real bf16 quantization noise at small non-aligned shapes measures 0.0045–0.0059 on an L40S. The follow-up will instead add a dedicated fp32 test through `channel_mixing_rwkv7` at `(B, T, n_embd, dim_ffn) = (2, 7, 200, 700)` (element counts no block size in [128, 8192] divides), comparing the forward output and all five gradients against the torch reference at a 1e-3 ratio bound: measured error there is ~1e-6, and a mask dropping even a single valid tail element registers at ~2e-2. I have it validated locally, including mutation checks — making any of the three fixed masks drop one valid element/column fails all its cases.

PR #1105, opened 2026-08-05
[Fix] Handle chunk-boundary-straddling key tiles in DeltaFormer forward

### LINKED ISSUE #1104: [Bug] DeltaFormer forward: a key tile straddling the chunk boundary skips the causal mask and reads the launch's own uninitialized output rows
### Checklist

### Describe the Bug

DeltaFormer's forward walks the sequence in chunks of `C`, sharing one `u` buffer per sequence: at each launch, rows `[0, T-C)` hold earlier chunks' results and rows `[T-C, T)` are still uninitialized `torch.empty_like` memory that this launch is about to write. `parallel_deltaformer_fwd_kernel` decides **once per `BLOCK_T` tile, from the tile's start offset `kv_i` alone**, whether the tile is "earlier chunks" (read `u`, no causal mask) or "the current chunk" (causal mask, no `u` read):

https://github.com/fla-org/flash-linear-attention/blob/2501ac83fd2c1223f3f0393563f2d58255df9b5d/fla/ops/deltaformer/parallel.py#L168-L192

`BLOCK_T` is autotuned over {64, 32}. When it does not divide `C`, one tile straddles the `T - C` boundary and the whole tile takes the "earlier chunks" branch:

1. the causal mask is never applied, so the current chunk's own key columns enter `rowmax`/`rowsum` unmasked — `lse`, `w` and `u` all come out wrong; and
2. the `u` load is bounded only by `T`, so it pulls in rows `[T-C, T)` — the launch's own not-yet-written output region, i.e. uninitialized memory. The answer then also changes with whatever garbage the allocator left there, so results are not reproducible run to run.

`C` is a public, documented, unvalidated keyword of `deltaformer_attn` (and of `ParallelDeltaformerFunction.apply`). Any `C` not a multiple of 64 can fire; any `C` not a multiple of 32 fires at every autotune choice. Nobody noticed because the default `C = 512` and every tested shape keep the boundary 64-aligned. The backward kernel in the same file already handles the straddle correctly (`parallel_deltaformer_bwd_kernel_u` uses `kv_i + BLOCK_T >= ...` with an elementwise mask), so this is the forward missing the same treatment.

Compares `u` from `ParallelDeltaformerFunction.apply` (the stage `deltaformer_attn` wraps; flash-attn only consumes `u` afterwards) against the recurrence of `fla/ops/deltaformer/naive.py`, and runs each config twice with different values planted in freed allocator memory:

```python
import math
import torch
import fla.ops.deltaformer.parallel as dfp
from fla.ops.deltaformer.naive import tril_softmax

def naive_u(q, k, v, beta):
    # stage 1 of fla/ops/deltaformer/naive.py (which only returns o):
    # u[t] = v[t] - beta[t] * sum_{j<t} softmax(q[t] @ k[:t]^T)[j] * u[j]
    qf, kf, vf = (t.float().transpose(1, 2) for t in (q, k, v))
    betaf = beta.float().transpose(1, 2)
    probs = tril_softmax(qf @ kf.transpose(-1, -2) * (1.0 / math.sqrt(q.shape[-1])), strict=True)
    us = []
    for t in range(q.shape[1]):
        u_t = vf[:, :, t]
        if t > 0:
            u_prev = torch.stack(us, dim=-2)
            u_t = u_t - betaf[:, :, t, None] * (probs[:, :, t, :t, None] * u_prev).sum(-2)
        us.append(u_t)
    return torch.stack(us, dim=2).transpose(1, 2).contiguous()

def poison_allocator(value, numel=1 << 22):
    junk = torch.full((numel,), value, device='cuda')
    del junk
    torch.cuda.synchronize()

torch.manual_seed(42)
B, T, H, D = 1, 64, 2, 64
q, k, v = (torch.randn(B, T, H, D, device='cuda') for _ in range(3))
beta = torch.randn(B, T, H, device='cuda').sigmoid()

for C, label in [(16, 'C=16  (straddling: BLOCK_T in {64, 32} does not divide 16)'),
                 (32, 'C=32  (straddling when the autotuner picks BLOCK_T=64)'),
                 (64, 'C=64  (aligned control)')]:
    with torch.no_grad():
        poison_allocator(777.0)
        u1 = dfp.ParallelDeltaformerFunction.apply(q, k, v, beta, C, None)
        poison_allocator(-333.0)
        u2 = dfp.ParallelDeltaformerFunction.apply(q, k, v, beta, C, None)
    ref = naive_u(q, k, v, beta)
    print(f'{label}\n    max |u - naive_u| {(u1 - ref).abs().max().item():.6f}'
          f'    max |u_run1 - u_run2| across allocator poisons {(u1 - u2).abs().max().item():.6f}')
```

```
C=16  (straddling: BLOCK_T in {64, 32} does not divide 16)
    max |u - naive_u| 0.486892    max |u_run1 - u_run2| across allocator poisons 0.559144
C=32  (straddling when the autotuner picks BLOCK_T=64)
    max |u - naive_u| 0.000998    max |u_run1 - u_run2| across allocator poisons 0.000000
C=64  (aligned control)
    max |u - naive_u| 0.000873    max |u_run1 - u_run2| across allocator poisons 0.000000
```

`compute-sanitizer --tool initcheck` with `PYTORCH_NO_CUDA_MEMORY_CACHING=1` on the `C=16` call reports 1,024 uninitialized global reads, all at the `u` load:

```
========= Uninitialized __global__ memory read of size 16 bytes
=========     at parallel_deltaformer_fwd_kernel+0x1cd0 in fla/ops/deltaformer/parallel.py:191
...
========= ERROR SUMMARY: 1024 errors
```

### Expected Behavior

`u` matches the naive recurrence within fp32 noise for every chunk size (as the aligned control does), deterministically. PR with the fix and regression tests: #PR_NUMBER.

### Environment Information

1. Torch: 2.13.0+cu129
2. Triton: 3.7.1
3. Commit tested: 2501ac83 (current main), Python 3.12, NVIDIA L40S, driver CUDA 12.4

truong-v:
## Summary

`parallel_deltaformer_fwd_kernel` decides per `BLOCK_T` tile, from the tile's start offset alone, whether a key tile belongs to earlier chunks (read `u`, no causal mask) or to the current chunk (causal mask, no `u` read). The chunk size `C` is a public, unvalidated keyword of `deltaformer_attn`; when it is not a multiple of the autotuned `BLOCK_T` (64 or 32), a prefix/current-chunk boundary `T - C` can fall strictly inside one key tile. That straddling tile is treated wholly as "earlier chunks": the current chunk's columns enter the softmax statistics unmasked, and the `u` load sweeps into rows `[T-C, T)`, the launch's own not-yet-written output region (uninitialized `torch.empty_like` memory). The result is wrong (`max |u - naive| = 0.487` at `C=16, T=64` vs `0.0009` for the aligned control) and nondeterministic (changes by up to 0.56 with the garbage the allocator left behind). The bug needs an explicitly non-aligned `C`: the default `C = 512` is a multiple of both `BLOCK_T` values, so every boundary lands on a tile edge — the shipped layer and the existing test shapes never hit it.

Fixes #1104.

Two changes in the kernel, following the treatment `parallel_deltaformer_bwd_kernel_u` in the same file already uses for its own boundary (`kv_i + BLOCK_T >= ...` with an elementwise mask):

- The causal-mask branch now also covers straddling tiles: `if kv_i >= T - C` becomes `if kv_i + BLOCK_T > T - C`. The mask expression is unchanged — it is already elementwise in global column indices, so prefix columns of a straddling tile evaluate to false and keep their scores.
- The `u` load is bounded by the boundary instead of the tensor end: mask `o_kv < T - C` instead of `o_kv < T`, so a straddling tile reads only initialized prefix rows (masked lanes load 0.0 and contribute nothing, matching the aligned behavior where current-chunk tiles skip the `u` dot entirely).

Aligned chunk sizes compile to the same tile behavior as before; open PR #603 restructures this loop for performance but does not address the straddling-tile boundary handling.

## Test plan

- `python scripts/find_dependent_tests.py fla/ops/deltaformer/parallel.py` → `tests/ops/test_deltaformer.py`.
- New `test_parallel_deltaformer_u` tests the stage this PR touches — `ParallelDeltaformerFunction` — directly against the naive `u` recurrence, forward output and all four gradients, at `(B, T, H, D) = (2, 192, 2, 64)` with straddling chunk sizes `C = 16` and `C = 48` plus an aligned control `C = 64`. It needs no flash-attn, so it runs on the CI GPU runners where the rest of this file is skipped; the flash-attn skip moved from module scope onto the two `deltaformer_attn`-level tests it actually applies to.
- `test_deltaformer_attn` keeps the new `C` axis with chunk sizes the `BLOCK_T` values do not divide, now at lengths the chunk size divides — `(B, T, C) = (2, 192, 48)` and `(1, 384, 96)` — so these cases stay clear of an unrelated pre-existing backward issue when `C` does not divide `T` (forward chunks from 0, backward walks back from `T - C`, and `ws` is indexed with mismatched offsets).
- The straddling cases fail deterministically without the fix: the NaN memory poisoning in `tests/conftest.py` fills the not-yet-written `u` rows the unfixed kernel reads. Verified by reverting the kernel change, on both an L40S (sm_89) and an H100 (sm_90): `C = 16` and `C = 48` fail, the `C = 64` control passes. With the fix, on both GPUs: 3 passed, 10 skipped (no local flash-attn build; the CI runners exercise the full `deltaformer_attn` path), worst error ratio 5.3e-4 against bounds of 6e-3/8e-3.
- Without the fix, at `C=16, T=64` fp32: `max |u - naive| = 0.48`, gradient errors through `ParallelDeltaformerFunction` up to `1.02` (`dbeta`); with the fix: `0.001`–`0.013`, identical to the aligned control's noise floor.
- `compute-sanitizer --tool initcheck` on the public `C=16` path: 1,024 uninitialized global reads at `parallel.py:191` before, 0 errors after.

## Benchmark / NCU

`triton.testing.do_bench` on an NVIDIA L40S, fp16, `(B, T, H, D) = (1, 4096, 8, 128)` at the default `C = 512` (block-aligned, so both versions do identical work); three interleaved before/after runs:

| workload | before | after |
|---|---|---|
| full `u` computation (8 chunks) | 4.813–4.840 ms | 4.791–4.822 ms |
| last-chunk forward kernel alone | 161.4–164.8 us | 161.7–163.5 us |

Conclusion: neutral — differences are within run-to-run spread. Straddling chunk sizes previously computed a wrong answer, so there is no meaningful "before" to compare there.

## Breaking changes

None. Aligned chunk sizes (including the default) are unchanged; straddling chunk sizes go from silently wrong and nondeterministic to correct.

## Checklist

## Environment

- Commit base: 2501ac83 (main)
- GPU: NVIDIA L40S (sm_89), driver CUDA 12.4
- torch 2.13.0+cu129, triton 3.7.1, Python 3.12

### DIFF
--- a/fla/ops/deltaformer/parallel.py
+++ b/fla/ops/deltaformer/parallel.py
@@ -172,7 +172,8 @@ def parallel_deltaformer_fwd_kernel(
         k = tl.load(k_blk_ptr, mask=m_kv[None, :], other=0.0)
         qk = tl.dot(q, k) * qk_scale
 
-        if kv_i >= T - C:
+        # tiles straddling the T - C boundary need the mask too; prefix columns evaluate to false
+        if kv_i + BLOCK_T > T - C:
             mask = (T - C - kv_i + rowid_block[:, None] - colid_block[None, :] < 1)
             qk = tl.where(mask, -1e6, qk)
 
@@ -188,7 +189,9 @@ def parallel_deltaformer_fwd_kernel(
 
         if kv_i < T - C:
             u_blk_ptr = u_ptr + pid_h * D + o_kv[:, None] * (H * D) + o_d[None, :]
-            u = tl.load(u_blk_ptr, mask=m_kv[:, None], other=0.0)
+            # rows at and past T - C are this launch's own output, not yet written
+            m_u = o_kv < T - C
+            u = tl.load(u_blk_ptr, mask=m_u[:, None], other=0.0)
             acc = tl.dot(p.to(u_ptr.dtype.element_ty), u, acc)
 
     lse = rowmax + tl.math.log2(rowsum)
--- a/tests/ops/test_deltaformer.py
+++ b/tests/ops/test_deltaformer.py
@@ -23,14 +23,18 @@ pytestmark = [
 
 
 @pytest.mark.parametrize(
-    ('B', 'T', 'H', 'D', 'dtype'),
+    ('B', 'T', 'H', 'D', 'C', 'dtype'),
     [
-        pytest.param(*test, id="B{}-T{}-H{}-D{}-{}".format(*test))
+        pytest.param(*test, id="B{}-T{}-H{}-D{}-C{}-{}".format(*test))
         for test in [
-            (2, 128, 2, 64, torch.float16),
-            (1, 256, 4, 64, torch.float16),
-            (2, 512, 4, 64, torch.float16),
-            (4, 1024, 4, 128, torch.float16),
+            (2, 128, 2, 64, 512, torch.float16),
+            (1, 256, 4, 64, 512, torch.float16),
+            (2, 512, 4, 64, 512, torch.float16),
+            (4, 1024, 4, 128, 512, torch.float16),
+            # chunk sizes the autotuned BLOCK_T values (64, 32) do not divide, so a key
+            # tile straddles the chunk boundary
+            (2, 200, 2, 64, 48, torch.float16),
+            (1, 400, 2, 64, 96, torch.float16),
         ]
     ],
 )
@@ -43,6 +47,7 @@ def test_deltaformer_attn(
     T: int,
     H: int,
     D: int,
+    C: int,
     dtype: torch.dtype,
 ):
     """
@@ -64,7 +69,7 @@ def test_deltaformer_attn(
     ref_dv, v.grad = v.grad.clone(), None
     ref_dbeta, beta.grad = beta.grad.clone(), None
 
-    tri = deltaformer_attn(q, k, v, beta)
+    tri = deltaformer_attn(q, k, v, beta, C=C)
     tri.backward(do)
     tri_dq, q.grad = q.grad.clone(), None
     tri_dk, k.grad = k.grad.clone(), None
--- a/tests/ops/test_deltaformer.py
+++ b/tests/ops/test_deltaformer.py
@@ -5,21 +5,24 @@
 # For a list of all contributors, visit:
 #   https://github.com/fla-org/flash-linear-attention/graphs/contributors
 
+import math
+
 import pytest
 import torch
 
 from fla.ops.deltaformer import deltaformer_attn
-from fla.ops.deltaformer.naive import naive_deltaformer_attn
+from fla.ops.deltaformer.naive import naive_deltaformer_attn, tril_softmax
+from fla.ops.deltaformer.parallel import ParallelDeltaformerFunction
 from fla.utils import IS_INTEL_ALCHEMIST, assert_close, device, find_spec_cached
 
-pytestmark = [
-    # Module-level importorskip would exit pytest with code 5 (no tests collected), which
-    # breaks CI's per-file `pytest "$f" || exit 1` loop; mark tests as skipped instead.
-    pytest.mark.skipif(
-        find_spec_cached("flash_attn") is None,
-        reason="deltaformer_attn requires flash-attn (`pip install flash-attn --no-build-isolation`).",
-    ),
-]
+# Only the public deltaformer_attn needs flash-attn (to consume u); the u computation itself
+# does not, so the mark sits on individual tests rather than the module. A module-level
+# importorskip would also exit pytest with code 5 (no tests collected), which breaks CI's
+# per-file `pytest "$f" || exit 1` loop.
+requires_flash_attn = pytest.mark.skipif(
+    find_spec_cached("flash_attn") is None,
+    reason="deltaformer_attn requires flash-attn (`pip install flash-attn --no-build-isolation`).",
+)
 
 
 @pytest.mark.parametrize(
@@ -33,11 +36,12 @@ pytestmark = [
             (4, 1024, 4, 128, 512, torch.float16),
             # chunk sizes the autotuned BLOCK_T values (64, 32) do not divide, so a key
             # tile straddles the chunk boundary
-            (2, 200, 2, 64, 48, torch.float16),
-            (1, 400, 2, 64, 96, torch.float16),
+            (2, 192, 2, 64, 48, torch.float16),
+            (1, 384, 2, 64, 96, torch.float16),
         ]
     ],
 )
+@requires_flash_attn
 @pytest.mark.skipif(
     IS_INTEL_ALCHEMIST,
     reason="Skipping test on Intel Alchemist due to known issues with SRAM.",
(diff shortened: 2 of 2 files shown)


zhiyuan1i (changes requested):
The kernel fix itself looks correct: moving the straddling tile into the causal branch and masking the `u` load with `o_kv < T - C` fixes both the softmax statistics and the uninitialized reads, and the aligned default path is unaffected. Requesting changes on the regression test only.

First, the H100 test-ops check does not actually exercise these cases: all 10 tests in test_deltaformer.py are skipped there because flash_attn is unavailable, so the green check is currently vacuous for this PR. Please make at least one regression case actually run in CI — e.g. a focused test at the `ParallelDeltaformerFunction` level that does not depend on flash_attn.

Second, the new shapes mix in an unrelated pre-existing backward bug. Fixed-length forward chunks from zero (`range(0, T, C)`), while backward starts at `T - C` and walks backwards; with `(T, C) = (200, 48)` the backward chunk boundaries do not coincide with the forward ones and `ws` is indexed with mismatched offsets. Could you use divisible lengths instead — `(192, 48)` and `(384, 96)`? `C=48` still guarantees a straddling tile for both `BLOCK_T=32` and `64`.

One precision on the trigger conditions, for the description: this only fires when a prefix/current boundary `T - C` falls strictly inside a `BLOCK_T` tile, i.e. for an explicitly non-aligned `C`; the default `C=512` path never straddles.

truong-v:
Thanks for the review — all three points addressed.

- CI coverage: added `test_parallel_deltaformer_u`, which tests the stage this PR touches — `ParallelDeltaformerFunction` — directly against the naive `u` recurrence (forward output and all four gradients) with no flash-attn dependency, at straddling `C = 16` and `C = 48` plus an aligned `C = 64` control. The flash-attn skip moved from module scope onto the two `deltaformer_attn`-level tests it actually applies to, so the new test collects and runs on the H100 runner. Revert-checked on both an L40S and an H100: without the kernel fix the two straddling cases fail deterministically (the conftest NaN poisoning fills the not-yet-written `u` rows the unfixed kernel reads) and the control passes.
- Shapes: switched to `(192, 48)` and `(384, 96)` as suggested — thanks for the pointer on the backward chunk-offset mismatch with non-divisible `T`.
- Description: tightened the trigger wording — the bug fires only when a prefix/current-chunk boundary `T - C` falls strictly inside a `BLOCK_T` tile, i.e. for an explicitly non-aligned `C`; the default `C = 512` path never straddles.

zhiyuan1i (approved):

PR #1113, opened 2026-08-08
[Fix] Ascend NPU: guard against tl.dot left-operand clobber across triton_ascend kernels

sunyi0505:
## Summary

On Ascend NPU, tl.dot(lhs, rhs, ...) may overwrite the left‑hand operand in‑place (within UB), which differs from CUDA Triton behavior. Reusing the left‑hand operand tile across successive tl.dot calls within the same kernel can cause silent numerical divergence (mismatch against Torch oracle with no compilation errors).
This PR unifies the fix for this issue across multiple triton_ascend kernels and adds Ascend performance skill documentation.
Root cause
The Ascend backend tl.dot clobbers the left‑hand operand. Typical scenarios:
fwd wy_fast: In the u‑loop, b_u = tl.dot(b_A, ...) corrupts b_A, yet the w‑loop still requires the original A
bwd wy_fast: b_A gets clobbered in the k‑loop and is reused in the v‑loop
inter‑chunk solve / chunk_o / chunk_delta_h: One tile participates in multiple dot operations (e.g., b_q is used for both b_o and b_A)
Changes
Two remediation strategies:
Reload from GM: Reload the original tile before subsequent dot operations (wy_fast fwd/bwd, chunk_bwd dAv)
0.0 Keep pristine copy: Make a copy prior to dot to prevent lhs overwriting (chunk_o, chunk_delta_h, solve_tril, chunk_intra)

## Test plan

FLA_NPU_XDIST=1 ASCEND_RT_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 pytest --exitfirst -n 8 --dist load tests/modules     tests/ops/utils     tests/ops/test_gdn_kernels.py     tests/ops/test_gdn.py     tests/ops/test_kda.py     tests/ops/test_attnres.py     tests/ops/test_solve_tril.py

hardware: Atlas 800T A3(X86)
568417cb2ffb088f682ce03427b80d4d

None

## Breaking changes

None

## Checklist

### DIFF
--- a/fla/ops/common/backends/triton_ascend/chunk_o.py
+++ b/fla/ops/common/backends/triton_ascend/chunk_o.py
@@ -201,8 +201,10 @@ def chunk_fwd_kernel_o_npu(
                 b_o += tl.dot(b_q, tl.trans(b_h))
             else:
                 b_o += tl.dot(b_q, b_h)
+            # Ascend tl.dot clobbers lhs; keep a pristine copy for the A dot.
+            b_q_c = b_q + 0.0
             # [BT, BK] @ [BK, BT] -> [BT, BT]
-            b_A += tl.dot(b_q, b_k)
+            b_A += tl.dot(b_q_c, b_k)
 
         if USE_G:
             # g is transposed to [B, HV, T] in wrapper for contiguous T-load.
@@ -413,12 +415,14 @@ def chunk_bwd_kernel_dv_local_hv1_npu(
     m_t = o_t < T
     m_A = (o_t[:, None] <= o_t[None, :]) & (m_t[:, None] & m_t)
     b_A = tl.where(m_A, b_A, 0)
+    b_A_pristine = b_A + 0.0
 
     for i_v in range(tl.cdiv(V, BV)):
         p_do = tl.make_block_ptr(do, (T, V), (HV * V, 1), (i_t * BT, i_v * BV), (BT, BV), (1, 0))
         p_dv = tl.make_block_ptr(dv, (T, V), (HV * V, 1), (i_t * BT, i_v * BV), (BT, BV), (1, 0))
         b_do = tl.load(p_do, boundary_check=(0, 1))
-        b_dv = tl.dot(b_A.to(b_do.dtype), b_do, allow_tf32=False)
+        b_A_i = b_A_pristine + 0.0
+        b_dv = tl.dot(b_A_i.to(b_do.dtype), b_do, allow_tf32=False)
         tl.store(p_dv, b_dv.to(p_dv.dtype.element_ty), boundary_check=(0, 1))
 
 
@@ -745,11 +749,12 @@ def chunk_bwd_kernel_dqkwg_npu(
             p_k_c = tl.make_block_ptr(k, (T, K), (H * K, 1), (i_tc_c, i_k * BK), (BC, BK), (1, 0))
             b_k_c = tl.load(p_k_c, boundary_check=(0, 1))
             b_ds = tl.where(m_blk, b_ds, 0).to(b_k_c.dtype)
+            b_ds_c = b_ds + 0.0
             b_dq_r += tl.dot(b_ds, b_k_c, allow_tf32=False)
 
             p_dk_acc = tl.make_block_ptr(dk_f32, (T, K), (HV * K, 1), (i_tc_c, i_k * BK), (BC, BK), (1, 0))
             b_dk_acc = tl.load(p_dk_acc, boundary_check=(0, 1))
-            b_ds_dk = tl.dot(tl.trans(b_ds), b_q_r, allow_tf32=False)
+            b_ds_dk = tl.dot(tl.trans(b_ds_c), b_q_r, allow_tf32=False)
             if not USE_G and not USE_G_GAMMA:
                 b_ds_dk = b_ds_dk * scale
             b_dk_acc += b_ds_dk
--- a/fla/ops/common/backends/triton_ascend/chunk_o.py
+++ b/fla/ops/common/backends/triton_ascend/chunk_o.py
@@ -196,13 +196,15 @@ def chunk_fwd_kernel_o_npu(
             # [BK, BV]
             b_h = tl.load(p_h, boundary_check=(0, 1))
 
+            # Ascend tl.dot clobbers lhs; copy before the first dot on b_q.
+            b_q_c = b_q + 0.0
             # [BT, BK] @ [BK, BV] -> [BT, BV]
             if STATE_V_FIRST:
                 b_o += tl.dot(b_q, tl.trans(b_h))
             else:
                 b_o += tl.dot(b_q, b_h)
             # [BT, BK] @ [BK, BT] -> [BT, BT]
-            b_A += tl.dot(b_q, b_k)
+            b_A += tl.dot(b_q_c, b_k)
 
         if USE_G:
             # g is transposed to [B, HV, T] in wrapper for contiguous T-load.
@@ -413,12 +415,14 @@ def chunk_bwd_kernel_dv_local_hv1_npu(
     m_t = o_t < T
     m_A = (o_t[:, None] <= o_t[None, :]) & (m_t[:, None] & m_t)
     b_A = tl.where(m_A, b_A, 0)
+    b_A_pristine = b_A + 0.0
 
     for i_v in range(tl.cdiv(V, BV)):
         p_do = tl.make_block_ptr(do, (T, V), (HV * V, 1), (i_t * BT, i_v * BV), (BT, BV), (1, 0))
         p_dv = tl.make_block_ptr(dv, (T, V), (HV * V, 1), (i_t * BT, i_v * BV), (BT, BV), (1, 0))
         b_do = tl.load(p_do, boundary_check=(0, 1))
-        b_dv = tl.dot(b_A.to(b_do.dtype), b_do, allow_tf32=False)
+        b_A_i = b_A_pristine + 0.0
+        b_dv = tl.dot(b_A_i.to(b_do.dtype), b_do, allow_tf32=False)
         tl.store(p_dv, b_dv.to(p_dv.dtype.element_ty), boundary_check=(0, 1))
 
 
@@ -507,8 +511,9 @@ def chunk_bwd_kernel_dv_local_npu(
                 p_q1 = tl.make_block_ptr(q, (K, T), (1, H * K), (i_k * BK, i_tc1), (BK, BC), (0, 1))
                 b_q0 = tl.load(p_q0, boundary_check=(0, 1))
                 b_q1 = tl.load(p_q1, boundary_check=(0, 1))
+                b_k0_c = b_k0 + 0.0
                 b_A00 += tl.dot(b_k0, b_q0, allow_tf32=False) * scale
-                b_A01 += tl.dot(b_k0, b_q1, allow_tf32=False) * scale
+                b_A01 += tl.dot(b_k0_c, b_q1, allow_tf32=False) * scale
                 b_A11 += tl.dot(b_k1, b_q1, allow_tf32=False) * scale
 
             if USE_G or USE_G_GAMMA:
@@ -745,11 +750,12 @@ def chunk_bwd_kernel_dqkwg_npu(
             p_k_c = tl.make_block_ptr(k, (T, K), (H * K, 1), (i_tc_c, i_k * BK), (BC, BK), (1, 0))
             b_k_c = tl.load(p_k_c, boundary_check=(0, 1))
             b_ds = tl.where(m_blk, b_ds, 0).to(b_k_c.dtype)
+            b_ds_c = b_ds + 0.0
             b_dq_r += tl.dot(b_ds, b_k_c, allow_tf32=False)
 
             p_dk_acc = tl.make_block_ptr(dk_f32, (T, K), (HV * K, 1), (i_tc_c, i_k * BK), (BC, BK), (1, 0))
             b_dk_acc = tl.load(p_dk_acc, boundary_check=(0, 1))
-            b_ds_dk = tl.dot(tl.trans(b_ds), b_q_r, allow_tf32=False)
+            b_ds_dk = tl.dot(tl.trans(b_ds_c), b_q_r, allow_tf32=False)
             if not USE_G and not USE_G_GAMMA:
                 b_ds_dk = b_ds_dk * scale
             b_dk_acc += b_ds_dk
--- a/fla/ops/gated_delta_rule/backends/triton_ascend/wy_fast.py
+++ b/fla/ops/gated_delta_rule/backends/triton_ascend/wy_fast.py
@@ -178,7 +178,6 @@ def recompute_w_u_fwd_kernel_npu(
         offs_bt = tl.arange(0, BT)[None, :]
         ptr_A = A + (bos * HV + i_h) * BT + offs_t_2d * (HV * BT) + offs_bt * 1
         mask_A = mask_t[:, None]
-        b_A = tl.load(ptr_A, mask=mask_A, other=0.0).to(tl.float32)
 
         ptr_beta = beta + bos_bh + i_h * T_max + global_offs_t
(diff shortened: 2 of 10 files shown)


sunyi0505:

fla/ops/common/backends/triton_ascend/chunk_o.py
```diff
@@ -201,8 +201,10 @@ def chunk_fwd_kernel_o_npu(
                 b_o += tl.dot(b_q, tl.trans(b_h))
             else:
                 b_o += tl.dot(b_q, b_h)
+            # Ascend tl.dot clobbers lhs; keep a pristine copy for the A dot.
+            b_q_c = b_q + 0.0
```
zhiyuan1i: Could you double-check the ordering here? `b_q_c` is created only after the first `tl.dot(b_q, ...)`. Under the clobber semantics described in the PR, that dot may already have overwritten `b_q` in UB, so the copy could hold the clobbered result instead of the original Q tile. If so, the copy needs to move before the first dot (or Q reloaded from GM before the second). If Triton-Ascend in fact materializes `b_q + 0.0` from a pre-dot snapshot, it would be great to understand why — the same question applies to the other `+ 0.0` copies below.

fla/ops/kda/backends/triton_ascend/chunk_intra.py
```diff
@@ -383,8 +383,10 @@ def chunk_kda_fwd_kernel_inter_solve_fused_npu(
             b_Aqk20 += tl.dot(b_qg2, b_kgt, allow_tf32=False)
             b_Akk20 += tl.dot(b_kg2, b_kgt, allow_tf32=False)
             b_kgt = tl.trans(b_k1 * exp2(b_gn2[None, :] - b_g1))
-            b_Aqk21 += tl.dot(b_qg2, b_kgt, allow_tf32=False)
-            b_Akk21 += tl.dot(b_kg2, b_kgt, allow_tf32=False)
+            b_qg2_c = b_qg2 + 0.0
+            b_kg2_c = b_kg2 + 0.0
+            b_Aqk21 += tl.dot(b_qg2_c, b_kgt, allow_tf32=False)
```
zhiyuan1i, line 388: Same ordering question for `b_qg2`/`b_kg2` (and the `c1`/`c2` copies for `b_qg3`/`b_kg3` below): these copies are made after the first lhs dot. Can you confirm they still hold the original tiles under the clobber semantics? If not, the pristine source has to be established before the first dot, with disposable lhs copies derived from it.

fla/ops/kda/backends/triton_ascend/chunk_intra.py
```diff
@@ -457,29 +469,30 @@ def chunk_kda_fwd_kernel_inter_solve_fused_npu(
     if NC >= 3:
         b_Ai21 = -tl.dot(
             tl.dot(b_Ai22, b_Akk21, allow_tf32=False),
-            b_Ai11,
+            b_Ai11_c,
             allow_tf32=False,
         )
         b_Ai20 = -tl.dot(
```
zhiyuan1i, line 481: Please confirm the lifetimes in this solve sequence: `b_Ai22_c` serves as a dot lhs (for `b_Ai20`) and is later read again as rhs (for `b_Ai32`); `b_Ai33_c2` is copied from `b_Ai33_c` after `b_Ai33_c` was already consumed as an lhs, and `b_Ai33_c` itself is what gets stored at the end. Under lhs clobbering, would any of these later reads see corrupted data?

fla/ops/utils/backends/triton_ascend/solve_tril.py
```diff
@@ -243,23 +248,24 @@ def merge_16x16_to_64x64_inverse_kernel_npu(
     )
     b_Ai_43 = -tl.dot(
```
zhiyuan1i, line 254: Same theme in the 64x64 path: `b_Ai_22` is an lhs around :240 and read again around :246; `b_Ai_33_c`/`b_Ai_44_c` are consumed as lhs before their `_c2` copies are taken and are themselves stored at the end; `b_A_42`/`b_A_43` each appear as lhs twice. Could you confirm these are safe?

fla/ops/gated_delta_rule/backends/triton_ascend/wy_fast.py
```diff
@@ -302,6 +304,8 @@ def prepare_wy_repr_bwd_kv_npu(
                 b_kbg = b_k * b_b[:, None]
             b_dw = tl.load(p_dw, boundary_check=(0, 1)).to(tl.float32)
             b_dA += tl.dot(b_dw, tl.trans(b_kbg), allow_tf32=False)
```
zhiyuan1i, line 307: One spot this PR may have missed: `b_dw` is the lhs of the first dot and is then read as the rhs of the next one (`b_dkbg = tl.dot(b_A, b_dw, ...)`). Same for `b_du` around :332-335. Could you confirm whether the original operand survives being an lhs, or do these need a reload/copy as well?

fla/ops/kda/backends/triton_ascend/chunk_bwd.py
```diff
@@ -106,7 +104,9 @@ def chunk_kda_bwd_kernel_dAv_npu(
         b_v = tl.load(p_v, boundary_check=(0, 1))
         b_do = tl.load(p_do, boundary_check=(0, 1))
         b_dA += tl.dot(b_do, b_v, allow_tf32=False)
```
zhiyuan1i, line 107: Same pattern here: `b_do` is the lhs of the `b_dA` dot and is then read as rhs for `b_dv = tl.dot(b_A_i, b_do, ...)`. Could you confirm this read still sees the original `do`?

zhiyuan1i (commented):
Thanks for digging into this — the root-cause analysis (Ascend `tl.dot` clobbering its lhs in UB, unlike CUDA Triton) is convincing, and the fixes in `wy_fast` fwd and `chunk_o` bwd (`b_A_pristine`, `b_ds_c`) look correct to me.

I'm leaving these as confirmation requests rather than change requests: several new guards copy the tile only *after* its first lhs dot, and a scan of the remaining `triton_ascend` kernels finds the same lhs-reuse pattern in a few places this PR doesn't touch. Could you confirm whether these are real exposures under the actual Triton-Ascend behavior? Beyond the inline ones, we also spotted:
- `fla/ops/common/backends/triton_ascend/chunk_o.py:514-515`: `b_k0` is the lhs of two consecutive dots (`b_A00`, `b_A01`).
- `fla/ops/common/backends/triton_ascend/chunk_delta_h.py:750,770,790,810` (`STATE_V_FIRST` path): `b_do` is reused as lhs across K slabs.

Two broader requests:

1. **Systematic audit.** Since the clobber semantics apply to every `tl.dot` on this backend, would you be able to help audit the remaining `triton_ascend` kernels for the same pattern (any tile used as lhs more than once, or read again after being a dot's lhs)? Fixing one kernel at a time is whack-a-mole; a single sweep — or a shared checklist in the fla-ascend-performance skill doc — would be much more reliable.

2. **Long-term compiler fix.** Hand-guarding every kernel is fragile: any new kernel can silently reintroduce the bug. Is lhs clobbering a deliberate Triton-Ascend design choice (e.g. UB capacity pressure) or a known backend limitation? The durable fix belongs in the compiler — making `tl.dot` preserve its lhs, or at least emitting a diagnostic when the lhs is read after the dot. If you can point us to the right upstream repo/component, we're happy to open a tracking issue for it.

Two nits on the PR body: the test plan lists only `test_gdn.py`/`test_kda.py`, but the A2 CI also runs `test_gdn_kernels.py` and `test_solve_tril.py` — worth listing since they cover the changed code directly. And the Benchmark section says `None` while the checklist claims same-hardware before/after numbers; since this adds GM reloads and extra UB tiles, representative numbers (or a corrected checklist) would help.

sunyi0505:
> Thanks for digging into this — the root-cause analysis (Ascend `tl.dot` clobbering its lhs in UB, unlike CUDA Triton) is convincing, and the fixes in `wy_fast` fwd and `chunk_o` bwd (`b_A_pristine`, `b_ds_c`) look correct to me.
> 
> I'm leaving these as confirmation requests rather than change requests: several new guards copy the tile only _after_ its first lhs dot, and a scan of the remaining `triton_ascend` kernels finds the same lhs-reuse pattern in a few places this PR doesn't touch. Could you confirm whether these are real exposures under the actual Triton-Ascend behavior? Beyond the inline ones, we also spotted:
> 
> * `fla/ops/common/backends/triton_ascend/chunk_o.py:514-515`: `b_k0` is the lhs of two consecutive dots (`b_A00`, `b_A01`).
> * `fla/ops/common/backends/triton_ascend/chunk_delta_h.py:750,770,790,810` (`STATE_V_FIRST` path): `b_do` is reused as lhs across K slabs.
> 
> Two broader requests:
> 
> 1. **Systematic audit.** Since the clobber semantics apply to every `tl.dot` on this backend, would you be able to help audit the remaining `triton_ascend` kernels for the same pattern (any tile used as lhs more than once, or read again after being a dot's lhs)? Fixing one kernel at a time is whack-a-mole; a single sweep — or a shared checklist in the fla-ascend-performance skill doc — would be much more reliable.
> 2. **Long-term compiler fix.** Hand-guarding every kernel is fragile: any new kernel can silently reintroduce the bug. Is lhs clobbering a deliberate Triton-Ascend design choice (e.g. UB capacity pressure) or a known backend limitation? The durable fix belongs in the compiler — making `tl.dot` preserve its lhs, or at least emitting a diagnostic when the lhs is read after the dot. If you can point us to the right upstream repo/component, we're happy to open a tracking issue for it.
> 
> Two nits on the PR body: the test plan lists only `test_gdn.py`/`test_kda.py`, but the A2 CI also runs `test_gdn_kernels.py` and `test_solve_tril.py` — worth listing since they cover the changed code directly. And the Benchmark section says `None` while the checklist claims same-hardware before/after numbers; since this adds GM reloads and extra UB tiles, representative numbers (or a corrected checklist) would help.

Thanks for the thorough review — you're right on all counts. The original + 0.0 guards were placed after the first lhs dot in several spots, which means they captured the clobbered UB value rather than the original tile. I've pushed a follow-up commit that fixes every inline concern plus the two additional exposures you spotted.

Additional exposures (confirmed real, now fixed):

chunk_o.py:514-515 — b_k0_c = b_k0 + 0.0 before the first dot; b_A00 uses b_k0, b_A01 uses b_k0_c.
chunk_delta_h.py K-slabs — b_do_c = b_do + 0.0 after gate/scale folding, before the first K-slab dot; slab 1 uses b_do, slabs 2–4 use b_do_c.
Systematic audit: Added an audit checklist to .agents/skills/fla-ascend-performance/SKILL.md (§4 + Review checklist) and expanded references/cases.md with copy-ordering rules and multi-use patterns (lhs→lhs, lhs→rhs, lhs→store, K-slab reuse). The rule of thumb: for every tl.dot, check whether the lhs tile is used again — if so, either copy before the first lhs dot or reload from GM.

Verification: test_gdn_kernels.py (54 passed) and test_solve_tril.py (10 passed) on Atlas 800T A3.

fla/ops/utils/backends/triton_ascend/solve_tril.py
```diff
@@ -277,9 +286,9 @@ def merge_16x16_to_64x64_inverse_kernel_npu(
     p_Ai_42 = tl.make_block_ptr(Ai, (T, BT), (H * BT, 1), (i_t * BT + 48, 16), (16, 16), (1, 0))
     p_Ai_43 = tl.make_block_ptr(Ai, (T, BT), (H * BT, 1), (i_t * BT + 48, 32), (16, 16), (1, 0))
     tl.store(p_Ai_11, b_Ai_11.to(p_Ai_11.dtype.element_ty, fp_downcast_rounding='rtne'), boundary_check=(0, 1))
-    tl.store(p_Ai_22, b_Ai_22.to(p_Ai_22.dtype.element_ty, fp_downcast_rounding='rtne'), boundary_check=(0, 1))
-    tl.store(p_Ai_33, b_Ai_33.to(p_Ai_33.dtype.element_ty, fp_downcast_rounding='rtne'), boundary_check=(0, 1))
-    tl.store(p_Ai_44, b_Ai_44.to(p_Ai_44.dtype.element_ty, fp_downcast_rounding='rtne'), boundary_check=(0, 1))
+    tl.store(p_Ai_22, b_Ai_22_c.to(p_Ai_22.dtype.element_ty, fp_downcast_rounding='rtne'), boundary_check=(0, 1))
+    tl.store(p_Ai_33, b_Ai_33_c.to(p_Ai_33.dtype.element_ty, fp_downcast_rounding='rtne'), boundary_check=(0, 1))
+    tl.store(p_Ai_44, b_Ai_44_c.to(p_Ai_44.dtype.element_ty, fp_downcast_rounding='rtne'), boundary_check=(0, 1))
```
zhiyuan1i: Almost there — but `b_Ai_44_c` is consumed as an lhs in the outer dot at :264-269 and is then stored here. Under the clobber semantics that store writes the clobbered value. `b_Ai_44_c2` is also an lhs (:270-276), so neither copy is store-safe. Please keep one more pristine copy of `b_Ai_44` that is never used as an lhs (created before the first dot at :253-257) and store that one instead.

zhiyuan1i (commented):
Thanks for the quick follow-up — I re-audited the new head and can confirm the fixes for the `chunk_o` fwd `b_q`, `chunk_intra` qg/kg copies and solve sequence, `wy_fast` `b_dw`/`b_du`, `chunk_bwd` dAv `b_do`, `chunk_o` `b_k0`, and the `chunk_delta_h` STATE_V_FIRST `b_do` slabs are all correct now (pristine source established before the first lhs dot, disposable copies for each lhs, store tiles never consumed as lhs). The audit checklist in the skill doc is exactly what we hoped for.

Two remaining items:

1. Inline: the 64x64 `solve_tril` path still stores a tile that was consumed as an lhs.

2. One more exposure the sweep missed (this one pre-exists on main, same file as your dAv fix): in `fla/ops/kda/backends/triton_ascend/chunk_bwd.py:554-588`, `b_dw_r` is loaded once per `s_r` outside the inner `s_c` loop, then used as the lhs of `tl.dot` at :587 on every `s_c` iteration (4 iterations by default with `BT=64, BC=16`). The first dot may clobber it for the remaining three. Note `b_dw_r.to(tl.float32)` doesn't save it here — `dw` is already fp32, so the cast is an identity with no guaranteed independent UB copy. Suggest deriving a per-iteration disposable copy from a pristine `b_dw_r` that is never an lhs, or reloading from `p_dw_r` inside the loop.

With those two addressed I expect this is good to approve.

sunyi0505:
> Thanks for the quick follow-up — I re-audited the new head and can confirm the fixes for the `chunk_o` fwd `b_q`, `chunk_intra` qg/kg copies and solve sequence, `wy_fast` `b_dw`/`b_du`, `chunk_bwd` dAv `b_do`, `chunk_o` `b_k0`, and the `chunk_delta_h` STATE_V_FIRST `b_do` slabs are all correct now (pristine source established before the first lhs dot, disposable copies for each lhs, store tiles never consumed as lhs). The audit checklist in the skill doc is exactly what we hoped for.
> 
> Two remaining items:
> 
> 1. Inline: the 64x64 `solve_tril` path still stores a tile that was consumed as an lhs.
> 2. One more exposure the sweep missed (this one pre-exists on main, same file as your dAv fix): in `fla/ops/kda/backends/triton_ascend/chunk_bwd.py:554-588`, `b_dw_r` is loaded once per `s_r` outside the inner `s_c` loop, then used as the lhs of `tl.dot` at :587 on every `s_c` iteration (4 iterations by default with `BT=64, BC=16`). The first dot may clobber it for the remaining three. Note `b_dw_r.to(tl.float32)` doesn't save it here — `dw` is already fp32, so the cast is an identity with no guaranteed independent UB copy. Suggest deriving a per-iteration disposable copy from a pristine `b_dw_r` that is never an lhs, or reloading from `p_dw_r` inside the loop.
> 
> With those two addressed I expect this is good to approve.

Fixed. Thanks for your suggestions!

zhiyuan1i (approved):

Issue #674, opened 2025-12-01
[RFC] Context Parallel Implementations


garrett361:
### Proposal

We propose adding context-parallel (CP) implementations of the kernels available in FLA. We will organize the CP GDN contribution such that its distributed components can be largely reused by other kernels and would like to introduce a common CP interface that will also be re-used across different FLA algorithms.

Specifically, we propose starting with the implementation for Gated DeltaNet (GDN), as we already began this work as part of our GPU MODE IRL hackthon effort ([see the third place entry here](https://www.gpumode.com/v2/news)). Some of the hackathon code, which will be cleaned and reorganized, can be found in #626. Working group: @Baladhurgesh @bargav25 @garrett361 @MatthewGleeson @noyonikap-glitch

The initial work will be relevant to all kernels which use the short causal convolution and/or are defined by a recurrent structure. GDN, Mamba, Mamba2, DeltaNet, LinearAttention and more fall into this category, for instance. CP will be implemented as follows:

- The CP causal convolution on CP rank `cp_rank` can performed after receiving a small amount of information from its predecessor rank, `cp_rank -1`, namely the last `K - 1` tokens, where `K` is the convolution kernel size. We will add these P2P ops and use a memory-efficient strategy for producing the correct convolution outputs.
- The CP recursion relations are handled by passing the final state of `cp_rank` to `cp_rank + 1`, where it serves as the latter device's initial state.  Also implemented through P2P.

Backwards passes through these CP operations can be handled similarly to the forwards passes.

The above components can wrap the existing FLA kernels, leaving the triton code untouched. Kernel-specific optimizations which go beyond the above and change the triton impls exist, but we will focus on simplicity, correctness, and reusability to start with, so that the initial GDN work can be quickly recycled for other kernels. Leaving the triton code unchanged also helps separate concerns and drastically increases maintainability by enabling the CP implementations to automatically inherit any triton kernel improvements or corrections automatically (assuming no changes of signature).

We plan to target only the training code paths, initially; inference code paths will be treated in subsequent PRs.

## CP Module API

A specific area we'd like feedback on is what the preferred CP interface for the various FLA `nn.Module` classes should look like. Some options we've considered:

- Create entirely new `nn.Module` classes for the CP implementations. E.g. a `GatedDeltaNetCP(nn.Module)` class which takes in a 1D `cp_device_mesh: DeviceMesh` argument in its `__init__` method.
[...]

PR #1116, opened 2026-08-11
[CP] Enable context parallelism for GDN-2

eric-ide:
## Summary

Enhancement for issue #674 . Enables context parallelism for GDN-2 through a similar implementation to that of KDA.

We reuse the existing CP machinery used by KDA, such as the preprocess. The formulation of GDN-2 is very similar to KDA and the structure of the summaries is the same, so the implementation is almost identical.

## Test plan

`pytest tests/context_parallel/test_cp_gdn2.py`

## Breaking changes

None. `cp_context` is optional, behaviour of GDN-2 without `cp_context` is unchanged.

## Checklist


zhiyuan1i (approved):
LGTM. Could you rebase onto the latest main? There are conflicts in `fla/ops/gdn2/chunk.py` now, since recent main changes (e.g. #1107, #1173, #1180) touched the shared chunk paths. A rebase would also help CI check it against the current main.

PR #1118, opened 2026-08-12
[Fix] Restructure Ascend chunk_gated_delta_rule_bwd_dhu to AICore task loop

sunyi0505:
## Summary

Fixes Ascend NPU timeout on `chunk_gated_delta_rule_bwd_dhu` under varlen stress tests (#1077) by restructuring the backward dh/du kernel launch model and addressing related correctness issues.
- **AICore task loop**: Replace the previous 2D grid launch `(nv_chunks, nh_total)` with a single-grid `(num_aicore,)` pattern. Each AICore iterates over `(batch, head)` tasks via `tl.range(core_id, task_num, num_core)`, and V-dimension tiling is moved into an inner `for i_v in range(NV)` loop inside the kernel. Forward `chunk_gated_delta_rule_fwd_h` is unified onto the same `_launch_core_grid` helper.
- **Ascend `tl.dot` workaround**: Introduce a separate `dv2` output buffer so intermediate `dv` results are written out-of-place, avoiding lhs clobbering by Ascend's `tl.dot` implementation across K-slabs.
- **int64 index safety**: Cast `cu_seqlens`, `bos`, and `eos` to `tl.int64` across all affected `triton_ascend` kernels (GDN, KDA, cumsum, solve_tril, chunk_o, etc.) to prevent overflow in long-sequence varlen paths.
- **Re-enable varlen stress test**: Remove the NPU skip on `test_chunk_gated_delta_rule_bwd_dhu_varlen_many_documents`.

## Test plan

0d0f90721e8d2df24b71520a8aab6ab0

image

image

## Breaking changes

None

## Checklist

### DIFF
--- a/fla/ops/common/backends/triton_ascend/chunk_delta_h.py
+++ b/fla/ops/common/backends/triton_ascend/chunk_delta_h.py
@@ -19,13 +19,10 @@ from fla.ops.utils.cache import fla_cache_autotune
 from fla.ops.utils.op import exp2
 from fla.utils import input_guard
 from fla.utils.ascend_ub_manager import (
-    ASCEND_MAX_GRID_DIM,
     compute_row_tile_block_size,
     get_ub_manager,
-    max_grid_axis_chunks,
 )
 
-_NUM_WARPS = 4
 # b_h[64,BV] fp32 + b_w[64,64] + b_v[64,BV] + b_k[64,64] peak during recurrence.
 _FWD_H_MEM_MULT = 8.0
 _SAFETY_MARGIN = 0.80
@@ -53,13 +50,7 @@ def _get_bv(K: int, V: int) -> int:
 
 def _dhu_peak_bytes(BK: int, BV: int, n_slabs: int, BT: int = _DHU_BT) -> int:
     # All K-slabs of dh stay live; q/w are one-slab at a time.
-    return (
-        n_slabs * BK * BV * 4
-        + 2 * BK * BT * 2
-        + BT * BV * 2
-        + BT * BV * 4
-        + 12 * BT
-    )
+    return n_slabs * BK * BV * 4 + 2 * BK * BT * 2 + BT * BV * 2 + BT * BV * 4 + 12 * BT
 
 
 def _dhu_tile_cost(K: int, V: int, BK: int, BV: int) -> int:
@@ -81,9 +72,7 @@ def _select_bwd_dhu_tiles(
     if state_v_first:
         return 64, _get_bv(K, V)
 
-    soft_cap = int(get_ub_manager().ub_capacity_bytes * (
-        _DHU_UB_GATE_INLINE if gate_inline else _DHU_UB_SOFT
-    ))
+    soft_cap = int(get_ub_manager().ub_capacity_bytes * (_DHU_UB_GATE_INLINE if gate_inline else _DHU_UB_SOFT))
     # bwd kernel is blockdim64: K is covered by up to four BK=64 slabs only.
     max_bk = 64
     desired_v = triton.next_power_of_2(V)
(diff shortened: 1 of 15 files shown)


sunyi0505:

fla/ops/common/backends/triton_ascend/chunk_delta_h.py
```diff
@@ -532,319 +518,334 @@ def chunk_gated_delta_rule_bwd_kernel_dhu_blockdim64_npu(
...
-            if USE_GK:
-                o_k4 = 3 * BK + o_k1
-                b_gk_last4 = tl.load(gk + last_idx * HV * K + o_k4, mask=(o_k4 < K), other=0.).to(tl.float32)
-            if STATE_V_FIRST:
-                b_dv_part = tl.dot(b_dh4.to(b_k.dtype), tl.trans(b_k), allow_tf32=False)
-                b_dv += tl.trans(b_dv_part)
-            else:
-                b_dv += tl.dot(b_k, b_dh4.to(b_k.dtype), allow_tf32=False)
+            if USE_FINAL_STATE_GRADIENT:
+                dht_base = dht + i_nh * K * V
```
zhiyuan1i, line 597: Non-blocking follow-up: `i_nh * K * V` is computed before any int64 promotion — `i_nh` comes out of the task loop as an int32 scalar, so with large `B*HV` and `K*V` (e.g. 256x256) this product can overflow int32 before being widened. Same pattern at :825 (`dh0_base`). Please cast `i_nh` to int64 *before* the multiply, consistent with the other offsets fixed in this PR.

fla/ops/common/backends/triton_ascend/chunk_delta_h.py
```diff
@@ -906,50 +907,53 @@ def chunk_gated_delta_rule_bwd_dhu_npu(
     else:
         dh = q.new_empty(B, NT, HV, K, V)
     dh0 = torch.empty_like(h0, dtype=torch.float32) if h0 is not None else None
+    dv2 = torch.empty_like(dv)
```
zhiyuan1i, line 912: Non-blocking follow-up: now that #1113 has landed, this full-size `dv2` buffer duplicates the pristine-copy guard for the same lhs-clobber pattern, at the cost of an extra dv-sized allocation. Consider reusing the #1113 pattern here (a pristine `dv` plus a disposable per-slab lhs copy) in a follow-up instead of keeping a second output buffer.

zhiyuan1i (approved):
The AICore task-loop restructure looks correct — task partitioning, tail tasks, empty sequences, and per-task state reset all check out statically, and it's good to see the re-enabled many-document varlen stress test actually running green on the A2 CI (946 passed). Routing `fwd_h` through the same launch helper is justified by the same timeout.

Two non-blocking follow-ups inline (int64 promotion order; the dv2 buffer now duplicating the #1113 pristine-copy guard).

One suggestion, also non-blocking: the dhu tests don't cover `K=V=256` + `state_v_first=True` + `USE_GK=True`, so the third/fourth K slabs never execute in CI — a single such case would close that gap. And when you get a chance, the before/after numbers for the 2050x16 varlen case (base timeout vs head completion) would be useful for the record.

sunyi0505:
> The AICore task-loop restructure looks correct — task partitioning, tail tasks, empty sequences, and per-task state reset all check out statically, and it's good to see the re-enabled many-document varlen stress test actually running green on the A2 CI (946 passed). Routing `fwd_h` through the same launch helper is justified by the same timeout.
> 
> Two non-blocking follow-ups inline (int64 promotion order; the dv2 buffer now duplicating the #1113 pristine-copy guard).
> 
> One suggestion, also non-blocking: the dhu tests don't cover `K=V=256` + `state_v_first=True` + `USE_GK=True`, so the third/fourth K slabs never execute in CI — a single such case would close that gap. And when you get a chance, the before/after numbers for the 2050x16 varlen case (base timeout vs head completion) would be useful for the record.

Thanks — both follow-ups are in.

i_nh is now task_id.to(tl.int64) at the top of both task loops, so i_nh * K * V is widened before the multiply (dht/dh0 and the matching h0/ht sites in fwd).
Kept dv2: it matches the CUDA output contract (do not mutate input dv). The #1113 + 0.0 copies are the in-register lhs-clobber guard and are independent of that allocation. 
Added test_chunk_gated_delta_rule_bwd_dhu_k256_state_v_first_gk so K-slabs 3/4 run in CI.
2050×16 varlen bwd_dhu on A2 (N=2050, T=16, HV=32, D=32): base was an AICore timeout (test skipped); first launch including compile is 0.95 s, steady-state 27.1 ms.

test_chunk_gated_delta_rule_bwd_dhu_k256_state_v_first_gk  passed.
image

Issue #1119, opened 2026-08-12
[Bug] chunk_kda/fused_recurrent_kda **kwargs silently swallow unknown flags — Kimi K3's use_beta_sigmoid_in_kernel=True is a no-op on 0.5.0, wrong numerics with no warning


xiaguan:
### Summary

`chunk_kda` / `fused_recurrent_kda` accept a trailing `**kwargs`, so a semantically **required** flag that a given release does not know about is silently discarded instead of raising. Concretely: Kimi K3's published remote modeling code calls both entry points with `use_beta_sigmoid_in_kernel=True`. On fla-core/flash-linear-attention **0.5.0** that parameter does not exist in either signature — the call lands in `**kwargs`, beta is never passed through a sigmoid, and the model computes wrong recurrences **with no error and no warning**. The parameter exists on 0.5.2 (and current `main`), so up-to-date installs are fine, but any environment pinned to 0.5.0 (ours was) runs K3 silently wrong on the HF-transformers + public-fla path.

### How we found it

While building per-boundary numerical oracles for Kimi K3 (bf16, checking a re-implementation against the published modeling code + fla as reference), the KDA boundary refused to converge past cosine ≈ 0.5 against the recurrent path. An exhaustive hypothesis matrix over {scale, state layout, decay axis, beta sigmoid, bias, l2norm} matched the fla 0.5.0 output **bitwise** only with the beta sigmoid *disabled* — i.e. 0.5.0 was not applying it, despite the model code passing `use_beta_sigmoid_in_kernel=True`. Inspecting the 0.5.0 signatures confirmed the kwarg is absorbed by `**kwargs`:

```python
import inspect
from fla.ops.kda import chunk_kda, fused_recurrent_kda   # fla 0.5.0
'use_beta_sigmoid_in_kernel' in inspect.signature(chunk_kda).parameters            # False
'use_beta_sigmoid_in_kernel' in inspect.signature(fused_recurrent_kda).parameters  # False
# both signatures end in **kwargs -> call succeeds, flag ignored
```

The intent is unambiguous (the docstrings on current main require pre-sigmoided beta when the flag is off, and vLLM's KDA implementation applies the sigmoid), so 0.5.0 + K3 remote code = silently wrong numerics.

### Ask

[...]

PR #1124, opened 2026-08-13
[Rodimus] Fix TypeError from stale head_first argument in chunk path

EastZeus:
## Summary

`chunk_gla` dropped its `head_first` parameter in the repo-wide cleanup (4a7a9f4c), but `RodimusAttention` still passes `head_first=False`. The layer defaults to `mode='chunk'` and only falls back to `fused_recurrent` for single-token decode, so any Rodimus forward with more than one token currently raises `TypeError` at the `chunk_gla` call. The Rodimus model tests cover exactly this shape (T=1024), but CI selects tests from changed files and the GLA cleanup never touched the Rodimus files, which is how the stale call site survived.

Changes:

- Drop `head_first=False` from the `chunk_gla` call (the crash) and from the `fused_recurrent_gla` call, where a bare `**kwargs` was silently swallowing it (the same failure class as #1119).
- Remove the `fused_chunk` mode from the allowed set. `fused_chunk_gla` is a permanent `NotImplementedError` stub, so the mode could be selected at construction time but always failed at runtime. Its branch also never passed `state_v_first=True`, so had it ever worked it would have used a different state layout than the other two paths.

Blast radius: Rodimus only. The chunk path goes from crashing to working; no numerics change anywhere else. No checkpoint or state-dict impact.

## Test plan

- `tests/models/test_modeling_rodimus.py` exercises the fixed path directly (forward/backward at T=1024 and generation at T=2000), and this PR touches `fla/layers/rodimus.py`, so CI will select it: `python scripts/find_dependent_tests.py fla/layers/rodimus.py` returns `tests/models/test_modeling_rodimus.py`, `tests/models/test_hybrid_attention.py`, `tests/layers/test_attn_varlen_pack_layout.py`, `tests/layers/test_layer_cache_layer_idx.py`.
- No GPU available locally; verified the fix statically against the current `chunk_gla` / `fused_recurrent_gla` signatures, plus `ruff check` and byte-compilation.

Neutral, no kernel changes.

## Breaking changes

Constructing `RodimusAttention` (or a `RodimusConfig`) with `attn_mode='fused_chunk'` now fails the mode assert at construction time instead of raising `NotImplementedError` on the first forward call. No working configuration is affected since the mode never ran. If you would rather keep the mode accepted and failing at runtime, I am happy to drop that part and keep only the `head_first` fix.

## Checklist

- [x] Dependent tests pass locally or in CI; new behavior is covered by tests where applicable. (Existing Rodimus model tests cover the fixed path; they need CI hardware.)
- [x] Kernel changes include same-hardware before/after benchmark numbers (dense + varlen where applicable). (Not a kernel change.)

### DIFF
--- a/fla/layers/rodimus.py
+++ b/fla/layers/rodimus.py
@@ -28,7 +28,7 @@ from fla.layers.utils import (
 )
 from fla.modules import RMSNorm, RotaryEmbedding, ShortConvolution
 from fla.modules.layernorm_gated import RMSNormGated
-from fla.ops.gla import chunk_gla, fused_chunk_gla, fused_recurrent_gla
+from fla.ops.gla import chunk_gla, fused_recurrent_gla
 
 if TYPE_CHECKING:
     from transformers.processing_utils import Unpack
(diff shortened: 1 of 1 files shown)


zhiyuan1i (approved):
Overall， LGTM

PR #1129, opened 2026-08-14
[Docs] Fix broken star history chart link

OctoBored:
The star history chart embedded in the README is currently broken because it relies on the GitHub stargazer API, which now restricts such requests. This change points the chart to a working alternative provider so the chart renders correctly again.

The badge (Stargazers roster) is left untouched.


zhiyuan1i:
Closing as moot — the Star History section was removed from the README in #1168, so there is no chart link left to fix. Thanks for the contribution anyway!

PR #1130, opened 2026-08-15
[Perf] Optimize Ascend KDA intra backward with split kernels and 1D core-grid

sunyi0505:
## Summary

Refactor KDA chunk intra backward on Ascend NPU: split the original monolithic chunk_kda_bwd_kernel_intra_npu into three independent kernels based on debug_barrier, and adopt a 1D AICore task loop to increase BC/BK, reduce Cube tile counts and host launch overhead.

Kernel Splitting (split by UB peak value with independent tiling for each)
• chunk_kda_bwd_kernel_intra_dq_db_npu: past + diag path, writes dq / db, and spills dk_part / dg_part to workspace
• chunk_kda_bwd_kernel_intra_dkt_future_npu: dkt for future subchunks; skip directly for the last chunk which has no future
• chunk_kda_bwd_kernel_intra_dk_dg_npu: diag path + accumulation of dk_part / dkt_part, outputs dk / dg

Scheduling
• Replace 3D grid plus sliced launch on host side with _KDA_BWD_INTRA_LAUNCH_BLOCK_BUDGET, using grid=(num_aicore,) + tl.range(core_id, task_num, num_core)
• Single launch covers all NK × NC × NT × B × HV tasks, remove the 4096‑block budget loop

Tiling (chunk_size=64, D=256, bf16)
• BC 16 → 32: NC 4 → 2; pair‑block of dkt_future drops from 6 to 1; Cube tile changes from 16×16 to 32×32
• dq_db: mem_mult=9, maximum BK=128 (NK 4 → 2)
• dk_dg: smaller live set, mem_mult=4.5, maximum BK=256 (NK=1)
• dkt_future keeps BK=128: BK=256 passes compilation but triggers UB address OOB at runtime
• fp32 I/O caps BK upper limit to 32 to prevent UB overflow on the SAFE_GATE diag path

Memory Layout
• Transpose g / beta to [B, HV, T, …] when HV > 1 for stride‑1 load along T
• Use int64 for pointers to avoid cu_seqlens overflow under varlen scenarios
• Rebind local pointers for each task (q_ptr = q + …) to circumvent incorrect compilation of ptr += in Ascend Triton

## Test plan

hardware: Atlas 800T A3(X86)
c9c95b2c2a2f0a848de739c497608e99

image

## Breaking changes

None

## Checklist

### DIFF
--- a/fla/ops/kda/backends/triton_ascend/chunk_intra.py
+++ b/fla/ops/kda/backends/triton_ascend/chunk_intra.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import torch
 import triton
 import triton.language as tl
+from triton.runtime import driver
 
 from fla.ops.kda.backends.triton_ascend.wy_fast import recompute_w_u_fwd_kda_npu as _recompute_w_u_fwd_npu
 from fla.ops.kda.chunk_intra_token_parallel import chunk_kda_fwd_intra_token_parallel
@@ -162,13 +163,14 @@ def chunk_kda_fwd_kernel_diag_solve_npu(
         bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
         T = (eos - bos).to(tl.int32)
     else:
-        bos, eos = i_b * T, i_b * T + T
+        bos = tl.cast(i_b, tl.int64) * T
+        eos = bos + T
 
     i_ti = i_t * BT + i_i * BC
     if i_ti >= T:
         return
 
-    Akkd = Akkd + (bos * HV + i_hv) * BC
+    Akkd = Akkd + (bos * HV + i_hv).to(tl.int64) * BC
     o_i = tl.arange(0, BC)
     m_A = o_i[:, None] > o_i[None, :]
     m_I = o_i[:, None] == o_i[None, :]
@@ -177,7 +179,7 @@ def chunk_kda_fwd_kernel_diag_solve_npu(
     b_Akk = tl.load(p_Akk, boundary_check=(0, 1)).to(tl.float32)
     b_Ai = -tl.where(m_A, b_Akk, 0)
     for i in range(2, min(BC, T - i_ti)):
-        b_a = -tl.load(Akkd + (i_ti + i) * HV * BC + o_i)
+        b_a = -tl.load(Akkd + (i_ti + i).to(tl.int64) * HV * BC + o_i)
         b_a = tl.where(o_i < i, b_a, 0.)
         b_a += tl.sum(b_a[:, None] * b_Ai, 0)
         b_Ai = tl.where((o_i == i)[:, None], b_a, b_Ai)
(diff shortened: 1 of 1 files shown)


sunyi0505:

zhiyuan1i:
Thanks for the split-kernel rework. I checked the three-launch dependency chain (dq_db → dkt_future → dk_dg, all stream-ordered), the NC_FUT task skip, the int64 casts, and tl.dot lhs-clobber exposure — all look correct, and the 1D core-grid launcher matches the merged #1118 pattern. The fwdbwd numbers look great.

One suggestion (fine to defer to a follow-up): `dk_part`/`dg_part` duplicate the already-allocated `dk2`/`dg2`. Since the launches are stream-ordered and dk_dg loads each tile before storing back to it, dq_db could write the past+diag partials straight into `dk2`/`dg2` and let dk_dg accumulate in place — that drops two full [B,T,HV,K] tensors (~3 GiB at B=8,T=8192,HV=64,K=128), same consideration as the `dv2` cleanup in #1118.

One thing to confirm: KDA's fwd/bwd numerics are validated at scale, and this PR introduces one point that diverges from that baseline — `dk_part` keeps dk's bf16 dtype while `dg_part`/`dkt_part` are fp32, so the past+diag partial gets an intermediate bf16 rounding before `dkt` is added back (base kept it in fp32 until the final store). Is this intentional and covered by your accuracy validation?

sunyi0505:
> Thanks for the split-kernel rework. I checked the three-launch dependency chain (dq_db → dkt_future → dk_dg, all stream-ordered), the NC_FUT task skip, the int64 casts, and tl.dot lhs-clobber exposure — all look correct, and the 1D core-grid launcher matches the merged #1118 pattern. The fwdbwd numbers look great.
> 
> One suggestion (fine to defer to a follow-up): `dk_part`/`dg_part` duplicate the already-allocated `dk2`/`dg2`. Since the launches are stream-ordered and dk_dg loads each tile before storing back to it, dq_db could write the past+diag partials straight into `dk2`/`dg2` and let dk_dg accumulate in place — that drops two full [B,T,HV,K] tensors (~3 GiB at B=8,T=8192,HV=64,K=128), same consideration as the `dv2` cleanup in #1118.
> 
> One thing to confirm: KDA's fwd/bwd numerics are validated at scale, and this PR introduces one point that diverges from that baseline — `dk_part` keeps dk's bf16 dtype while `dg_part`/`dkt_part` are fp32, so the past+diag partial gets an intermediate bf16 rounding before `dkt` is added back (base kept it in fp32 until the final store). Is this intentional and covered by your accuracy validation?

Addressed.

dq_db now writes the past+diag partials directly into dk2/dg2; dk_dg loads each tile and accumulates in place. That removes the extra dk_part/dg_part [B,T,HV,K] workspaces.

CUDA keeps that partial in fp32 until the final store; dk2 is now fp32 (same as dg2/dkt_part) so dkt is fused before any downcast. Covered by the existing test_chunk / test_chunk_varlen fwd+bwd suite, including bf16.

zhiyuan1i (approved):

PR #1154, opened 2026-08-20
[Fix] Drop unused NB from causal_conv1d autotune cache key

### LINKED ISSUE #1147: [Bug] causal_conv1d repeatedly autotunes with dynamic packed sequence lengths
### Checklist

### Describe the Bug

```markdown
### Description

When using FLA through Megatron-Core with variable-length sequence packing, the packed token count changes between iterations. This causes `causal_conv1d_fwd_kernel` and `causal_conv1d_bwd_kernel` to repeatedly run Triton autotuning during training.

The autotuning appears as `do_bench` on the GPU timeline and can cause a large iteration-time spike. In one workload, throughput temporarily dropped from approximately 5,000–7,000 to 1,550 tokens/s/GPU.

### Root cause

Both kernels use:

```python
@fla_cache_autotune(
    ...
    key=['D', 'W', 'NB'],
)
```

where `NB` is computed as:

```python
NB = triton.cdiv(B * T, 1024)
```

With dynamic packing, `T` changes frequently, producing new `NB` values and therefore new autotune keys.

However, `NB` is not used in either kernel body. It is only passed as a `tl.constexpr` and included in the autotune key, so it unnecessarily fragments both the autotune and compilation caches.

fla/modules/conv/triton/kernels.py  
causal_conv1d_fwd_kernel and causal_conv1d_bwd_kernel  
change NB param

### Expected Behavior

Changing packed sequence lengths or `cu_seqlens` should not trigger a new autotune run when `D`, `W`, dtype, and kernel mode remain unchanged.

### Environment Information

1. Torch: 2.10
2. Triton: 3.4.0   
- Megatron-Core training
- BF16
- `D=8192`, `W=4`
- 8 GPUs
- Dynamically packed sequence lengths around 16K–19K

AmirF194:
## Summary

`causal_conv1d_fwd_kernel` and `causal_conv1d_bwd_kernel` (`fla/modules/conv/triton/kernels.py`) declare `NB: tl.constexpr` and are decorated `@fla_cache_autotune(..., key=['D', 'W', 'NB'], ...)`. `NB` is computed in the host wrapper as `triton.cdiv(B*T, 1024)` (`fla/modules/conv/triton/ops.py`) and passed through to the kernel launch, but neither kernel body reads it: it exists only in the signature and the autotune key.

Because `NB` sits in the key, `AutotuneKey.build` (`fla/ops/utils/cache.py`) produces a distinct cache entry for every distinct `B*T`, even when `D` and `W`, the values that actually determine which config is fastest, are unchanged. Under dynamic sequence packing (`B*T` varies every step, e.g. Megatron-Core variable-length packing), this forces a fresh autotune benchmark sweep on every step instead of reusing the config already found for the same `D`/`W`. Issue #1147 reports the production consequence: throughput dropped from ~5,000-7,000 to ~1,550 tokens/s/GPU during training as autotuning kept re-triggering.

Fix: drop `NB` from the `key` list on both kernels. `NB` stays in the kernel signature and launch call unchanged; only the cache key narrows.

Fixes #1147

## Test plan

- `tests/ops/test_cache.py::test_causal_conv1d_autotune_key_excludes_unused_nb` (new, parametrized over both kernels): asserts `NB` is absent from the decorated kernel's autotune `keys`, and that `AutotuneKey.build` (the exact function `CachedAutotuner.run` calls) produces the same cache key for two calls sharing `D`/`W` but differing `NB`. It fails on main (`NB` present, keys differ) and passes on this branch.
- No GPU is available on the authoring machine, and this repo's `tests/ops`/`tests/modules` autouse `poison_torch_memory` fixture calls `torch.cuda.synchronize()` at teardown unconditionally, so I could not run the test through `pytest` end to end here. I instead ran the same assertions directly against the real decorated kernel objects and `AutotuneKey.build` (no kernel launch, no GPU op) via `python -m pytest`-free scripts on main and on this branch, confirming the fail-on-main/pass-on-branch differential. `scripts/find_dependent_tests.py` also names `tests/modules/test_conv.py` and `tests/context_parallel/test_cp_conv.py`; both are GPU-only numerical tests unaffected by this change (it touches only cache-key metadata, not kernel arithmetic) and I did not run them.
- `pre-commit run --files fla/modules/conv/triton/kernels.py tests/ops/test_cache.py` and `python scripts/check_header.py --check` pass locally.

Neutral: the diff removes one entry from a Python-level cache-key list, it does not touch kernel arithmetic, launch parameters, or compiled code. No GPU available on the authoring machine to produce before/after numbers regardless.

## Breaking changes

None. `NB` remains a kernel argument; only the autotune cache key narrows.

## Checklist

### DIFF
--- a/fla/modules/conv/triton/kernels.py
+++ b/fla/modules/conv/triton/kernels.py
@@ -30,7 +30,7 @@ STATIC_WARPS = 32 if not IS_AMD else 16
         for BD in [16, 32, 64, 128]
         for num_warps in NUM_WARPS_AUTOTUNE
     ],
-    key=['D', 'W', 'NB'],
+    key=['D', 'W'],
     **autotune_cache_kwargs,
 )
 @triton.jit
@@ -151,7 +151,7 @@ def causal_conv1d_fwd_kernel(
         for BD in [16, 32, 64, 128]
         for num_warps in [4, 8, 16, 32]
     ],
-    key=['D', 'W', 'NB'],
+    key=['D', 'W'],
     **autotune_cache_kwargs,
 )
 @triton.jit
--- a/tests/ops/test_cache.py
+++ b/tests/ops/test_cache.py
@@ -10,7 +10,8 @@ import torch
 import triton
 import triton.language as tl
 
-from fla.ops.utils.cache import fla_cache_autotune
+from fla.modules.conv.triton.kernels import causal_conv1d_bwd_kernel, causal_conv1d_fwd_kernel
+from fla.ops.utils.cache import AutotuneKey, fla_cache_autotune
 from fla.utils import device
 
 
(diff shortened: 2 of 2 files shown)


zhiyuan1i (approved):

zhiyuan1i:
In the past, Auto-Tune was primarily used for achieving the best possible audio quality during the mixing and mastering process. However, it now seems to be causing more problems than it solves.

Overall, LGTM. Thanks!

AmirF194:
Ha, fair point about the audio kind causing more problems than it solves. Thanks for the quick review and merge.

PR #1167, opened 2026-08-24
[Perf] Optimize Ascend fused_attnres with N-tiling, fused bwd prep, and residual reuse

sunyi0505:
## Summary

Optimize the Ascend NPU (`triton_ascend`) fused AttnRes kernels to vectorize along the contiguous N axis, cut host TensorMoves, and skip redundant restacks / prep launches. Public AttnRes API is unchanged; GPU path is untouched.
Tiling / vectorization
- Add `_get_bn` (`BN` up to 8). Grid is over N-tiles, not one program per token.
- `BN==1` keeps 1D `[BD]` tiles: a 2D gather along L compiles to scalar on Triton-Ascend. `BN>1` uses contiguous `[BN, BD]` tiles along N.
- `SINGLE_D` (`BD >= D`) hoists `q*w` out of the L loop and keeps tiles in UB.
- `fwd_p2` is D-outer so `b_o` stays in UB and `o_mix` is written once per D-tile. `LOAD_OMIX=0` skips the first-chunk zero load.
Host / launch savings
- When the stacked residual fits in the 768 MiB L-chunk, reuse it for `fwd_p2` and save it for bwd instead of restacking sources.
- `checkpoint_level=1` still keeps fp32 `o_mix` so bwd skips the p2 restack.
- `_dres_chunk` returns views into the stacked buffer instead of `copy_` out (avoids a handful of TensorMoves per chunk).
- Replace `attnres_bwd_kernel_dqdw_npu` with a host `dqw.sum(dim=0)`.
Backward fusion
- No-onorm + `SINGLE_D`: `FUSE_PREP` inlines `bwd_prep` (`delta` from `o_mix`) into `bwd_dv`. `LOAD_DQW=0` skips the first-chunk zero load of `dqw`.
Bench isolation
- `benchmarks/ops/run.py` / `verify.py`: HEAD and `--base` each run in an isolated subprocess so the parent never holds NPU/CUDA tensors across refs (avoids HBM starvation). Add `--no-base`; lazy-import `device_name`.

## Test plan

hardware: Atlas 800T A3(X86)
image

image

## Breaking changes

None

## Checklist

### DIFF
--- a/fla/ops/attnres/backends/triton_ascend/fused.py
+++ b/fla/ops/attnres/backends/triton_ascend/fused.py
@@ -9,14 +9,20 @@
 
 GPU passes a padded tuple of per-source pointers (``res[i]``). Triton-Ascend
 cannot bind tuple arguments, so sources are stacked in L-chunks as
-``[L_chunk, N, D]`` (capped at 512 MiB per chunk) instead of one ``[L, N, D]``
-buffer. Cross-chunk softmax stats are merged on the host; the weighted sum
-accumulates in a single fp32 ``o_mix`` buffer to match GPU register precision.
+``[L_chunk, N, D]`` (capped at 768 MiB per chunk). Cross-chunk softmax
+stats are merged on the host; the weighted sum accumulates in a single
+fp32 ``o_mix`` buffer to match GPU register precision.
+
+A 2D gather along L compiles to scalar, so ``fwd_p2`` / ``bwd_dv`` use 1D
+``[BD]`` tiles when ``BN==1``, and contiguous ``[BN, BD]`` tiles along N
+otherwise. ``BN>1`` is only launched when ``SINGLE_D`` (``BD >= D``); the
+host falls back to ``BN==1`` for the D-loop path. ``SINGLE_D`` keeps
+``q*w`` in UB; no-onorm ``SINGLE_D`` fuses ``bwd_prep`` (``FUSE_PREP``).
 """
 
 from __future__ import annotations
 
-from collections.abc import Callable, Iterator, Sequence
+from collections.abc import Callable, Iterable, Iterator, Sequence
 
 import torch
 import triton
@@ -32,9 +38,13 @@ from fla.utils.ascend_ub_manager import (
 
 _FWD_MEM_MULT = 3.0
 _BWD_DV_MEM_MULT = 6.0
+_FWD_BN_MEM_MULT = 3.0
+# bwd_dv holds more live tiles under Ascend multi-buffering than the host UB model.
+_BWD_BN_MEM_MULT = 8.0
 _SAFETY_MARGIN = 0.85
 _FALLBACK_BD = 256
-_MAX_L_CHUNK_BYTES = 512 << 20
+_MAX_BN = 8
+_MAX_L_CHUNK_BYTES = 768 << 20
 
 
 def _l_chunk_size(L: int, N: int, D: int, elem_size: int) -> int:
@@ -76,10 +86,6 @@ def _merge_online_softmax(
     return m_new, acc_new
 
 
-def _get_bl(L: int) -> int:
-    return min(8, max(1, triton.next_power_of_2(L)))
-
-
 def _get_bd(row_dim: int, col_dim: int, *, memory_multiplier: float) -> int:
     return compute_row_tile_block_size(
         row_dim,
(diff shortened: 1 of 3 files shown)


sunyi0505:

fla/ops/attnres/backends/triton_ascend/fused.py
```diff
@@ -504,7 +785,11 @@ def fused_attnres_fwd_npu(
             eps=eps,
             BD=_get_bd(1, D, memory_multiplier=2.0),
         )
-    return o, o_pre, rstd, logit, lse
+    # keep fp32 o_mix at ckpt=1 so bwd skips the p2 restack
+    if o_pre is None:
+        o_pre = o_mix
```
zhiyuan1i: This changes the ckpt=1 memory semantics. Previously nothing was saved and bwd recomputed o_mix; now every layer keeps an fp32 o_mix (N*D*4 bytes) alive from fwd to bwd. All layers' buffers are alive at the same time during training — e.g. 60 layers at N=8192, D=4096 adds ~7.7GiB — so ckpt=1 becomes more memory-hungry than ckpt=0 (N*D*2B), and unlike `stacked` there is no size cap here. The math is unchanged, but users pick ckpt=1 expecting a low-memory recompute mode. Please either (a) document this as a Default before/after note with the memory formula (N*D*4B x layers) in the PR body and the checkpoint_level docstring, or (b) add a size gate that falls back to the bwd recompute path above a threshold.

fla/ops/attnres/backends/triton_ascend/fused.py
```diff
@@ -126,41 +147,108 @@ def attnres_fwd_p1_chunk_kernel(
...
+                b_v_qw = tl.sum(tl.where(m_nd, b_v * b_qw[None, :], 0.0), axis=1)
+                b_rstd = tl.rsqrt(b_v_sq / D + eps)
+                b_logit = b_v_qw * b_rstd
+                b_s = tl.where(m_n, b_logit * scale, float('-inf'))
+                b_mp = b_m
+                b_m = tl.maximum(b_m, b_s)
+                b_acc = b_acc * exp(b_mp - b_m) + exp(b_s - b_m)
+                tl.store(rstd + g_l * N + o_n, b_rstd.to(rstd.dtype.element_ty), mask=m_n)
+                tl.store(logit + g_l * N + o_n, b_logit.to(logit.dtype.element_ty), mask=m_n)
+        else:
```
zhiyuan1i: This BN>1 && !SINGLE_D arm is only reachable on A3-class UB with D>8192, so A2 CI and the current test matrix (D<=7186) never exercise it. The masks/math look right on audit, but an unreachable and untested path is a regression blind spot: every reachable route should be an optimized path with tests, an existing fallback, or an explicit unsupported error — not an implicit fourth state. Please delete this arm (bn==1 already covers it correctly) or document the reachability condition and a coverage plan.

fla/ops/attnres/backends/triton_ascend/fused.py
```diff
@@ -297,90 +417,234 @@ def attnres_bwd_dv_chunk_kernel(
...
+                b_v = b_qw[None, :] - b_v
+                b_v = (b_ds * b_rstd)[:, None] * b_v
+                b_v = b_p[:, None] * b_do + b_v
+                tl.store(
+                    dres + i_l64 * stride_ln + o_n[:, None] * D + o_d[None, :],
+                    b_v.to(dres.dtype.element_ty),
+                    mask=m_nd,
+                )
+            tl.store(dqw + o_n[:, None] * D + o_d[None, :], b_dqw, mask=m_nd)
+        else:
```
zhiyuan1i, line 535: Same issue as the p1 arm above: with _BWD_BN_MEM_MULT=8.0 this BN>1 && !SINGLE_D branch is unreachable on any chip (bn is always 1). Please delete it or document the reachability condition — dead code without test coverage will mislead future maintainers.

zhiyuan1i (changes requested):
Verified correctness branch by branch (int64 offsets, mask coverage, FUSE_PREP equivalence with the prep kernel, view aliasing, zero GPU-path impact), and the perf evidence is same-machine before/after with full registry shape coverage. Two majors to resolve before this can land:

sunyi0505:
> Verified correctness branch by branch (int64 offsets, mask coverage, FUSE_PREP equivalence with the prep kernel, view aliasing, zero GPU-path impact), and the perf evidence is same-machine before/after with full registry shape coverage. Two majors to resolve before this can land:

ckpt=1 

Agreed. ckpt=1 now matches CUDA: fwd drops o_mix, bwd recomputes it from the sources. ckpt=0 still keeps the mixed residual. Math is unchanged; ckpt=1 is again the low-memory recompute path.

fwd p1 BN>1 && !SINGLE_D

Deleted. Host now launches BN>1 only when SINGLE_D (BD >= D); otherwise it falls back to BN==1, which already covers the D-loop correctly.

bwd BN>1 && !SINGLE_D

Same as p1: deleted the untested arm. _n_tile_bn clamps to BN=1 whenever BD < D, so that path cannot be launched.

zhiyuan1i (approved):

PR #1180, opened 2026-08-27
[CP] use tf32x3 affine chain in kcp

iclementine:
## Summary

Use tf32x3 in state update & affine transform matrix merging(pre_process_bwd_kernel_merged, pre_process_fwd_kernel_merged and merge_fwd_bwd_kernel) to avoid precision loss in long context.

KCP works like a 3-kernel prefix scan.
1. The first kernel that works as block-reduce is `pre_process_fwd_kernel_merged` or `pre_process_bwd_kernel_merged`, where the transformation done by a sub-sequence is summarized into affine parameters M & H by combining corresponding parameters for each chunk in the subsequence. The transition matrix M is combined by matmul Ms from all chunks. The matmul is `tl.dot(f32, f32, input_precision="tf32")` by default.
2. The second kernel that works as root-scan is `merge_fwd_bwd_kernel`, where the initial state is consecutively transformed by the affine parameters (M & H) for each sub-sequence. Here the transform is again done in `tl.dot(f32, f32, input_precision="tf32")`.
3. The third kernel that works as block-scan is `chunk_gated_delta_rule_fwd_h` or `chunk_gated_delta_rule_bwd_dhu`. With the corrected initial states produced by the first 2 kernels, it works the in the same way as in cases where no CP is involved. In this kernel, however, the state propagation along chunks is done in high precision. It is not a standard affine transform (S* = MS+H), but a subtle formula (S* = Diag(gamma)S + dot(Aqk, pseudo_V) instead. Matmul between Diag(gamma) and S is done in f32 fma on cuda cores, which has higher precision that TF32 matmuls on tensor cores.

This introduces numerical loss in KCP compared to KDA without CP, which accumulates as sequence length grows. We use tf32x3 emulation in the affine chain in the first 2 kernels to alleviate this numerical loss. Please refer to 

Ootomo, Hiroyuki, and Rio Yokota. “Recovering Single Precision Accuracy from Tensor Cores While Surpassing the FP32 Theoretical Peak Performance.” The International Journal of High Performance Computing Applications 36, no. 4 (2022): 475–91. https://doi.org/10.1177/10943420221090256.

for more details.

## Test plan

Added 3 CP tests in `tests/context_parallel/test_cp_kda.py` (`-k tf32x3`: sequence cut, single long sequence, and state_v_first), exercising the tf32x3 affine chain via `FLACPContext.use_tf32x3_affine_chain` on 2/4 GPUs (they skip on single-GPU runners). All other CP/KDA tests run unaffected on the default path.

Neutral — this is an opt-in accuracy path, not a performance change. With the flag off (default), kernel launches omit `input_precision`, so NVIDIA behavior is bit-identical to before this PR; platforms without tf32 support (AMD, NPUs, NVIDIA GPUs with compute capability < 8.0) fall back to `"ieee"` explicitly.

## Checklist

### DIFF
--- a/ENVs.md
+++ b/ENVs.md
@@ -55,6 +55,7 @@ Each backend is gated by a single env var below.
 | `FLA_FLASH_KDA`                | unset (`1`) | `0` / `1`  | Enable the [FlashKDA](https://github.com/MoonshotAI/FlashKDA) CUTLASS forward for `chunk_kda` (inference only). Requires `flash_kda`.    |
 | `FLA_INTRACARD_CP`             | unset (`0`) | `0` / `1`  | Opt in to the intra-card context-parallel backend for shared delta-rule ops (`chunk_gated_delta_rule_fwd_h`). Inference + varlen only.  |
 | `FLA_INTRACARD_MAX_SPLITS`     | `32`        | int ≥ 1    | Max number of sub-sequences per original sequence used by the intra-card CP backend. Caps merge-chain depth to control precision loss.  |
+| `FLA_INTRACARD_TF32X3`         | `0`         | `0` / `1`  | Use tf32x3 for the affine-chain dots in the pre-scan/merge kernels of the intra-card CP backend. NVIDIA only.                            |
 
 ---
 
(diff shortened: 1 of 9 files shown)


zhiyuan1i:
LGTM overall after the rework commit — hanging the flag on `FLACPContext`, the explicit `ieee` fallback on platforms without tf32 support, and covering intracard all fit FLA's philosophy much better now.

PR #1195, opened 2026-08-29
[Fix] Keep wall_attn decode's pre-rescaled keys in bf16 to avoid fp16 overflow

zhiyuan1i:
## Summary

`test_wall_attn_cache.py::test_empty_sliding_window_cache_prefill[fp16]` (merged in #1189) fails with NaN on fp16 while bf16 passes.

Root cause: wall_attn decode pre-rescales keys into `k_tilde = k * exp2(R - P)` — the inverse of the accumulated decay inside a chunk, a positive exponent that grows with position. The final score stays in range because the query side carries the matching negative exponent, but the intermediate `k_tilde` itself exceeds the fp16 range after enough steps and becomes `inf` when cast to `k.dtype`.

Three host-side spots materialized this intermediate in the input dtype:

- `_decode_rebuild` (windowed decode): `build_wall_kv_cache` with default `out_dtype=k.dtype` — the CI failure.
- `_decode_incremental` prefill: same builder with default dtype; the cache then persists in that dtype (non-windowed chunks accumulate up to 64 steps, worse than the windowed suffix).
- `_decode_incremental` step path: `_rescale_new_keys` ends with `.to(k.dtype)`.

All three now store bf16 — the training-time default precision: same 8-bit exponent as fp32 (range 2^128, no overflow), half the memory of fp32 for the persistent decode cache. The decode kernel already upcasts to fp32 on load, so kernel behavior is unchanged.

## Test plan

`tests/layers/test_wall_attn_cache.py` on main currently fails at fp16 decode (NaN). With this change it passes on both dtypes; CI covers it.

Neutral. No kernel change; host-side cache dtype only.

## Breaking changes

None. The decode cache's internal k_tilde dtype becomes bf16; the cache is an internal runtime object, not a serialized format.

## Checklist

### DIFF
--- a/fla/layers/wall_attn.py
+++ b/fla/layers/wall_attn.py
@@ -241,7 +241,7 @@ class WallAttention(nn.Module):
 
         if prior is None:
             # Prefill: lean on the tested builder for the full pass.
-            k_tilde_new, _ = build_wall_kv_cache(k, p_new, C)
+            k_tilde_new, _ = build_wall_kv_cache(k, p_new, C, out_dtype=torch.float32)
         else:
             prior_len = prior[2].shape[1]
             k_tilde_new = self._rescale_new_keys(k, p_new, prior[2], prior_len, C, G)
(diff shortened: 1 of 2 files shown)


taking-lying-flat:
@zhiyuan1i 
The current H100 failure is caused by the mixed dot operands introduced by the bf16 cache: `b_q_til` is explicitly converted back to `b_q.dtype` (fp16), while `b_kt` is loaded from the new bf16 `k_tilde` cache.

A minimal fix is to keep the scaled query in fp32 until the dot, then convert it to the cache operand dtype:

```diff
-        b_q_til = (b_q.to(tl.float32) * exp2(b_pq - b_R)).to(b_q.dtype)
+        b_q_til = b_q.to(tl.float32) * exp2(b_pq - b_R)
...
-        b_s = tl.dot(b_q_til, b_kt) * scale * RCP_LN2
+        b_s = tl.dot(b_q_til.to(b_kt.dtype), b_kt) * scale * RCP_LN2
```

Casting `b_kt` back to fp16 would compile, but would defeat this PR: bf16 values above the fp16 range could become `inf` again. Casting the query side to `b_kt.dtype` keeps both dot operands compatible while preserving the wide-range key cache. It also retains the existing fp16×fp16 behavior for callers whose cache remains fp16.

I tested this patch on top of `801fb0c6` on an RTX A1000 (SM86):

- `test_empty_sliding_window_cache_prefill[float16]`: pass, decode max diff `0.000122`
- a matching non-windowed fp16 prefill + cached decode check: pass, max diff `0.000061`

H100 CI is still needed to validate SM90. It would also be useful to add a non-windowed fp16 regression, since the existing failing layer test covers `_decode_rebuild`, while this PR also changes the two `_decode_incremental` cache construction paths.

PR #1207, opened 2026-08-31
[Fix] Fix NPU triton_ascend kernel recompilation on variable-length sequences

ys2025-AI:
## Summary

Fix NPU `triton_ascend` kernel recompilation on variable-length (packed/varlen, `cu_seqlens`) sequences. Host-split grid offsets and per-sequence-length params were annotated `tl.constexpr`, causing Triton to recompile per distinct value. Those params are moved to runtime via `do_not_specialize`. All changes are in the `triton_ascend` backend — GPU paths (`triton`/`cuda`/`tilelang`) are untouched. `do_not_specialize` does not change what a kernel computes, only Triton's recompilation granularity.

| File | Fix | Effect |
|---|---|---|
| `fla/modules/backends/triton_ascend/causal_conv1d.py` | fwd/bwd coregrid kernels add `do_not_specialize` (fwd: `['B','T','NUM_CHKS','NUM_BLKS_D']`; bwd: `['B','T','NUM_BLKS_D','NUM_CHKS','NT_STRIDE','I_T_OFFSET']`) | conv1d per-T recompilation eliminated |
| `fla/ops/common/backends/triton_ascend/chunk_delta_h.py` | oneslab kernel: drop `NT: tl.constexpr`, compute `NT = T // BT` at runtime, add `"T"` to existing `do_not_specialize`; host no longer passes `NT`; updated the oneslab docstring whose "host NT constexpr" wording was invalidated by this change | T-multiple recompilation eliminated |
| `fla/ops/utils/backends/triton_ascend/solve_tril.py` | **all three** `merge_*`/`solve_*` kernels launched by `_launch_solve_tril_kernel` — `solve_tril_16x16_kernel_npu` (BT=16), `merge_16x16_to_32x32_inverse_kernel_npu` (BT=32), `merge_16x16_to_64x64_inverse_kernel_npu` (BT=64) — get `NT_OFFSET`/`BH_OFFSET` moved from `tl.constexpr` to runtime and added to `do_not_specialize` | solve_tril offset recompilation eliminated across all BT paths exercised by `test_solve_tril.py` (16/32/64) |

> **Deferred to a separate perf PR** (per review, https://github.com/fla-org/flash-linear-attention/pull/1207#discussion_r3893236300): the `chunk_o.py` `_MAX_BK 64→128` change is dropped from this PR. On current main it is a no-op on the gated-GDN hot path (`_get_dqkwg_full_tiles` still picks BK from `_FULL_BT_BK_CANDIDATES=(64,32,16)`, so the ceiling never exceeds 64; only the 3D fallback via `_get_bk` is affected, which gated models do not take by default), and it is unrelated to the recompilation fix. A follow-up perf PR will add `128` to `_FULL_BT_BK_CANDIDATES` and re-benchmark `bwd_dqkwg` on current main.

> **Note**: An earlier draft also patched `fla/ops/utils/backends/triton_ascend/cumsum.py` (`chunk_local_cumsum_scalar_kernel_npu` `NT_OFFSET`/`BH_OFFSET` constexpr→runtime). That is **no longer part of this PR** — upstream commit `865a52f` (PR #1061, "Perf: Rewrite Ascend local_cumsum") rewrote that kernel into a persistent form with `do_not_specialize=['T','B','task_num','num_core']`, which fully resolves the cumsum recompilation.

## Test plan

Dependent tests identified via `python scripts/find_dependent_tests.py <file>`:

- `causal_conv1d.py` → `tests/modules/test_conv.py`, `tests/context_parallel/test_cp_conv.py`
- `chunk_delta_h.py` → `tests/ops/test_gdn.py`, `tests/ops/test_gdn_kernels.py`, `tests/ops/test_solve_tril.py`, plus a wide fan-out of `tests/layers/*` and `tests/models/*`
- `solve_tril.py` → `tests/ops/test_solve_tril.py` (exercises BT=16/32/64 — now all three paths fixed)

These exercise the `triton_ascend` backend and require an Ascend NPU (`torch_npu`); not runnable in this dev environment. Verified locally in lieu of device runs:

- `python -m py_compile` on all 3 files — OK.
- Every argument listed in `do_not_specialize` is confirmed declared as a runtime param (not `tl.constexpr`) in its kernel signature, and (for oneslab) the host launcher no longer passes the removed `NT` kwarg — the `do_not_specialize` ↔ signature ↔ launcher chain is consistent.
- `inspect.getsourcefile` confirmed all kernels resolve to FLA `triton_ascend` backend files, not MindSpeed-overridden symbols.

To run on NPU: `pytest tests/ops/test_gdn.py tests/ops/test_gdn_kernels.py tests/ops/test_solve_tril.py tests/modules/test_conv.py`.

## Benchmark / NPU (recompilation-elimination only)

Hardware: Ascend A3, FLA 0.6.0, ms-swift, mcore-bridge. Method: cache-miss tracking via `JITFunction.run` monkey-patch over 20 steps.

The **recompile-miss** numbers below are the clean FLA-only signal — `do_not_specialize` only changes *when* Triton recompiles, so the miss-count drop is attributable entirely to this PR. The **wall-clock** numbers are labeled as "combined" because they also include a MindSpeed-side l2norm autotune fix landed in the same measurement window; do not attribute the full wall-clock speedup to FLA alone.

| Workload (Qwen3.5-4B, LongAlpaca 8192, cold cache) | Before | After |
|---|---|---|
| Total triton-miss / 20 steps (FLA-only signal) | 2647 | 203 (−92.3%) |
| Per-step miss (stable, step 8-18) | 0–274 (high variance) | 0 (all zero) |
| Spiked-step time (combined w/ MindSpeed l2norm fix) | ~120-129 s/it | 1-6 s/it |
| Total time / 20 steps (combined w/ MindSpeed l2norm fix) | 1526 s | 137 s |

Memory: identical before/after (compilation fix, no compute/memory change).

Correctness: loss trajectory consistent (1.47 → 0.98 both before/after), no NaN/divergence. `do_not_specialize` does not change what a kernel computes — only when Triton recompiles — so numerics are unchanged for these kernels.

Cross-model (0 miss after fix): Qwen3.5-4B (HV=32, K=V=128), Qwen3.6-27B (HV=48), Qwen3.6-35B-A3B (HV=32) — universal across dense + MoE. GPU paths unaffected (different backends).

### Breaking changes

None. All changes are internal to the `triton_ascend` backend — kernel decorators and runtime-vs-constexpr parameterization. No public API in `fla/layers/` or `fla/models/` is touched, and no launcher entry-point signature changes. Runtime vs constexpr specialization does not change a kernel's output, only Triton's recompilation granularity.

## Checklist

- [x] I have read [AGENTS.md](../AGENTS.md) and, where my change matches its scope, the relevant skill under [.agents/skills](../.agents/skills) — loaded `fla-ascend-performance`; `fla-mr-readiness` covers the NPU recompilation/`do_not_specialize` work.
- [x] Dependent tests pass locally or in CI, and new behavior is covered by tests where applicable — NPU tests require `torch_npu` and are not runnable in this env; `py_compile` + `do_not_specialize`↔signature↔launcher consistency check done. Pending: `pytest tests/ops/test_gdn.py tests/ops/test_gdn_kernels.py tests/ops/test_solve_tril.py tests/modules/test_conv.py` on Ascend.
- [x] Kernel changes include same-hardware before/after benchmark numbers, dense + varlen where applicable — see above (Ascend910_9382); varlen/packed is the primary scenario for the recompile fix.
- [x] This PR is **not** minor; functional perf/correctness-of-compilation fix.


fla/ops/utils/backends/triton_ascend/solve_tril.py
```diff
@@ -151,7 +151,7 @@ def merge_16x16_to_32x32_inverse_kernel_npu(
     tl.store(p_Ai_21, b_Ai_21.to(p_Ai_21.dtype.element_ty, fp_downcast_rounding='rtne'), boundary_check=(0, 1))
 
 
-@triton.jit(do_not_specialize=['T'])
+@triton.jit(do_not_specialize=['T', 'NT_OFFSET', 'BH_OFFSET'])
```
zheliuyu, line 154: Same defect left in two sibling kernels. The `NT_OFFSET/BH_OFFSET`: `tl.constexpr` pattern remains in `solve_tril_16x16_kernel_npu` and `merge_16x16_to_32x32_inverse_kernel_npu`, launched by the same host loop with the same varying offsets — they'll still recompile per offset once grid chunking kicks in. `test_solve_tril.py` exercises chunk sizes 16/32, so these are live paths. Either convert them here (same 2-line change) or state in the PR description that only BT=64 is fixed.

fla/ops/common/backends/triton_ascend/chunk_delta_h.py
zheliuyu: "host NT constexpr" — which is exactly what this PR removes. please update it in this commit.

zheliuyu:
By the way — what experiments did you actually run in `MindSpeed`? Does `MindSpeed` have `flash-linear-attention` integrated?

fla/ops/common/backends/triton_ascend/chunk_o.py
```diff
@@ -28,7 +28,7 @@
 _SAFETY_MARGIN = 0.80
 _FALLBACK_BK = 16
 _FALLBACK_BV = 16
-_MAX_BK = 64
+_MAX_BK = 128
```
sunyi0505: `_MAX_BK=128` is essentially a no‑op on the current main GDN hot path — fla/ops/common/backends/triton_ascend/chunk_o.py:31

The description states bwd_dqkwg −32%, caused by removing the NK dimension. In the current code, gated GDN uses the full‑BT 1D core‑grid:

`_get_dqkwg_full_tiles` uses `_MAX_BK` for k_cap, yet the maximum candidate value remains 64, so the full path still runs with BK=64. The NK grid only appears in the 3D fallback (`_get_bk` → `_launch_bwd_3d_kernel`), which gated models do not take by default.

The numbers are from FLA 0.6.0; the full‑BT path may not have existed at that time. This is also unrelated to "fixing recompilation". Two suggested options: add 128 into `_FULL_BT_BK_CANDIDATES` and re‑test on the current main branch; or remove this line from this PR.

fla/ops/utils/backends/triton_ascend/solve_tril.py
ys2025-AI: Good catch, agreed. Converted both siblings the same way as BT=64: solve_tril_16x16_kernel_npu (BT=16) and merge_16x16_to_32x32_inverse_kernel_npu (BT=32) now have NT_OFFSET/BH_OFFSET as runtime params and listed in do_not_specialize. All three kernels launched by _launch_solve_tril_kernel are consistent now, so the 16/32 paths test_solve_tril.py exercises no longer recompile per offset.

fla/ops/common/backends/triton_ascend/chunk_delta_h.py
ys2025-AI: thx, fixed it

fla/ops/common/backends/triton_ascend/chunk_o.py
ys2025-AI: @sunyi0505 Dropped the _MAX_BK 64→128 change from this PR.

ys2025-AI:
> By the way — what experiments did you actually run in `MindSpeed`? Does `MindSpeed` have `flash-linear-attention` integrated?

@zheliuyu To clarify the setup: the workload in the benchmark table is Qwen3.5-4B LoRA fine-tuning run through **ms-swift's megatron backend** (LongAlpaca, seqlen 8192). FLA is *not* integrated into MindSpeed as a library — the model's GDN layers call FLA's `triton_ascend` backend kernels directly (`causal_conv1d`, `chunk_delta_rule`, `solve_tril`), while MindSpeed ships its own triton kernels for other ops. The two coexist in the same training stack.

The "MindSpeed l2norm fix" refers to a separate issue surfaced during this work: the `l2norm` op in the forward path turned out to use a MindSpeed-provided triton kernel (not an FLA kernel), and a MindSpeed-side autotune/BT adjustment was applied to it in the same measurement window. That's why the wall-clock numbers are labeled "combined" — this PR's FLA-only contribution is the recompile-miss reduction (the clean signal, since `do_not_specialize` only changes *when* Triton recompiles, not what it computes). The MindSpeed-side l2norm change is external to FLA and not part of this PR.

zheliuyu:
LGTM. @zhiyuan1i

zheliuyu:
@ys2025-AI If feasible, it might be worth integrating `flash-linear-attention` directly into `MindSpeed`. This would allow us to maintain only one set of `GDN/KDA` code.

ys2025-AI:
> @ys2025-AI If feasible, it might be worth integrating `flash-linear-attention` directly into `MindSpeed`. This would allow us to maintain only one set of `GDN/KDA` code.

@zheliuyu Integrating into MindSpeed-Ops may be a good choice.

zhiyuan1i (changes requested):
Style issue:

**P1**: `chunk_delta_h.py:750` — `kernel_kwargs={**kwargs}` is a leftover dict copy after removing the `NT` key. `_launch_core_grid` only reads from it, so pass `kernel_kwargs=kwargs` directly.

**P2**: `chunk_delta_h.py:544` — docstring mentions `(T // BT)` which restates line 555. Keep "NT computed at runtime" but drop the formula.

zhiyuan1i (approved):

PR #1208, opened 2026-08-31
[Models] Add MomentumDeltaNet implementation from arXiv:2605.05838 (ICML2026)

bhaochen:
## Summary

Add MomentumDeltaNet (MDN) as proposed in **MDN: Parallelizing Stepwise Momentum for Delta Linear Attention** ([arXiv:2605.05838](https://arxiv.org/abs/2605.05838), ICML 2026).

This slice intentionally follows **reference implementation first, Triton kernels as follow-up**. The pure-PyTorch implementation covers the full stepwise momentum formulation with dual state `[S, M]`:

- `w_t = -(v_t - p_t @ S_{t-1})`
- `M_t = mu_t * M_{t-1} + (eta_t * k_t)^T @ w_t`
- `S_t = alpha_t * S_{t-1} - beta_t * M_t`

The update also:

- threads final-state and initial-state gradients through both chunk and fused-recurrent backward paths;
- fixes fused-recurrent gradient arity, l2-normalization wiring, GVA head dimensions, `use_beta=False`, and padded chunk final states;
- removes the duplicate degenerate momentum operators and reuses the existing `delta_rule` operators for degenerate tests;
- rejects `attention_mask` and `cu_seqlens` explicitly while variable-length momentum support is pending;
- adds full momentum chunk/recurrent/reference forward and backward parity coverage, including initial states and final-state upstream gradients.

## Test plan

- `python -m compileall -q fla/layers/momentum_deltanet.py fla/ops/momentum_delta_rule tests/ops/test_momentum_delta.py` — passed
- `git diff --check` — passed
- pre-commit checks (ruff, autopep8, conflict/private-key checks, Triton forbidden-symbol check) — passed
- `pytest -q tests/ops/test_momentum_delta.py -x` — not runnable in this checkout because `torch` is not installed; CI is expected to run the full test file on the supported runtime

## Benchmark / NCU

N/A for this update: it contains no Triton kernel changes. The intended follow-up is to add Triton kernels after the reference path is validated.

## Breaking changes

- The duplicate `chunk_mode_rule` and `fused_recurrent_mode_rule` exports are removed; callers needing degenerate DeltaRule semantics should use the existing `chunk_delta_rule` and `fused_recurrent_delta_rule` operators.
- `MomentumDeltaNet` now raises `NotImplementedError` for `attention_mask` and `cu_seqlens` instead of silently selecting a different algorithm.

## Checklist

- [x] Dependent tests pass locally or in CI; CI is expected to run them, while local execution is blocked because `torch` is unavailable in this checkout.
- [x] Kernel changes include same-hardware before/after benchmark numbers, dense + varlen where applicable (N/A: no kernel code changed in this update).

N/A


zhiyuan1i (changes requested):
Three P0 issues that break training on this PR:

1. **Backward gradient count mismatch**: `chunk_momentum_delta_rule` and `fused_recurrent_momentum_delta_rule` return 11 gradients (copied from `chunk_delta_rule`'s template which has `scale`), but their forwards have 10 and 8 inputs respectively. First backward call crashes with "incorrect number of gradients". `dh0` also lands in the `output_final_state` slot, so `initial_state` gets None.

2. **fused_recurrent bwd passes raw `v` instead of `u`**: the kernel expects `u = v - k·h` (what the fwd saves), but receives the original `v`. All of dk/db/dq are wrong. The layer routes q_len≤64 through this path.

3. **Default `qk_norm='l2'` never takes effect**: the op call doesn't pass `use_qk_l2norm_in_kernel`, so q/k are unnormalized under default config — silently diverging from DeltaNet family math.

Also: no tests at all. Any naive fwd/bwd parity check would catch all three. And there's no momentum coefficient anywhere — if the paper's momentum cancels algebraically, that needs a comment; otherwise the implementation doesn't match the paper.

Please fix the gradient returns, pass `u` to bwd, wire up l2norm, and add fwd/bwd + varlen numerical tests.

zhiyuan1i (changes requested):
Style issues (in addition to the P0 correctness issues from the previous review):

1. **P1**: `chunk.py:186,198` — `chunk_indices` is computed and `save_for_backward`'d but never used: fwd/bwd hardcode `chunk_indices=None` everywhere (7 sites). Dead code chain, and each sub-kernel recomputes indices.

2. **P1**: Naming `chunk_mode_rule`/`fused_recurrent_mode_rule` is uninformative and misleading. Should be `chunk_momentum_delta_rule`/`fused_recurrent_momentum_delta_rule` per repo convention.

3. **P1**: Comments like "(fixes P0 #3)" and "Regression for P0 #1/#2/#3" reference review round numbers that are meaningless after merge. Remove the numbers, keep self-contained descriptions.

4. **P1**: `test_momentum_delta.py:17,39` — `scale` parametrize is never used in the test body (op hardcodes `1/sqrt(D)`), dead parameter.

5. **P1**: `fused_recurrent.py:55` — `w, u, A = prepare_wy_repr_fwd(...)` computes `w` and `A` then immediately discards them. Dead computation.

6. **P2**: `momentum_deltanet.py` `__init__` missing `assert mode in ['chunk', 'fused_recurrent']` (`delta_net.py:126` has it).

7. **P2**: `use_qk_l2norm` temp variable used only twice — inline at call site like `delta_net.py:256`.

8. **P2**: Header comment blocks (13 lines in `chunk.py`, 9 lines in `momentum_deltanet.py`) too long — compress to 3 lines max.

9. **P2**: `__init__.py` uses double quotes, repo convention is single quotes.

Also the biggest concern: this is named MomentumDeltaNet but the implementation is standard delta rule (mu→0, alpha=1 degenerate). If the paper's momentum cancels algebraically, that needs a prominent comment; otherwise the naming doesn't match the semantics.

zhiyuan1i (changes requested):
Thanks for the updates — the v/u mix-up in the fused backward and the l2norm wiring are properly fixed now. However, the gradient-count class of bug is still present, and there are new blocking issues.

**Blocking (correctness)**

1. Gradient count mismatch, same class as before: `FusedRecurrentMomentumDeltaRuleFunction.forward` takes 15 inputs but `backward` returns 16 values (`fla/ops/momentum_delta_rule/fused_recurrent.py:170`) — a trailing `None` copied over from the chunk variant (which has 16 inputs, the extra one being `chunk_size`). This raises `RuntimeError: ... returned an incorrect number of gradients (expected 15, got 16)` when training with `mode='fused_recurrent'`. That path has zero test coverage, which is why CI stays green. Please drop the extra `None` and add a fused full-path backward test.

2. The full-path backward silently drops gradients (`chunk.py:160,180-183`; same in `fused_recurrent.py:158,168-170`):
   - `dst`/`dmt` (upstream grads of the final states) are ignored — the comment says "ignore for now". With `output_final_state=True` and `ht` participating in the loss, every input gradient is silently wrong.
   - `ds0 = torch.zeros_like(initial_S)`: the true gradient w.r.t. the initial state is never computed.
   Please either wire `initial_S`/`initial_M` into the recompute graph and thread `dst`/`dmt` through, or raise explicitly when they are required. Silent zeros are not acceptable.

3. varlen/attention_mask silently degrades semantics (`momentum_deltanet.py:298`): with a mask or `cu_seqlens`, the layer falls back to the degenerate delta rule with a pre-scaled beta — that is neither the paper's momentum nor standard DeltaNet, and there is no warning. Please either raise `NotImplementedError`, or use the un-scaled beta with a warning.

**Structure / style**

4. The two degenerate ops (`chunk_mode_rule` / `fused_recurrent_mode_rule`, ~250 lines) are line-by-line copies of the existing `chunk_delta_rule` / `fused_recurrent_delta_rule` in `fla/ops/delta_rule/`, differing only in not exposing `scale` — whose default already equals `1/sqrt(K)`. Please delete them and have the layer fallback import the existing ops directly. This also resolves the earlier naming comment.

5. Tests: all 14 tests exercise the degenerate path; the full momentum path has no numerical coverage. Please add fwd/bwd parity between the chunk and recurrent reference implementations, covering `initial_state` + `output_final_state` + `ht` gradients. (This is how item 1 slipped through.)

6. Dead code and stale comments: the three `pass` blocks in the layer fallback (`momentum_deltanet.py:334-359`), the always-True `_use_full_momentum` switch, unused `ctx.final_S`/`final_M`, the `+ torch.zeros_like(...)` no-op in `naive.py:174`, comments referencing the deleted Triton path ("as Triton path did", "avoid H100 TMA hang"), and the mixed-language "0 Triton新增" comment.

One thing for maintainers to confirm (not a correctness blocker): the default training path is now the pure-PyTorch naive implementation after the new Triton kernels were removed in dff3bcb. Please state explicitly in the PR description that the intended slicing is "reference implementation first, Triton kernels as follow-up".

bhaochen:
Fixed in c7f519e6. The full-path custom backward now passes `(o, final_S, final_M)` together with the positional upstream gradients `(do, dst, dmt)` directly to `torch.autograd.grad`, preserving output-gradient alignment and propagating final/initial state gradients without manually constructing a potentially misaligned scalar loss. H100 test-ops has been retriggered.

bhaochen:
Follow-up fix in 8a8a1348: the parity test now explicitly sets `use_p_times_alpha=False` for the custom chunk/recurrent paths, matching the naive reference inputs. The previous failure was a test configuration mismatch, not an implementation numerical divergence. CI has been retriggered.

bhaochen:
Follow-up fix in 2063f09d: when `output_final_state=False` (the layer default), PyTorch supplies `None` for `dst/dmt`. The custom backward now replaces those missing non-scalar gradients with zero tensors before calling `torch.autograd.grad`, while retaining real final-state gradients when outputs are consumed. This addresses the latest `grad can be implicitly created only for scalar outputs` failure.

bhaochen:
Follow-up fix in aed51365: full-path l2 backward now passes contiguous `dq`, `dk`, and `dp` tensors to `l2norm_bwd`, whose implementation uses `view()`. The latest run passed the full momentum parity test and failed only in the layer l2 backward due to a non-contiguous autograd gradient. CI has been retriggered.

zhiyuan1i (approved):
All RC items addressed; remaining dead code and the redundant parity test were removed in 20ef8776. CI green. LGTM.

PR #1210, opened 2026-08-31
[Perf] Replace split Ascend KDA bwd_intra kernels with a single persistent kernel

qianjinqiu:
## Summary

Replaces the three Ascend bwd_intra kernels (`dq_db` + `dkt_future` + `dk_dg`) with one
persistent kernel.

- **Fusion**: drops the `dkt` partial `[T, HV, K]` fp32 GM round-trip between `dq_db` and
  `dk_dg`; 3 launches → 1.
- **Bare pointers instead of `tl.make_block_ptr`** — largest win. With identical
  BK=256/BC=32 tiling, the split `dk_dg` alone costs 45.3ms at D256; this whole kernel
  (all three stages) costs 14.3ms.
- **`BK = next_pow2(K)` single-block tiles**: no NK fragmentation, no q/k/g re-reads per
  K fragment.
- **Persistent 1D core-grid** with a grid-stride task loop.
- **BC gated by padded BK against the 192KB UB budget** (sg1: 32 if BK≤256, 16 if BK≤512;
  sg0: 16 if BK≤128, 8 if BK≤256) — covers non-pow2 K (e.g. 192); out-of-envelope K
  raises `NotImplementedError` instead of failing inside BiShengIR.
- **Outputs stored via `element_ty`** — no fp32→bf16→fp16 double rounding on fp16.

## Test plan

- Ascend 910B3, CANN 9.1.0, triton-ascend 3.2.2.
- `pytest tests/ops/test_kda.py` — 83 passed, 6 skipped (dense + varlen, both `safe_gate`
  values; includes `test_triton_ascend_backend_routing`, no silent fallback).
- Numerical cross-check vs the split kernels on identical inputs: official 6 shapes +
  varlen × both `safe_gate` paths, plus K=100/192/512 and fp16 probes → dq/dg bitwise
  equal, dk ≤ 2 bf16 ulp, db fp32 rounding noise.
- Dependent tests per `python scripts/find_dependent_tests.py fla/ops/kda/backends/triton_ascend/chunk_intra.py`
  (tests/ops/test_kda.py, test_gdn_kernels.py, test_precond_kda.py, context-parallel and
  model suites) — covered by CI.

torch_npu profiler, device-side µs/call, 12-sample median, same inputs, dense bf16,
official `benchmarks/ops/registry.py` shapes:

| shape | split (sum of 3)/ms | single kernel/ms | speedup |
| --- | --- | --- | --- |
| B1_T8192_H96_D128 | 76.6 | 19.0 | 4.03× |
| B2_T16384_H16_D128 | 51.1 | 12.9 | 3.96× |
| B4_T2048_H16_D128 | 12.8 | 3.3 | 3.88× |
| B4_T4096_H64_D128 | 102.1 | 25.3 | 4.04× |
| B8_T2048_H32_D256 | 102.2 | 14.3 | 7.17× |
| B8_T1024_H8_D64 | 3.3 | 1.8 | 1.81× |
| varlen B1 H8 D128 (cu=[1,1234,3456,5678,8000]) | 6.31 | 1.80 | 3.51× |
| safe_gate=False, B8_T2048_H32_D256 | 234.1 | 91.8 | 2.55× |

Measured on 910B3; A2 CI numbers to be attached once available.

## Breaking changes

None reachable through the public API: `chunk_kda` asserts `K <= 256`
(fla/ops/kda/chunk.py:409) before any backend dispatch, and every officially reachable
configuration compiles and runs as before. The `NotImplementedError` guards are
defense-in-depth for direct callers of `chunk_kda_bwd_intra` (K > 256 with
`safe_gate=False`); the mainline fallback (`FLA_DISABLE_BACKEND_DISPATCH=1`) is verified
working on device for that case.

## Checklist

### DIFF
--- a/fla/ops/kda/backends/triton_ascend/chunk_intra.py
+++ b/fla/ops/kda/backends/triton_ascend/chunk_intra.py
@@ -680,413 +680,232 @@ def chunk_kda_fwd_intra_npu(
     return w, u, qg, kg, Aqk, Akk
 
 
-# Split at debug_barrier: dq/db half has lower peak live set than dk/dg half.
-# BC=32 cuts NC 4→2 (chunk_size=64) so the future-subchunk loop in dkt_future
-# drops from 6 pair-blocks to 1; Cube tiles are 32×32 instead of 16×16.
-# SAFE_GATE diag keeps many concurrent BC×BK fp32 tiles live. mem_mult=9 lets
-# host pick BK=128 (NK 4→2 on D256) while dq_db (past+diag) still fits 192KB UB.
-# dk_dg has a smaller live set and can use BK=256 (NK=1). dkt_future compiles at
-# BK=256 but faults at runtime (UB address OOB), so it stays on BK=128.
-_BWD_INTRA_BC = 32
-_BWD_INTRA_DQ_MEM_MULT = 9.0
-_BWD_INTRA_DK_MEM_MULT = 4.5
-_MAX_BK_DQ = 128
-_MAX_BK_DK = 256
-
-
 def get_npu_properties():
     device = torch.npu.current_device()
     return driver.active.utils.get_device_properties(device)
 
 
-def _get_bwd_intra_bk(K: int, BC: int = _BWD_INTRA_BC, *, element_size: int = 2) -> int:
-    # fp32 I/O roughly doubles compile-time element-tile UB vs bf16 in dq_db SAFE_GATE.
-    max_bk = _MAX_BK_DQ if element_size <= 2 else min(_MAX_BK_DQ, 32)
-    return compute_row_tile_block_size(
-        BC,
-        K,
-        _BWD_INTRA_DQ_MEM_MULT,
-        tiling_row=False,
-        safety_margin=_SAFETY_MARGIN,
-        fallback=_FALLBACK_BK,
-        min_block=16,
-        max_block=min(max_bk, triton.next_power_of_2(K)),
-    )
-
-
-def _get_bwd_intra_bk_dk(K: int, BC: int = _BWD_INTRA_BC, *, element_size: int = 2) -> int:
-    """Wider K tile for dk_dg (no past-subchunk live set). dkt_future cannot use this."""
-    max_bk = _MAX_BK_DK if element_size <= 2 else min(_MAX_BK_DK, 32)
-    return compute_row_tile_block_size(
-        BC,
-        K,
-        _BWD_INTRA_DK_MEM_MULT,
-        tiling_row=False,
-        safety_margin=_SAFETY_MARGIN,
-        fallback=_FALLBACK_BK,
-        min_block=16,
-        max_block=min(max_bk, triton.next_power_of_2(K)),
-    )
-
-
-def _launch_bwd_intra_core_grid(kernel, *, task_num: int, kernel_kwargs: dict) -> None:
-    num_core = get_npu_properties()["num_aicore"]
-    kernel[(num_core,)](task_num=task_num, num_core=num_core, **kernel_kwargs)
-
-
-def _gk_npu_arg(g: torch.Tensor, HV: int) -> tuple[torch.Tensor, bool]:
-    """Transpose g to [B, HV, T, K] for stride-1 row loads along T."""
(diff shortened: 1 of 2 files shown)


fla/ops/kda/backends/triton_ascend/chunk_intra.py
```diff
@@ -677,413 +677,219 @@ def chunk_kda_fwd_intra_npu(
...
         i_b, i_hv = i_bh // HV, i_bh % HV
         i_h = i_hv // (HV // H)
-        i_k, i_i = i_kc // NC, i_kc % NC
-        T_seq = T
-        if IS_VARLEN:
+
+        if cu_seqlens is not None:
             i_n, i_t = tl.load(chunk_indices + i_t * 2).to(tl.int32), tl.load(chunk_indices + i_t * 2 + 1).to(tl.int32)
-            bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
+            bos, eos = tl.load(cu_seqlens + i_n).to(tl.int32), tl.load(cu_seqlens + i_n + 1).to(tl.int32)
```
sunyi0505: The upper limit of int32 is approximately 2.15e9, and the official benchmark is already pushing the limit with its own number of heads:

Image

The maximum packed length of test_kda.py is 4096, and passing CI runs cannot prevent the issue. The symptom is silent read/write errors rather than hard crashes.

Suggestion: Apply .to(tl.int64) to bos/eos. Extend all base addresses and row offsets to 64‑bit before addition, following the same approach as (i_b * T).to(tl.int64) in chunk_bwd.py.

fla/ops/kda/backends/triton_ascend/chunk_intra.py
```diff
@@ -1105,112 +911,44 @@ def chunk_kda_bwd_intra_npu(
...
+    # follow the padded BK (non-pow2 K like 192 lands in the BK=256 tier).
+    # SAFE_GATE dot path uses 32x32 cube tiles; the pairwise per-column diag
+    # compiles only at BC<=16, and only at BC=8 once BK reaches 256.
+    if safe_gate:
+        if BK <= 256:
+            BC = min(32, BT)
+        elif BK <= 512:
+            BC = min(16, BT)
+        else:
+            raise NotImplementedError(
```
sunyi0505: It is incorrect to raise NotImplementedError in launch when safe_gate=False and K>256 (or safe_gate=True and K>512). This will result in a very poor user experience. A default path should be designed, and the system should fall back to the default path when the conditions are not met.

zhiyuan1i (approved):

zhiyuan1i (commented):
One nit: `bos = (i_b * T).to(tl.int64)` computes the product in int32 before casting — the old code was `tl.cast(i_b, tl.int64) * T` which is correct. B·T > 2^31 silently overflows. Same for `task_num = NT_TOTAL * NC * B * HV` computed in int32 inside the kernel. Please cast before multiply.

zhiyuan1i (changes requested):
Style issues to fix:

1. **P1**: `chunk_intra.py:835` — `NC_F = min(NC, tl.cdiv(T - i_t * BT, BC))` is identical to `NC_LOC` on line 754. Reuse the same variable.

2. `chunk_intra.py:755` — `for i_i in range(i_i0, min(i_i0+1, NC_LOC))` is a loop that runs at most once. Use `if i_i0 < NC_LOC` instead.

3. `__init__.py:43` — `1 << max(K - 1, 0).bit_length()` is inconsistent with `triton.next_power_of_2(K)` used elsewhere in the same file. Pick one.

4. `chunk_intra.py:762/823/896` — section comments like `# ---- head loads ----` carry no information. Remove or make them meaningful.

5. `chunk_intra.py:932` — `getattr(f, '__wrapped__', f)` needs a one-line comment explaining why (skip double-dispatch when `_DISPATCH_DISABLED`).

6. Two-line comments at `__init__.py:39-40` and `chunk_intra.py:756-757` can each be collapsed to one line.

zhiyuan1i (approved):
Overall, great job, LGTM

PR #1213, opened 2026-09-01
[Ops] Add Triton-Ascend compatibility for GDN-2

hazelduan:
## Summary

Add Triton-Ascend support for GDN-2 chunk training and prefill with dense and packed sequences in FP16, BF16 and FP32.

- Dispatch forward intra-chunk and fused WY backward computation to Ascend kernels, reusing existing shared kernels.
- Handle UB capacity, launch splitting and 64-bit token addressing.
- Extend reference-based output/gradient tests with dispatch assertions, launch boundaries, unequal head dimensions and opt-in 32K coverage. Include GDN-2 in A2 correctness and benchmark CI.

Shared state kernels, public APIs and numerical tolerances are unchanged.

## Test plan

Tested on Ascend910B3 (CANN 9.1-beta.1, PyTorch 2.9.0, torch_npu 2.9.post2, Triton-Ascend 3.2.2; 192 KiB UB):

- Full retained GDN-2 suite: **58 passed, 1 skipped**; only opt-in 32K is skipped.
- Opt-in B1/T32768/H1/K32/V32 FP16 reference test: **1 passed**, checking output, final state and all gradients.
- Both existing KDA D100/FP16 cases that previously failed with UB overflow (dense B3/T1024/H4 and ragged H4): **2 passed**, including gradient comparisons.
- Ruff 0.14.10, autopep8 2.3.2, copyright headers and diff checks pass.

The three added dense state-layout/recomputation combinations are deferred following review; existing layout and ragged tests remain. The shared state kernel is unchanged from the PR base. The historical NaN in the deferred V-first/recompute case remains unresolved. Existing references and tolerances are retained.

**Scope: compatibility enablement; no performance improvement claimed.** The figures compare historical PR revisions [`6e640ba9`](https://github.com/fla-org/flash-linear-attention/actions/runs/34242630212) with [`8e6d4822`](https://github.com/fla-org/flash-linear-attention/actions/runs/34330124617), both already containing the Ascend backend; these figures predate the review rollback. Each artifact contains all **7 shapes × 2 modes**.

Ascend910B3, CANN 9.1, PyTorch 2.9; dense BF16, K=V=D, QK normalization enabled, default state layout and recomputation. GDN-2 runs on device 2 after fused_attnres. The unified runner uses three warmup calls, then 25 ms warmup / 100 ms repetition windows. These are separate CI runner instances: the identical p20/p50/p80 values do not establish variability or a causal performance change.

GDN-2 forward latency across the six default benchmark shapes

GDN-2 forward and backward latency across the six default benchmark shapes

GDN-2 32K sequence latency comparison

The default-shape latencies are 1.5–2.0% higher in the later snapshot; 32K is 1.4–1.9% lower, at **135.70 ms forward / 141.51 ms forward + backward**. These are descriptive CI measurements, not a controlled regression verdict.

<details>
<summary>Reproduction and same-card shared-kernel sample</summary>

```bash
export ASCEND_RT_VISIBLE_DEVICES=2
python -m pytest tests/ops/test_gdn2.py -v
FLA_TEST_GDN2_32K=1 python -m pytest tests/ops/test_gdn2.py -k 32k -v

export FLA_BENCH_OP_WARMUP_ITERS=3 FLA_BENCH_WARMUP_MS=25 FLA_BENCH_REP_MS=100
python -m benchmarks.ops.run --op chunk_gdn2 --base '' --json benchmark_chunk_gdn2.json
python -m benchmarks.ops.run --op chunk_gdn2 --base '' \
  --custom-shapes '{"gdn2_32k":{"B":1,"T":32768,"H":1,"D":32}}' \
  --json benchmark_chunk_gdn2_32k.json
```

An earlier same-card comparison of the shared `chunk_gated_delta_rule_bwd_dhu_npu` wrapper, `6e640ba9` → `8e6d4822`, measured:

| Layout | Before (ms) | After (ms) |
| ------ | ----------: | ---------: |
| Dense  |    0.314535 |   0.315253 |
| Packed |    0.352238 |   0.352313 |

Ascend910B3, CANN 9.1-beta.1 / torch_npu 2.9.post2 / triton-ascend 3.2.2; B1/T512/H2/K64/V64 FP16, default state layout, zero per-key log gates, inputs scaled by 0.01, packed boundaries `[0,63,300,512]`. Synchronized host-wrapper timing, three warmups, seven samples of 200 calls, alternating revisions. This local stack used the 64 KiB UB-manager fallback; the sample covers the shared wrapper only. Full-operator packed performance and peak memory are not measured here.

</details>

## Breaking changes

None. Public API, checkpoint compatibility, numerical tolerances and the `chunk_size=64` contract are unchanged.

## Checklist


tests/ops/test_gdn2.py
```diff
@@ -7,22 +7,203 @@
...
 from fla.ops.kda.gate import naive_kda_gate, naive_kda_lowerbound_gate
-from fla.utils import assert_close, device
+from fla.utils import IS_AMD, IS_NPU, IS_NVIDIA, assert_close, device
+
+_requires_accelerator = pytest.mark.skipif(
+    not (IS_NVIDIA or IS_AMD or IS_NPU),
+    reason="CUDA/ROCm or Ascend NPU required",
+)
+
+_LONG_SEQUENCE = os.environ.get("FLA_GDN2_LONG_SEQUENCE") == "1"
```
zheliuyu: Just wondering — is the experimental code in `test_gdn2` still not cleaned up? I wouldn't expect the test file to require this many changes.

sunyi0505:
nit: UT only tests small shapes. It is recommended to add the benchmark to ascend‑a2‑benchmark‑ci.yml to guard functionality and performance under large‑shape scenarios.

zhiyuan1i (changes requested):
Style issues:

1. **P1**: `chunk_intra.py:394,398` — `b_Ai22_c = b_Ai22 + 0.0` and `b_Ai33_c` are dead code (never used).

2. **P1**: `chunk_bwd.py:163-165,291` and `chunk_intra.py:392-402` — multiple `x_c = x + 0.0` copies with no comment. If these are CANN compiler workarounds, each needs a one-line comment saying what it avoids. `_c/_c2/_c3` naming is also uninformative.

3. **P1**: `chunk_bwd.py:54` — `_launch_dA_finalize` is called only once (line 523). Inline the 25-line launch branch into `chunk_gdn2_bwd_wy_dqkg_fused_npu`.

tests/ops/test_gdn2.py
hazelduan: Thanks for pointing this out. Initial diff mixed temporary NPU validation code into the canonical GDN2 UT and was broader than necessary. The current change in tests/ops/test_gdn2.py only replaces the CUDA-only skip condition with the repository-level accelerator guard covering NVIDIA, AMD, and Ascend NPU.

zheliuyu:
Looks like `test_gdn2` is failing — please check the cause.

hazelduan:
> Looks like `test_gdn2` is failing — please check the cause.

The failure came from unmasked padded lanes in the shared （common）Ascend backward kernel; masking them before reduction fixes the NaN gradients.

zheliuyu:
Thanks for the revisions.

One change I’d like in this PR: the intra verifier is too open-ended on `K`. The grouped kernel uses `BK = next_power_of_2(K)` with no K-slab and no UB cap, so a large head dim (512/1024) will fail at compile time instead of the backend returning `(False, reason)` and letting dispatch fall back. KDA bwd already rejects with `next_power_of_2(K) > limit`. Please add a real K/UB bound in `chunk_gdn2_fwd_intra_verifier` (and the bwd verifier if that path has the same exposure). The K=256 test doesn’t cover this.

Fine as follow-ups, no need to block: WY recompute and `chunk_gdn2_bwd_intra` still use the CUDA Triton autotune (`num_warps` / `num_stages`).

hazelduan:
> Thanks for the revisions.
> 
> One change I’d like in this PR: the intra verifier is too open-ended on `K`. The grouped kernel uses `BK = next_power_of_2(K)` with no K-slab and no UB cap, so a large head dim (512/1024) will fail at compile time instead of the backend returning `(False, reason)` and letting dispatch fall back. KDA bwd already rejects with `next_power_of_2(K) > limit`. Please add a real K/UB bound in `chunk_gdn2_fwd_intra_verifier` (and the bwd verifier if that path has the same exposure). The K=256 test doesn’t cover this.
> 
> Fine as follow-ups, no need to block: WY recompute and `chunk_gdn2_bwd_intra` still use the CUDA Triton autotune (`num_warps` / `num_stages`).

Good catch. I added a next_power_of_2(K) <= 256 UB bound to the forward intra verifier and a K=257 boundary test, so oversized inputs fall back before compilation. I left the backward verifier unchanged because that path tiles K across K_OFFSET slabs and does not have the same single-slab UB exposure.

zheliuyu:
Thanks. A hard `next_power_of_2(K) <= 256` cap is a bit rigid — deriving the bound from `get_ub_capacity` (or `ascend_ub_manager`) is the more robust end state. But, it's safe enough on A2/A3 for this PR.

We can stress-test larger K / tighter UB on the follow-up work in https://github.com/Ginray/flash-linear-attention/pull/1. @Ginray

No other comments from me. @zhiyuan1i could you take another look and confirm whether this is good to merge? Thanks.

zhiyuan1i (approved):
One concern: the reused shared `recompute_w_u_fwd_gdn2` uses `b_A` as `tl.dot`'s left operand in both the u and w loops with no reload (`fla/ops/gdn2/wy_fast.py:90,101,141`) — Ascend `tl.dot` may clobber the left operand (#1113; the KDA NPU variant of this same kernel reloads for exactly this), so is this reuse safe, or should it route to an NPU-safe variant?

hazelduan:
> One concern: the reused shared `recompute_w_u_fwd_gdn2` uses `b_A` as `tl.dot`'s left operand in both the u and w loops with no reload (`fla/ops/gdn2/wy_fast.py:90,101,141`) — Ascend `tl.dot` may clobber the left operand (#1113; the KDA NPU variant of this same kernel reloads for exactly this), so is this reuse safe, or should it route to an NPU-safe variant?

Thanks for pointing this out! That’s a valid concern: this PR reuses the existing CUDA-oriented WY kernel without modifying it. I ran additional comparisons against a PyTorch reference using triton-ascend 3.2.2, covering dense/packed sequences, FP16/BF16/FP32, and multiple K/V tiles. The original kernel passed all tested cases without reloading A, so I couldn’t reproduce the clobber issue on this stack. However, a follow-up performance PR will add an Ascend-specific WY implementation that reloads A before each tl.dot, also guarding against this potential clobber risk. Would you be comfortable moving forward with this compatibility PR, given the passing checks on triton-ascend 3.2.2? We can include the defensive reloads in the follow-up Ascend WY optimization.

PR #1225, opened 2026-09-07
[Perf] Optimize Ascend GDN recompute_w_u_fwd with block pointers and independent BK/BV tiling

sunyi0505:
## Summary

Speed up the Ascend NPU path of GDN WY recompute_w_u_fwd in fla/ops/gated_delta_rule/backends/triton_ascend/wy_fast.py. The kernel still recomputes u = A @ (v * β) and w = A @ (k * β * g) per chunk, but the launch and memory access now match the existing backward kernels.

Replace scalar pointer + mask loads/stores with tl.make_block_ptr and boundary_check, so K/V slabs use contiguous block traffic instead of per-element addressing.
Drop the hardcoded BK = BV = 64. Peak UB is max(u-slab, w-slab) rather than the sum, so BK and BV are tiled independently under a ~3.5× slab budget (cap 128). A small host search over {64, 128, 32, 16, 8} picks the pair that minimizes cdiv(V, BV) + cdiv(K, BK).
Stop unconditionally transpose(1, 2).contiguous() on beta/g. Reuse _beta_npu_arg / _g_npu_arg and pass BETA_T_CONTIG / G_T_CONTIG, so HV=1 keeps the original layout and avoids an extra host copy.
Keep the Ascend tl.dot left-operand reload of A on every V/K tile. Cast the scaled V/K slabs to the source dtype before the matmul.
Public GDN API, checkpoint semantics, and numerical contract are unchanged.

## Test plan

hardware: Atlas 800T A3(X86)
image

image

## Breaking changes

None

## Checklist

### DIFF
--- a/fla/ops/gated_delta_rule/backends/triton_ascend/wy_fast.py
+++ b/fla/ops/gated_delta_rule/backends/triton_ascend/wy_fast.py
@@ -39,6 +39,12 @@ _FALLBACK_TILE = 8
 _MAX_TILE_BWD = 128
 _MAX_TILE_BWD_KV = 256
 
+# recompute_w_u_fwd: peak UB is max(u-slab, w-slab), not sum — tile BK/BV independently.
+# Each slab is ~3.5× BT×{BV or BK} + BT×BT (no gk/qg vs KDA).
+_RECOMPUTE_FWD_MEM_MULT = 3.5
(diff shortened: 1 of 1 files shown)


sunyi0505:

zhiyuan1i (approved):
LGTM

PR #1235, opened 2026-09-09
[Fix] Eliminate per-call reference cycle in launch_grid_chunked

Ginray:
## Summary

`launch_grid_chunked` built a nested `_recurse` closure that referenced itself by name, creating a self-referential reference cycle on every call. The cycle holds `kernel_kwargs` and the kernel object alive until the generational GC runs. Under benchmark loops (`do_bench` fires hundreds of op calls inside one rep window), this garbage accumulates device tensors at ~0.65 GiB per call on large shapes — NPU benchmark runs of `chunk_gla` at `B4_T4096_H64_D128` OOMed with `53 GiB already allocated`.

Hoist the nested closure to a module-level `_launch_grid_chunked_recursive` with all state passed as explicit parameters. No closure, no per-call cycle; chunk traversal order, offsets, and per-launch kwargs are unchanged.

## Test plan

- `pytest tests/utils/test_ascend_ub_manager.py`: 34 passed (launch semantics unchanged)
- `pytest tests/ops/test_gla.py`: 30 passed on Ascend910_9382 (CANN 9.0.0, PyTorch 2.7.1)
- **Memory loop (50 `chunk_gla` fwd calls, B4_T4096_H64_D128, no `gc.collect()`): constant 1.50 GiB, previously a 6.75→19.00 GiB sawtooth**

Neutral for kernel performance — host-side launch helper, no kernel code touched. `chunk_gla` benchmark medians are within ±1% of the pre-fix numbers, and the previously OOMing shapes (B4_T4096_H64_D128 fwdbwd 267.146 ms, B8_T2048_H32_D256 fwd 147.048 ms / fwdbwd 459.219 ms) now complete.

## Breaking changes

None. Launch order, chunk boundaries, and per-launch `kernel_kwargs` are unchanged; all NPU backends using `launch_grid_chunked` behave identically except for the removed memory accumulation.

## Checklist

### DIFF
--- a/fla/utils/ascend_ub_manager.py
+++ b/fla/utils/ascend_ub_manager.py
@@ -584,6 +584,45 @@ def get_npu_properties() -> dict:
     return driver.active.utils.get_device_properties(torch.npu.current_device())
 
 
+def _launch_grid_chunked_recursive(
+    kernel,
+    grid: tuple[int, ...],
+    offset_keys: tuple[str, ...],
+    kernel_kwargs: dict,
+    quanta: tuple[int, ...],
+    budget: int,
+    extra: dict,
+    lens: list[int],
+    ax: int,
+    prod_so_far: int,
+) -> None:
+    quantum = quanta[ax]
+    rest = prod_so_far * quantum
+    for size in grid[ax + 1:]:
+        rest *= size
+    for off_q, len_q in iter_axis_launch_chunks(
+        triton.cdiv(grid[ax], quantum), rest, max_grid=budget,
+    ):
+        offset = off_q * quantum
+        lens[ax] = min(len_q * quantum, grid[ax] - offset)
+        kernel_kwargs[offset_keys[ax]] = offset
+        if ax == len(grid) - 1:
+            kernel[tuple(lens)](**kernel_kwargs, **extra)
+        else:
+            _launch_grid_chunked_recursive(
+                kernel,
+                grid,
+                offset_keys,
+                kernel_kwargs,
+                quanta,
+                budget,
+                extra,
+                lens,
+                ax + 1,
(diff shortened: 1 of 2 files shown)


fla/utils/ascend_ub_manager.py
```diff
@@ -584,6 +584,45 @@ def get_npu_properties() -> dict:
     return driver.active.utils.get_device_properties(torch.npu.current_device())
 
 
+def _launch_grid_chunked_recursive(
+    kernel,
```
zheliuyu, line 588: `tests/utils/test_ascend_ub_manager.py` never calls `launch_grid_chunked`, so the "34 passed, launch semantics unchanged" note doesn't actually lock this in.

A fake kernel that records (`grid`, `offset kwargs`) is enough: force 2D/3D chunking with a small `budget`, include the `quanta` path, and use a `weakref` sentinel in `kernel_kwargs` to show the object is collected after return without `gc.collect()`. The old nested `_recurse` would fail that. The 50-call memory loop in the PR body is good evidence, but it will regress the first time someone nests this helper again for locality.

zheliuyu (commented):
Nice first PR — the root cause is right, the hoist actually kills the per-call cycle, and the public launch API is unchanged. Happy to merge once this is covered.

fla/utils/ascend_ub_manager.py
Ginray: Added regression coverage for 2D/3D chunking, the `quanta` path, and weakref lifetime after return. Thanks!

zheliuyu:
LGTM. cc @zhiyuan1i

zhiyuan1i (approved):

PR #1237, opened 2026-09-09
[Ops] Enable TMA on AMD gfx1250

CRobeck:
## Summary
IS_TMA_SUPPORTED was Nvidia-only. gfx1250 is the first AMD arch whose Triton backend lowers tl.make_tensor_descriptor onto hardware supported tensor-descriptor loads and stores (TDM in AMD terminology). Add it behind a new IS_AMD_TMA_ARCH flag and a get_device_arch() helper. gfx942 and gfx950 have no TDM HW support. The path is gated behind FLA_USE_TMA=1.

 - Add IS_TF32_SUPPORTED to the solve_tril DOT_PRECISION gate. That list keyed off IS_TMA_SUPPORTED alone, which now admits AMD, where the tf32 dot precisions do not exist. 
- Fix the config order while there: it was derived from a set, so it varied with the interpreter hash seed.

## Test plan
Added tests/ops/utils/test_tma.py covering a 2-D tile transposed, skipped unless IS_TMA_SUPPORTED. All four shapes satisfy the descriptor alignment rules, which the compiler does not check and which fault at launch when violated; three leave a ragged tail so descriptor zero-fill on load and drop on store are exercised in place of explicit masks.

Updated the CONTRIBUTING TMA guidance and the FLA_USE_TMA row in ENVs.md to match, tested on a local gfx1250 there isn't currently a fla CI runner has a TMA-capable AMD arch.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -225,12 +225,9 @@ Don't hard-wrap prose at an arbitrary short column — this covers Markdown file
   any divisibility you rely on. Do not use `tl.make_block_ptr` / `tl.advance`:
   deprecated upstream and removed in triton main. (`backends/triton_ascend/` is
   exempt — triton-ascend still requires block pointers.)
-- `tl.make_tensor_descriptor` (TMA) is an opt-in optimization for hot-path
-  tiles on Hopper and newer, not a default substitute for block access. It
-  requires 16-byte-aligned bases and stride multiples, a stride-1 innermost
-  dim, no transposed blocks, and a registered allocator for device-side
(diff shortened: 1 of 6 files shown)


CRobeck:
Can also add a benchmark example to compare TMA vs standard masked load perf if worthwhile.

CRobeck:
@zhiyuan1i is this OK to merge?

zhiyuan1i (approved):
Overall, LGTM.   But we lack AMD GPUs to verify those changes：（

PR #1239, opened 2026-09-10
[Fix] Fold chunk state launch grids to fit the 65535-block limit

xy200303:
## Summary

Several chunk-state kernels launch a `(NK, NV, N*H)` grid. `N` is the number of packed
documents in the varlen path (and the batch size in the dense path), so a packed batch with
more than `65535 / H` documents puts `N*H` past CUDA's 65535 per-axis grid cap and the kernel
launch fails outright:

```
RuntimeError: Triton Error [CUDA]: invalid argument
  File ".../triton/backends/nvidia/driver.py", line 708, in __call__
    self.launch(gridX, gridY, gridZ, stream, function, ...)
```

`chunk_delta_h.py` (GDN's state kernels) was already fixed this way in #1077; the same kernels
in the GLA/simple_gla path, IPL-R, DPL-R and RWKV6 were missed. This PR folds the whole grid
into dim 0 for them, in the same style as #1077/#1174, and keeps the original block
enumeration order (`i_k` fastest, then `i_v`, then `i_nh`).

Reproduced on RTX 5090 D (sm_120), triton 3.4.0, with 2050 documents × 16 tokens, `H=32`,
i.e. `N*H = 65600 > 65535`. Before this PR: `chunk_simple_gla`, `chunk_gla`,
`chunk_linear_attn`, `chunk_rwkv6`, `chunk_iplr_delta_rule` and `chunk_dplr_delta_rule` all
fail with `invalid argument`; after: all six pass. `chunk_gated_delta_rule` passes both
before and after, which is the control — it is the kernel family #1077 already fixed.

The affected kernels: `common/chunk_h.py` (`chunk_fwd_h`, `chunk_bwd_dh`),
`utils/cumsum.py` (`chunk_global_cumsum_vector`), `generalized_delta_rule/iplr/chunk.py`,
`generalized_delta_rule/dplr/chunk_h_fwd.py`, `chunk_h_bwd.py`, `rwkv6/chunk.py`.

## Test plan

- New test: `tests/ops/test_simple_gla.py::test_chunk_varlen_many_documents` — 2050 documents
  of 16 tokens, `H=32` so `N*H = 65600`; compares one launch against two half-launches
  concatenated. Run on RTX 5090 D: `1 passed`, `o diff 0.000000`, `ht diff 0.000000`.
- Reproduced the failure (and the fix) with a small driver that calls each affected public op
  at `N*H = 65600` in a fresh process per op:
  `chunk_simple_gla`, `chunk_gla`, `chunk_linear_attn`, `chunk_rwkv6`,
  `chunk_iplr_delta_rule`, `chunk_dplr_delta_rule` → all `OK` after the change.
- Dependent tests run on RTX 5090 D (torch 2.8.0+cu128, triton 3.4.0):
  - `tests/ops/utils/test_cumsum.py` — 32 passed
  - `tests/ops/test_iplr_delta.py` — 14 passed
  - `tests/ops/test_rwkv6.py` — 13 passed
  - `tests/ops/test_linear_attn.py` — 30 passed
- Varlen / CP / model tests: not run (no matching coverage on this machine).

- Hardware: RTX 5090 D (sm_120), 32 GB, torch 2.8.0+cu128, triton 3.4.0
- Workload: `chunk_simple_gla` / `chunk_gla` / `chunk_linear_attn` / `chunk_iplr_delta_rule`,
  bf16, `(1, 4096, 8, 64)` varlen
- Before: n/a (baseline captured on the unmodified tree)
- After: same
- Conclusion: **neutral**. This is a launch-geometry change: only the block enumeration order
  changes, not the work each block does. With the affected kernel's config space pinned to a
  single config, outputs before and after are **bitwise identical** (verified with
  `torch.equal` on `o` and the final state). `chunk_iplr_delta_rule` diverges to NaN on the
  random inputs used for that comparison; the NaN positions are identical before and after and
  the finite elements are bitwise equal.

## Breaking changes

- None. No public signature, dtype, or numerical behavior changes; the fix only makes
  previously-failing launches work.

## Notes for reviewers

- `chunk_gsa` is **not** included. After folding its state grid it advances past the launch
  error and then hits a separate `illegal memory access`. That looks like an independent
  out-of-bounds issue, so it is deliberately left out to keep this PR's causality clean.
- Consumer Blackwell (sm_120) is not in the CI matrix (the NVIDIA runners are H100), so the
  reproduction above is a local one; the commands are in the PR discussion if anyone wants to
  re-run them.
- Out of scope, same class but a different axis: kernels that put `B*H` in grid dims 1/2 still
  overflow for large dense batches (e.g. `chunk_gated_delta_rule` at `B=2048, H=32`). #1174
  deliberately folded only the chunk-count axis and kept `B*H` in dim 1, so I left that
  decision alone rather than re-litigating it here.

## Checklist

### DIFF
--- a/fla/ops/common/chunk_h.py
+++ b/fla/ops/common/chunk_h.py
@@ -63,7 +63,13 @@ def chunk_fwd_kernel_h(
     IS_VARLEN: tl.constexpr,
     STATE_V_FIRST: tl.constexpr,
 ):
-    i_k, i_v, i_nh = tl.program_id(0), tl.program_id(1), tl.program_id(2).to(tl.int64)
+    # grid dims 1 and 2 are capped at 65535 blocks, so fold the whole grid into dim 0
+    pid = tl.program_id(0)
+    NK = tl.cdiv(K, BK)
+    NV = tl.cdiv(V, BV)
+    i_k = pid % NK
+    i_v = (pid // NK) % NV
+    i_nh = (pid // (NK * NV)).to(tl.int64)
     i_n, i_h = i_nh // H, i_nh % H
     if IS_VARLEN:
         bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
@@ -210,7 +216,13 @@ def chunk_bwd_kernel_dh(
     IS_VARLEN: tl.constexpr,
     STATE_V_FIRST: tl.constexpr,
 ):
-    i_k, i_v, i_nh = tl.program_id(0), tl.program_id(1), tl.program_id(2).to(tl.int64)
+    # grid dims 1 and 2 are capped at 65535 blocks, so fold the whole grid into dim 0
+    pid = tl.program_id(0)
+    NK = tl.cdiv(K, BK)
+    NV = tl.cdiv(V, BV)
+    i_k = pid % NK
+    i_v = (pid // NK) % NV
+    i_nh = (pid // (NK * NV)).to(tl.int64)
     i_n, i_hq = i_nh // HQ, i_nh % HQ
     i_h = i_hq // NG
     if IS_VARLEN:
@@ -337,7 +349,7 @@ def chunk_fwd_h(
     state_shape = (V, K) if state_v_first else (K, V)
     h = k.new_empty(B, NS, H, *state_shape, dtype=k.dtype if not states_in_fp32 else torch.float)
     ht = k.new_empty(N, H, *state_shape, dtype=torch.float) if output_final_state else None
-    def grid(meta): return (triton.cdiv(K, meta['BK']), triton.cdiv(V, meta['BV']), N * H)
+    def grid(meta): return (triton.cdiv(K, meta['BK']) * triton.cdiv(V, meta['BV']) * N * H,)
     chunk_fwd_kernel_h[grid](
         k=k,
         v=v,
@@ -403,7 +415,7 @@ def chunk_bwd_dh(
     dh = k.new_empty(B, NS, HQ, *state_shape, dtype=k.dtype if not states_in_fp32 else torch.float)
     dh0 = torch.empty_like(h0, dtype=torch.float) if h0 is not None else None
 
-    def grid(meta): return (triton.cdiv(K, meta['BK']), triton.cdiv(V, meta['BV']), N * H)
+    def grid(meta): return (triton.cdiv(K, meta['BK']) * triton.cdiv(V, meta['BV']) * N * H,)
     chunk_bwd_kernel_dh[grid](
         q=q,
         g=g,
--- a/fla/ops/generalized_delta_rule/dplr/chunk_h_bwd.py
+++ b/fla/ops/generalized_delta_rule/dplr/chunk_h_bwd.py
@@ -56,7 +56,13 @@ def chunk_dplr_bwd_kernel_dhu(
     USE_INITIAL_STATE: tl.constexpr,
     IS_VARLEN: tl.constexpr,
 ):
-    i_k, i_v, i_nh = tl.program_id(0), tl.program_id(1), tl.program_id(2).to(tl.int64)
+    # grid dims 1 and 2 are capped at 65535 blocks, so fold the whole grid into dim 0
+    pid = tl.program_id(0)
+    NK = tl.cdiv(K, BK)
+    NV = tl.cdiv(V, BV)
+    i_k = pid % NK
+    i_v = (pid // NK) % NV
+    i_nh = (pid // (NK * NV)).to(tl.int64)
     i_n, i_h = i_nh // H, i_nh % H
     if IS_VARLEN:
         bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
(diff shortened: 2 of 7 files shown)


xy200303:
The failing smoke test is pre-existing on `main` and unrelated to this PR.

`tests/layers/test_based.py::test_forward[B1-T1-H2-K16-V16-dtype1-chunk]` already fails the
same way on `main` before this branch existed. The scheduled `nvidia-h100-ci` run on
`cfaac246` (the current `main` tip) reports the identical error:

```
FAILED tests/layers/test_based.py::test_forward[B1-T1-H2-K16-V16-dtype1-chunk]
  - torch.AcceleratorError: CUDA error: an illegal memory access was encountered
```

https://github.com/fla-org/flash-linear-attention/actions/runs/34407286027

The dependent-test list is run in order and stops at the first failing file, so the smoke job
never reached the tests for the files changed here (`test_simple_gla.py`, `test_gla.py`,
`test_linear_attn.py`, `test_rwkv6.py`, `test_iplr_delta.py`, `test_dplr_delta.py`,
`tests/ops/utils/test_cumsum.py` all appear later in the same list). The new
`tests/ops/test_simple_gla.py::test_chunk_varlen_many_documents` did not get a chance to run
in this job.

For reference, the changed paths were verified locally on RTX 5090 D (sm_120, torch 2.8.0+cu128,
triton 3.4.0) — the six affected ops launch cleanly at `N*H = 65600` after the change, output is
bitwise identical to before with the config pinned, and `test_cumsum` / `test_iplr_delta` /
`test_rwkv6` / `test_linear_attn` pass (89 tests, 0 failures). sm_120 is not in the CI matrix, so
that evidence is local only.

Update: the base branch has since been updated and now includes `#1241`
(`[Fix] Reject odd K/V for 16-bit dtypes in chunk_bwd_dv`), which lands on the same path this
test exercises — `K=16` expands to an odd 153-wide feature map in fp16, and the backward goes
through `chunk_bwd_dv`. CI has been re-triggered on the updated head.

xy200303:
pytest tests/ops/test_simple_gla.py tests/ops/test_linear_attn.py tests/ops/utils/test_cumsum.py -q

xy200303:
The Smoke job is red from the 2-hour job cap, not from a test failure.

`gh api .../actions/jobs/103132039378/logs` shows no `FAILED` lines at all — every test that ran
passed. The job walks the dependent-test list alphabetically and was cancelled at
`The job has exceeded the maximum execution time of 2h0m0s` right after
`tests/ops/test_rwkv6.py` (one of the files this PR changes — it passed).

Touching `fla/ops/common/chunk_h.py` and `fla/ops/utils/cumsum.py` makes
`scripts/find_dependent_tests.py` expand to ~50 GPU test files, and the alphabetical tail of that
list — `test_rwkv7.py`, `test_simple_gla.py`, `test_wall_attn.py`, `tests/ops/utils/test_cumsum.py` —
never gets a chance to run inside the 2h budget. In particular the new
`tests/ops/test_simple_gla.py::test_chunk_varlen_many_documents` did not run here.

Two things worth noting for whoever picks this up:

- The tests that did run are green, including `tests/ops/test_rwkv6.py` and
  `tests/ops/test_linear_attn.py`, both of which exercise kernels this PR touches.
- Earlier, the same file `tests/layers/test_based.py::test_forward[B1-T1-H2-K16-V16-dtype1-chunk]`
  was failing here; that failure is gone from this run's list, so it is no longer masking anything.

I could not get H100 coverage for the changed files through the `pytest <args>` comment route —
both `Issue / PR pytest command` runs I triggered came back `skipped`. If there is a preferred way
to run a targeted subset (or if you would rather this PR not touch `cumsum.py` to keep the
dependent-test set smaller), let me know and I will adjust.

xy200303:
For anyone looking at the red `Smoke (PyTorch 2.12)` check: **this run is a budget timeout, not a test failure or a hang.**

The run I inspected (job `103132039378`, run `34557126450`, 2026-09-11) sat at the job's `timeout-minutes: 120` and was cancelled. Reconstructing the log:

- `find_dependent_tests.py` selected **48 test files** for this PR's changed files.
- The job runs each file in its own pytest process. Summing the 48 invocations gives **119.6 min** — essentially the entire 120-minute budget.
- There is **no single long hang**: the largest gap between consecutive log lines is 7.1 min, and 123 gaps exceed 5 s. The budget is consumed by breadth, spread across many files (slowest single invocation ≈ 20 min).

The reason the fanout is this wide is that `fla/ops/common/chunk_h.py` is imported by essentially every chunk-state kernel, so touching it pulls in nearly the whole ops suite. That is a property of the module, not of this PR.

Two other data points that point the same way:

- The same branch's earlier smoke runs (`34468307083`, `34467553463`, 09-10) failed in 1.6 min and 6.7 min respectively — the check is not deterministically red on this change.
- Recent runs on other branches show the same squeeze: `34745550725` on `fix/l2norm-autotune-key` ran 81.2 min before failing, against the same 120-minute ceiling.

Net: the code in this PR has not been shown to fail any test; the smoke tier just no longer fits the suite it now selects. Closing this out needs a `main`-side change — raising `timeout-minutes` for the smoke tier, or narrowing what it selects for core shared modules — rather than a change in this branch. I've left the fix as-is and not pushed anything to work around the timeout.

zhiyuan1i (approved):

Issue #1233, opened 2026-09-08
[RFC] Add Graph Capture Support for FLA Prefill and Training Paths on Ascend NPUs


FightingZhen:
## Proposal

Enable graph capture and replay for the GDN, KDA, and short-convolution paths in Flash Linear Attention (FLA) on Ascend NPUs.

The initial scope covers:

- GDN/KDA chunk attention forward;
- GDN/KDA chunk attention backward;
- Short convolution forward and backward;
- Complete GatedDeltaNet and KDA layer paths;
- Single-device Ascend NPU execution.

Context Parallel graph support will be implemented and validated in a subsequent phase.

This RFC is based on [[Issue #1155](https://github.com/fla-org/flash-linear-attention/issues/1155)](https://github.com/fla-org/flash-linear-attention/issues/1155). The issue proposes a backend-neutral graph contract based on static grids, static tensor addresses, device-side metadata, and capture-safe kernels. This RFC applies that contract to the Ascend NPU backend.

The first Ascend implementation will use TorchNPU's `NPUGraph` API. FLA operators will provide graph-safe fixed-capacity execution paths, while graph creation, buffer ownership, capture, replay, and bucket selection remain the responsibility of the caller.

## Motivation

The current FLA varlen paths contain several dynamic operations:

1. Kernel grids depend on the actual number of chunks or tokens.
2. `chunk_indices` are generated dynamically from `cu_seqlens`.
3. Metadata shapes depend on the current input.
4. Input, output, and intermediate tensor addresses may change between calls.
5. A single layer may require many host-side kernel launches.

In eager mode, the CPU repeatedly performs metadata construction, grid calculation, kernel submission, memory management, and synchronization. For short prefill requests and short-sequence training, this host-side overhead can become a significant portion of end-to-end latency.

Graph capture allows the complete fixed-capacity execution sequence to be captured once and replayed many times. Subsequent executions only update static input buffers and submit the captured graph, reducing host dispatch and synchronization overhead.

## Principles

- The initial scope is GDN/KDA chunk attention and short-convolution forward/backward paths.
- Decode-specific paths, including `fused_recurrent` and convolution cache-update paths, are out of scope.
- `use_graph=False` must preserve the existing eager behavior.
- Graph capture must not change the mathematical computation, state semantics, reduction semantics, or dtype behavior.
- Graph mode changes scheduling metadata, memory lifetime, and runtime submission, not the underlying algorithm.
[...]

PR #1245, opened 2026-09-12
[NPU] Add single-device NPUGraph support for KDA

FightingZhen:
## Summary

Adds opt-in, single-device NPUGraph capture and replay support for the KDA chunk operator on Ascend NPU. Part of #1233.

This PR adapts the Ascend execution path to the existing KDA graph contract, covering forward and backward execution, gate preprocessing, state handling, and intermediate buffers. It reuses device-side, fixed-capacity chunk metadata and preserves the original GPU capacity calculation.

Ascend kernels safely handle unused chunk entries through early returns or masked accesses. Graph-specific buffers are initialized where needed, and invalid tail tokens are masked before state-gradient reduction to prevent NaN propagation.

### Capture contract

Graph execution is enabled with `use_graph=True` and currently requires flattened variable-length inputs:

```text
q/k/v/g/beta: [1, T_max, ...]
cu_seqlens:   [N_max + 1]
```

Tensor shapes and capture-buffer addresses remain fixed. Input values, sequence splits, and the active token count may change between replays through updates to the live inputs and `cu_seqlens`.

The existing chunk-capacity bound is retained:

```text
NT_capacity = ceil(T_max / chunk_size) + N_max - 1
```

Unused sequence slots use repeated offsets, and unused chunk entries use sentinel metadata. Capture does not require a layout that fills the chunk capacity.

**Dense inputs and context parallelism are explicitly rejected on Ascend in graph mode for now.**

### Benchmark and CI integration

Graph benchmarks use the existing CLI entry point:

```bash
python -m benchmarks.ops.run --graph --op chunk_kda
```

The ordinary eager benchmark path remains unchanged. Internal modules separate the graph capture/timing lifecycle (`graph_benchmark.py`) from operator-specific inputs, configurations, and validation (`graph_registry.py`), allowing additional operators to reuse the same workflow.

The runner supports forward and forward-backward measurements, eager-reference validation, comparison against another Git revision, and JSON output. Ascend CI runs the consolidated graph tests serially and includes a small graph benchmark.

## Test plan

Local validation environment:

- Device: Ascend910 (Atlas A3)
- PyTorch: 2.9.0+cpu
- torch_npu: 2.9.0.post6
- CANN: 9.1.0

Graph coverage is integrated into `tests/ops/test_kda_graph.py`; there is no separate NPU test file.

```bash
ASCEND_RT_VISIBLE_DEVICES=0 FLA_NPU_XDIST=0 \
  pytest -q tests/ops/test_kda_graph.py
# 13 passed, 16 skipped

FLA_NPU_XDIST=0 \
  pytest -q tests/benchmarks/test_graph_benchmark.py
# 15 passed
```

The skipped cases exercise the existing CUDA/HIP-specific capture path.

The graph tests compare outputs, states, and active gradients with separate eager executions. Representative coverage includes:

- FP16/BF16, chunk sizes 32/64, MHA/GVA, and asymmetric K/V dimensions.
- Gate, beta, normalization, recomputation, and initial/final-state options.
- Changing inputs and sequence splits, partial-token replay, empty sequences, and recovery after shrinking the active layout.
- Inference intermediate-state output and a larger `T=2048, K=V=128` configuration.

Additional validation during development covered shared Ascend backward-kernel paths used by GDN. The unified benchmark completed all four default configurations in both forward and forward-backward modes with numerical validation.

Ruff, `git diff --check`, CI YAML parsing, and CI path-matching checks passed.

The benchmark compares `origin/main` eager, current-branch eager, and current-branch graph execution on the same device.

```bash
ASCEND_RT_VISIBLE_DEVICES=0 FLA_NPU_XDIST=0 \
  python -m benchmarks.ops.run --graph \
    --op chunk_kda \
    --base origin/main \
    --modes fwd fwdbwd \
    --warmup 20 \
    --iterations 100
```

The default configurations cover FP16/BF16, chunk sizes 32/64, fused options, and sequence lengths 128/512, with `B=1, H=2, D=64, N=2`.

Measurement methodology:

- Synchronized operator-level latency, reported as p50/p95.
- Live-input updates included in graph replay latency.
- One-time capture cost excluded from replay latency.
- Replay results validated against eager execution before graph timing.
- Forward-only eager and graph measurements both use `no_grad`.

### Results

image

The current eager-to-graph comparison measures replay benefit. Baseline-to-current eager differences are reported separately and should not be attributed to graph replay.

NCU profiling is not applicable to Ascend NPU. Dense graph measurements are outside the current flattened variable-length graph contract.

## Breaking changes

None to the public operator API. NPUGraph execution is opt-in, and ordinary eager execution retains its existing interface.

The invalid-tail masking also applies to the shared Ascend backward kernel in eager execution. GPU kernels and the original GPU graph-capacity logic are unchanged.

### Follow-up work

This PR targets single-device KDA operator-level graph execution. Follow-up work includes:

- Layer/model and decode integration, including short-convolution support.
- Dense inputs, context parallelism, and multi-device execution.
- Concurrent graphs, multiple shape buckets, and broader production workloads.
- Additional torch/torch_npu versions and support for other operators such as GDN.

Operator-level inference coverage in this PR does not imply full model-generation graph support.

## Checklist

### DIFF
--- a/fla/ops/kda/backends/triton_ascend/chunk_bwd.py
+++ b/fla/ops/kda/backends/triton_ascend/chunk_bwd.py
@@ -77,6 +77,8 @@ def chunk_kda_bwd_kernel_dAv_npu(
     i_b, i_hv = i_bh // HV, i_bh % HV
     if IS_VARLEN:
         i_n, i_t = tl.load(chunk_indices + i_t * 2).to(tl.int32), tl.load(chunk_indices + i_t * 2 + 1).to(tl.int32)
+        if i_n < 0:
+            return
         bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
         T = (eos - bos).to(tl.int32)
     else:
@@ -126,8 +128,6 @@ def chunk_kda_bwd_dAv_npu(
     chunk_indices: torch.LongTensor | None = None,
     use_graph: bool = False,
 ) -> tuple[torch.Tensor, torch.Tensor]:
-    if use_graph:
-        raise NotImplementedError("use_graph is not supported on the Ascend NPU backend")
     B, T, HV, V = k.shape[0], k.shape[1], do.shape[2], do.shape[-1]
     BT = chunk_size
     if chunk_indices is None and cu_seqlens is not None:
@@ -135,7 +135,7 @@ def chunk_kda_bwd_dAv_npu(
     BV = _get_dAv_bv(BT, V)
     NT = triton.cdiv(T, BT) if cu_seqlens is None else len(chunk_indices)
 
-    dA = v.new_empty(B, T, HV, BT, dtype=torch.float)
+    dA = (v.new_zeros if use_graph else v.new_empty)(B, T, HV, BT, dtype=torch.float)
     dv = torch.zeros_like(do)
 
     _launch_dAv_2d_kernel(
@@ -274,8 +274,10 @@ def chunk_kda_bwd_kernel_wy_v_part_npu(
 
         if IS_VARLEN:
             i_n, i_t = tl.load(chunk_indices + i_t * 2).to(tl.int32), tl.load(chunk_indices + i_t * 2 + 1).to(tl.int32)
+            is_valid = i_n >= 0
+            i_n = tl.maximum(i_n, 0)
             bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
-            T = (eos - bos).to(tl.int32)
+            T = tl.where(is_valid, eos - bos, 0).to(tl.int32)
         else:
             bos, eos = tl.cast(i_b, tl.int64) * T, tl.cast(i_b, tl.int64) * T + T
 
@@ -373,11 +375,14 @@ def chunk_kda_bwd_kernel_wy_k_part_npu(
         i_bh = task_id % BH
         i_b, i_hv = i_bh // HV, i_bh % HV
         i_h = i_hv // (HV // H)
+        is_valid = True
 
         if IS_VARLEN:
             i_n, i_t = tl.load(chunk_indices + i_t * 2).to(tl.int32), tl.load(chunk_indices + i_t * 2 + 1).to(tl.int32)
+            is_valid = i_n >= 0
+            i_n = tl.maximum(i_n, 0)
             bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
-            T = (eos - bos).to(tl.int32)
+            T = tl.where(is_valid, eos - bos, 0).to(tl.int32)
             i_tg = tl.load(chunk_offsets + i_n).to(tl.int64) + i_t.to(tl.int64)
         else:
             i_tg = tl.cast(i_b, tl.int64) * tl.cdiv(T, BT) + i_t
@@ -395,7 +400,7 @@ def chunk_kda_bwd_kernel_wy_k_part_npu(
         dg_ptr = dg + (bos * HV + i_hv) * K
 
         o_k = i_k * BK + tl.arange(0, BK)
-        m_k = o_k < K
+        m_k = (o_k < K) & is_valid
 
         p_gn = g_ptr + (min(T, i_t * BT + BT) - 1).to(tl.int64) * HV * K + o_k
         b_gn = tl.load(p_gn, mask=m_k, other=0).to(tl.float32)
@@ -522,8 +527,10 @@ def chunk_kda_bwd_kernel_wy_dw_part_npu(
 
         if IS_VARLEN:
             i_n, i_t = tl.load(chunk_indices + i_t * 2).to(tl.int32), tl.load(chunk_indices + i_t * 2 + 1).to(tl.int32)
+            is_valid = i_n >= 0
+            i_n = tl.maximum(i_n, 0)
             bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
-            T = (eos - bos).to(tl.int32)
+            T = tl.where(is_valid, eos - bos, 0).to(tl.int32)
             i_tg = tl.load(chunk_offsets + i_n).to(tl.int64) + i_t.to(tl.int64)
         else:
             i_tg = tl.cast(i_b, tl.int64) * tl.cdiv(T, BT) + i_t
@@ -657,8 +664,10 @@ def chunk_kda_bwd_kernel_wy_dA_finalize_npu(
 
         if IS_VARLEN:
             i_n, i_t = tl.load(chunk_indices + i_t * 2).to(tl.int32), tl.load(chunk_indices + i_t * 2 + 1).to(tl.int32)
+            is_valid = i_n >= 0
+            i_n = tl.maximum(i_n, 0)
             bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
-            T = (eos - bos).to(tl.int32)
+            T = tl.where(is_valid, eos - bos, 0).to(tl.int32)
         else:
             bos, eos = tl.cast(i_b, tl.int64) * T, tl.cast(i_b, tl.int64) * T + T
 
@@ -738,8 +747,6 @@ def chunk_kda_bwd_wy_dqkg_fused_npu(
     chunk_offsets: torch.LongTensor | None = None,
     use_graph: bool = False,
 ):
-    if use_graph:
-        raise NotImplementedError("use_graph is not supported on the Ascend NPU backend")
     B, T, H, K, HV, V = *k.shape, v.shape[2], v.shape[-1]
     BT = chunk_size
     if BT % _BC != 0:
@@ -749,12 +756,12 @@ def chunk_kda_bwd_wy_dqkg_fused_npu(
         chunk_indices = prepare_chunk_indices(cu_seqlens, BT)
     NT = triton.cdiv(T, BT) if cu_seqlens is None else len(chunk_indices)
 
-    dq = g.new_empty(B, T, HV, K, dtype=torch.float)
-    dk = g.new_empty(B, T, HV, K, dtype=torch.float)
-    dv2 = torch.empty_like(v)
-    dg = torch.empty_like(g, dtype=torch.float)
-    db = torch.empty_like(beta, dtype=torch.float)
-    dA = torch.empty_like(A, dtype=torch.float)
+    dq = (g.new_zeros if use_graph else g.new_empty)(B, T, HV, K, dtype=torch.float)
+    dk = (g.new_zeros if use_graph else g.new_empty)(B, T, HV, K, dtype=torch.float)
+    dv2 = (torch.zeros_like if use_graph else torch.empty_like)(v)
+    dg = (torch.zeros_like if use_graph else torch.empty_like)(g, dtype=torch.float)
+    db = (torch.zeros_like if use_graph else torch.empty_like)(beta, dtype=torch.float)
+    dA = (torch.zeros_like if use_graph else torch.empty_like)(A, dtype=torch.float)
     dA_acc = torch.zeros(B, T, HV, BT, dtype=torch.float, device=A.device)
     db_acc = torch.zeros(B, T, HV, dtype=torch.float, device=beta.device)
 
--- a/fla/ops/utils/backends/triton_ascend/cumsum.py
+++ b/fla/ops/utils/backends/triton_ascend/cumsum.py
@@ -168,11 +168,13 @@ def chunk_local_cumsum_scalar_kernel_npu(
                 tl.load(chunk_indices + i_t * 2).to(tl.int64),
                 tl.load(chunk_indices + i_t * 2 + 1).to(tl.int64),
             )
+            is_valid = i_n >= 0
+            i_n = tl.maximum(i_n, 0)
             bos, eos = (
                 tl.load(cu_seqlens + i_n).to(tl.int64),
                 tl.load(cu_seqlens + i_n + 1).to(tl.int64),
             )
-            T = eos - bos
+            T = tl.where(is_valid, eos - bos, 0)
         else:
             bos = tl.cast(i_b, tl.int64) * T
             eos = bos + T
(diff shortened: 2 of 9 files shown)


.github/workflows/ascend-a2-benchmark-ci.yml
```diff
@@ -160,7 +160,7 @@ jobs:
           print("matplotlib savefig smoke test: ok")
           PY
 
-      - name: Benchmark chunk_gdn, chunk_kda, fused_attnres, and conv1d
+      - name: Benchmark chunk_gdn, chunk_kda, KDA graph, fused_attnres, and conv1d
```
zheliuyu, line 163: The PR body reports "11 passed", but the NPU CI job on the latest commit is red — `test_chunk_kda_npugraph_option_matrix_matches_eager[bf16-production-safe-gate-partial]` fails with a non-finite `h0` gradient ([CI job](https://github.com/fla-org/flash-linear-attention/actions/runs/34828162501/job/103925015773)). Your local runs were on an Atlas A3 while CI runs on A2, so this may come down to the hardware difference — worth investigating from that angle first.

tests/ops/test_kda_graph_npu.py
```diff
@@ -0,0 +1,480 @@
...
+from fla.ops.kda import chunk_kda
+from fla.ops.utils import get_max_num_chunks
+from fla.utils import IS_NPU, device
+
+pytestmark = pytest.mark.skipif(not IS_NPU, reason="KDA NPUGraph tests require an Ascend NPU")
+
+if IS_NPU:
+    assert hasattr(torch, "npu") and hasattr(torch.npu, "make_graphed_callables"), (
+        "The Ascend graph test job requires torch.npu.make_graphed_callables"
+    )
```
zheliuyu: Could the graph tests be folded into the existing `tests/ops/test_kda_graph.py` instead of a separate `test_kda_graph_npu.py`? Similarly, the benchmark additions could likely extend `benchmarks/ops/run.py` / `registry.py` rather than adding parallel `run_graph.py` / `graph_registry.py` entry points.

zheliuyu (commented):
Thanks for your contribution! Before the full review, two early comments:

tests/ops/test_kda_graph_npu.py
FightingZhen: Thanks for the suggestions. I've made the following updates:

1. Removed `test_kda_graph_npu.py` and merged its test cases into `tests/ops/test_kda_graph.py`.

2. Unified the benchmark entry point under `benchmarks/ops/run.py --graph` and removed `run_graph.py`.

    For the implementation, I'd propose keeping `graph_benchmark.py` and `graph_registry.py` as internal modules, not separate entry points. Graph benchmarks require a distinct lifecycle: preparing static capture buffers and separate live inputs, capturing forward/backward graphs, validating replay results against eager execution, and measuring synchronized replay latency—including input updates but excluding capture cost. The existing eager runner uses `triton.testing.do_bench` without that capture/replay lifecycle.
    
    `graph_benchmark.py` handles this shared execution and timing flow, while `graph_registry.py` defines operator-specific inputs, configurations, and validation. This keeps the existing eager benchmark path unchanged and lets future operators such as GDN reuse the graph benchmark flow through an additional registration.

.github/workflows/ascend-a2-benchmark-ci.yml
FightingZhen: CI pass :)

fla/ops/kda/backends/triton_ascend/chunk_bwd.py
```diff
@@ -522,8 +527,10 @@ def chunk_kda_bwd_kernel_wy_dw_part_npu(
 
         if IS_VARLEN:
             i_n, i_t = tl.load(chunk_indices + i_t * 2).to(tl.int32), tl.load(chunk_indices + i_t * 2 + 1).to(tl.int32)
+            is_valid = i_n >= 0
+            i_n = tl.maximum(i_n, 0)
             bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
-            T = (eos - bos).to(tl.int32)
+            T = tl.where(is_valid, eos - bos, 0).to(tl.int32)
```
sunyi0505: Several task-loop kernels (wy_k_part / wy_dw_part / wy_v_part / wy_dA_finalize, recompute_w_u, GLA fwd_o, intra bwd) perform the following when i_n < 0:
```python
i_n, i_t = tl.load(chunk_indices + i_t * 2).to(tl.int32), tl.load(chunk_indices + i_t * 2 + 1).to(tl.int32)
is_valid = i_n >= 0
i_n = tl.maximum(i_n, 0)
bos, eos = tl.load(cu_seqlens + i_n).to(tl.int64), tl.load(cu_seqlens + i_n + 1).to(tl.int64)
T = tl.where(is_valid, eos - bos, 0).to(tl.int32)
i_tg = tl.load(chunk_offsets + i_n).to(tl.int64) + i_t.to(tl.int64)
```
The i_t of the sentinel row is 0. After i_n is clamped to 0, i_tg points to the first chunk of the first real sequence. Afterwards, tl.load is still performed on h/dh of this sequence, and writeback is prevented via boundary_check with T=0.

This is inconsistent with the convention on the GPU side: if USE_GRAPH and i_n < 0: return, as well as the convention in prepare_chunk_indices_static (negative i_n must return immediately). On Ascend, block pointers with T=0 are not always clean no-ops. If a write is mistakenly omitted, it pollutes the gradients of sequence 0 instead of the unused slot. Inside the task loop, it should be written as if not is_valid: continue, and the state of seq0 should not be reused.

In the same file, dA_finalize uses TAIL_MODE=0 without boundary_check for the dense path; varlen/graph currently enforces TAIL_MODE=1, so it is safe for now, but this falls into the same category of patterns as described above.

fla/ops/common/backends/triton_ascend/chunk_delta_h.py
```diff
@@ -1013,6 +1016,10 @@ def chunk_gated_delta_rule_bwd_kernel_dhu_blockdim64_npu(
                         m_t = (i_t * BT + tl.arange(0, BT)) < T
                         b_dv *= tl.where(m_t, b_g_ratio, 0)[:, None]
                 b_dv += tl.load(p_dv, boundary_check=(0, 1))
+                # dot padding on Ascend can retain NaNs across replays with different tails.
+                # clear invalid tokens before the dH reduction: zero weights do not cancel NaNs.
+                m_t = (i_t * BT + tl.arange(0, BT)) < T
+                b_dv = tl.where(m_t[:, None], b_dv, 0.)
```
sunyi0505: The root cause is that the tail axis is not aligned to 32 bytes, and the input already has precision issues. Therefore, this only circumvents the problem, which has been resolved at https://github.com/fla-org/flash-linear-attention/pull/1250

fla/ops/kda/chunk.py
```diff
@@ -322,7 +322,8 @@ def chunk_kda(
             - ``cp_context.pre_num_ranks_dev``/``post_num_ranks_dev``, persistent int32
               device scalars refreshed in place before each replay.
 
-            ``disable_recompute=True`` is not supported. Default: ``False``.
+            Ascend currently requires flattened variable-length inputs and does not support
```
sunyi0505, line 325: It is recommended to keep `disable_recompute=True` is not supported. Default: ``False``.

fla/ops/utils/backends/triton_ascend/cumsum.py
```diff
@@ -219,6 +221,8 @@ def chunk_local_cumsum_vector_kernel_npu(
     i_b, i_h = i_bh // H, i_bh % H
     if IS_VARLEN:
         i_n, i_t = tl.load(chunk_indices + i_t * 2).to(tl.int32), tl.load(chunk_indices + i_t * 2 + 1).to(tl.int32)
+        if i_n < 0:
```
sunyi0505: |    path  |  behevior  |
|:---------|:----------|
|  intra diag/sub/inter、cumsum vector、dAv | Unconditionally if i_n < 0: return |
| gate cumsum  | Only USE_GRAPH and i_n < 0 |
| wy / intra bwd / GLA o  | T=0 + clamp i_n |
| token-parallel | Look at cu_seqlens[N], no chunk sentinel required |

The actual_t judgment for token-parallel works correctly for the token grid, but the added use_graph parameter has no effect at all. The gate still requires USE_GRAPH; eager will trigger an OOB if it incorrectly gets -1, and other kernels will simply skip. It is recommended to align with GPU: uniformly use USE_GRAPH and i_n < 0, employ continue in the task loop, and avoid dummy execution with T=0.

tests/ops/test_kda_graph.py
```diff
@@ -22,12 +23,15 @@
...
-from fla.utils import assert_close, device
+from fla.utils import IS_AMD, IS_NPU, IS_NVIDIA, assert_close, device
 
 pytestmark = pytest.mark.skipif(
-    not torch.cuda.is_available() or not hasattr(torch.cuda, "CUDAGraph"),
-    reason="graph capture tests require a CUDA device with CUDAGraph support",
+    not (IS_NVIDIA or IS_AMD or IS_NPU),
+    reason="KDA graph tests require CUDA/HIP or Ascend NPU",
 )
+cuda_only = pytest.mark.skipif(IS_NPU, reason="direct CUDA capture and NCCL require a CUDA/HIP device")
```
sunyi0505, line 33: test_chunk_kda_graph_callable_* are not marked NPU-only (except the one that rejects dense/CP). Merging into test_kda_graph.py is correct, but the option matrix contains T=2048, K=V=128, D=100, disable_recompute=True. These will use torch.cuda.make_graphed_callables instead of the original CUDAGraph path. Once the full H100 pipeline runs, the timing and failure modes may change.

fla/ops/kda/chunk.py
FightingZhen: Restored the original “disable_recompute=True is not supported” note in the use_graph documentation as suggested.

fla/ops/kda/backends/triton_ascend/chunk_bwd.py
FightingZhen: Fixed. The task-loop kernels now guard the entire task body with `if is_valid`, before any sequence-dependent address calculation or memory access. Sentinel tasks no longer clamp `i_n` to zero or execute with `T=0`.

This skips only the current task, rather than returning from the kernel and dropping later valid tasks assigned to the same core. I've also added explanatory comments and a single-core replay test covering invalid tasks between valid ones.

fla/ops/common/backends/triton_ascend/chunk_delta_h.py
FightingZhen: Hi, in terms of the modifications in this file:
1. The A2 CI failure (bf16-production-safe-gate-partial: non-finite h0 gradient) is a shared backward-kernel numerics bug, not a graph-contract issue. On Ascend, padding rows of a partial chunk can retain NaNs through tl.dot/load, and zero-weight masking does not cancel them (NaN × 0 = NaN). Without clearing invalid tail rows of b_dv before the dH reduction, NaNs propagate into dh and all downstream gradients. 
2. An earlier commit on this branch dropped these guards in anticipation of #1250. Now restores them (7 lines, previously reviewed and CI-validated) so #1245 is correct and green on its own, rather than rebasing onto #1250, which is still open with unresolved review feedback and rewrites leftover handling across many kernels. Once #1250 lands, its leftover masking can supersede these guards in a follow-up.

fla/ops/utils/backends/triton_ascend/cumsum.py
FightingZhen: Aligned the chunk-indexed kernels: sentinel handling is now gated by `USE_GRAPH`. Single-task kernels return for negative sequence indices; task-loop kernels skip the entire invalid task body without terminating the loop. The clamp-to-sequence-0 / `T=0` fallback has been removed.

Its `use_graph` argument is retained for dispatch-signature compatibility, as documented in the function docstring; for variable-length inputs, both modes derive the actual token bound from `cu_seqlens[N]`.

tests/ops/test_kda_graph.py
FightingZhen: Agreed. The newly added `test_chunk_kda_graph_callable_*` tests are now NPU-only. The existing CUDA/HIP tests continue to use their original direct CUDAGraph capture path.

This keeps the new callable-based option matrix from expanding GPU CI coverage before it has been validated there.

FightingZhen:
~~This PR depends on #1250 for the underlying Ascend padding and leftover-lane fixes needed by variable-length graph workloads. I've removed the temporary workaround from this PR and will rerun validation once #1250 lands.~~

tests/ops/test_kda_graph.py
```diff
@@ -384,3 +394,381 @@ def step():
...
+                actual_grad,
+                reference.grad,
+                rtol=tolerance,
+                atol=tolerance,
+                msg=lambda m: f"{current_offsets}::{name}: {m}",
+            )
+
+        if IS_NPU and actual_t < case.T:
+            for name, tensor in zip(("o", "dq", "dk", "dv", "dg", "db"), (o, *(inputs[i].grad for i in range(5)))):
+                torch.testing.assert_close(
```
sunyi0505: nit: directly use the assert_close method in the repository here

FightingZhen: Fixed. Now `torch.testing.assert_close` are all replaced with `assert_close` in `fla.utils`

FightingZhen:
**Code update:** 
Aligns the Ascend graph buffer allocation with the mainline GPU form: same use_graph/else structure and the same static-buffer keys, via a new get_zeroed_static_buffer helper. Unlike CUDA, Ascend requires padding rows physically zeroed on every replay (dot padding can carry UB residue; NaN × 0 = NaN defeats zero-weight masks), so the helper records the zero-fill into the captured graph — semantics-equivalent to the previous new_zeros approach. Also aligns USE_GRAPH defaults with mainline and reverts the now-unneeded gdn2 callsite change. No GPU-side behavior changes.

detail modifications: https://github.com/fla-org/flash-linear-attention/pull/1245/changes/64d8d61686d86343e357697340d0c96384a5b11d

FightingZhen:
**Known duplication: graph benchmark runner vs. eager runner**

`benchmarks/ops/graph_benchmark.py` + `graph_registry.py` re-implement several pieces that already exist in `run.py` + `registry.py`:

- git label / project-root discovery (`_git_label`, `_project_root`)
- machine-info collection (`_machine_info` vs. `_get_machine_info`)
- the isolated-subprocess + temp-worktree baseline orchestration (`_bench_current` / `_bench_at_ref` / worker env guard)
- the op registry singleton (`register_graph_op` / `get_graph_op` / `list_graph_ops`)
- KDA input construction in `_build_kda_case`, overlapping with the `chunk_kda` entry in `registry.py`

This is intentional for this PR, not an oversight. The graph path needs capture/replay-specific machinery the eager runner does not have — separate capture vs. live inputs, per-replay input updates, and a numerical validation gate before any graph latency is reported. Folding that into the shared runner means refactoring infrastructure used by every registered op, which is a design change with a much larger blast radius than this PR's scope (KDA graph capture support).

Tracked in #1275.

zheliuyu:
Dedicated issue tracking has been set up for some known issues. This PR is ready for review. @zhiyuan1i

PR #1250, opened 2026-09-15
[Fix] Fix NPU (Triton-Ascend) accuracy: leftover masking and 32-byte last-dim alignment

### LINKED ISSUE #1251: [Bug] NPU triton-ascend kernels silently mismatch on unaligned K/V/T (CANN leftover DMA aliases the next row)
### Checklist

### Describe the Bug

On Ascend NPU (`triton_ascend` backends), several FLA ops produce **wrong numerics** when `K` / `V` / `T` is not tile-aligned, or on **varlen**. Compile succeeds. `D=64/128` dense cases pass. `D=60/100`, leftover `T % BT != 0`, and packed varlen fail against the Torch / CUDA reference.
This is a silent accuracy bug, not an MTE `DDR address out of range` fault.
**Root cause (CANN leftover DMA).** Triton-Ascend leftover DMA of a `make_block_ptr` tile is rounded to **32 elements**, not to the logical remainder. If the last-dim stride is not a multiple of `max(tile, 32)`, that leftover aliases the **next row** (next token / next head / next batch).
A partial last tile is worse: leftover DMA starts at logical `n` and is 32 wide, so the row must also satisfy `n + 32 <= padded`. Ceil-to-tile is not enough:
| Logical `n` | Tile | `cdiv(n, tile)*tile` | Required pad | Why |
|---|---|---|---|---|
| 128 | 128 | 128 | 128 | already aligned |
| 64 | 64 | 64 | 64 | already aligned |
| 60 | 64 | 64 | **128** | leftover at 60 is 32 wide; `60+32=92 > 64` |
| 100 | 128 | 128 | **256** | `100+32=132 > 128` |
Packed `[T, H, K]` with `K=60` is the same bug in another layout: leftover of head 0 is physically head 1 of the **same token**.
**Follow-on failures on the leftover path**
1. `boundary_check` does not reliably zero leftover lanes; they enter `tl.dot` unless we `tl.where(mask, tile, 0)`.
2. Masked / `boundary_check` stores **RMW destination lanes**. `torch.empty` NaNs leak into valid leftover-chunk tokens; leftover-T of one batch’s last chunk can RMW the next batch. Padded outputs must be `zeros`, not `empty`.
3. Gated kernels: leftover `g` loads as 0, then `exp2(0 - g_valid)` overflows to inf (easy to hit with `gate_logit_normalizer=0.01`). Mask gate diffs **before** `exp2`.
4. Leftover `tl.where` defeats `enable_ubuf_saving`. Tiles that compile on the aligned packed path (e.g. D256 `BK=256/BV=128`) UB-overflow on leftover.

On NPU, without last-dim pad / leftover mask, the existing unaligned cases fail:
```bash
# KDA chunk: D=60, D=100, leftover T, varlen
pytest tests/ops/test_kda.py -k "D60 or D100 or 500 or varlen" -q
# GLA chunk: D=60 and leftover T=63
pytest tests/ops/test_gla.py -k "D60 or T63" -q
# KDA gate D=12 (last-dim leftover on a small feature)
pytest tests/ops/test_kda.py -k "gate" -q
```

Representative shapes already in the suite:

test_kda.py: (B=2, T=512, H=3, HV=3, D=60), (T=1000, D=100), varlen D=60 with cu_seqlens=[0, 15]
test_gla.py: (B=2, T=1024, H=4, D=60), fused recurrent (T=63, D=64)
Minimal pattern (what the kernels do):
```bash
# last-dim stride = logical K, tile BK=64, K=60
p = tl.make_block_ptr(q, (T, K), (H * K, 1), (i_t * BT, i_k * BK), (BT, BK), (1, 0))
b_q = tl.load(p, boundary_check=(0, 1))
# leftover DMA [60, 92) aliases the next head / next row; those lanes enter tl.dot
```

Host pad that actually works:
```bash
from fla.utils import npu_pad
# npu_pad(60, 64) == 128, not 64
# npu_pad(100, 128) == 256, not 128
KP = npu_pad(K, BK)
```

### Expected Behavior

NPU results should match the Torch / CUDA reference on unaligned K/V/T and varlen to the same tolerances as D=64/128. Public API shapes stay logical (K, V); padded workspace is internal.

npu_pad(n, tile): ceil to align = max(tile, 32), then +align if n % 32 != 0 and n + 32 > padded.
MASK_LEFTOVER zeros leftover lanes; aligned K/V/T with no varlen skips those copies and keeps large Cube tiles.
Padded leftover outputs are zeros, not empty.
Gate diffs are masked before exp2.
Leftover path does not reuse aligned packed BK/BV when that overflows UB.

### Environment Information

Torch: 2.9.0 (torch_npu==2.9.0.post6)
Triton: triton-ascend 3.2.2
Python: 3.12
Platform: Atlas 800T A3(X86), CANN 9.1.0

sunyi0505:
## Summary

Fixes #1251 

On CANN / Triton-Ascend, leftover DMA for blocked tiles is rounded up to **32 bytes**. When the **last dimension’s byte stride is not a multiple of 32**, that extra transfer can alias into the next row and corrupt results. This change addresses that in two layers:

1. **Contract**: Document and enforce **32-byte last-dim alignment**; unsupported K/V/D shapes are rejected in backend verifiers with a pointer to [#1250](https://github.com/fla-org/flash-linear-attention/issues/1250).
2. **Kernels**: Use **`MASK_LEFTOVER`** to zero spurious lanes for **partial T tiles** and **varlen** paths. Last-dim K/V tails are handled with honest tensor shapes plus **`boundary_check`**, without gating on `K % BK` (which would force extra copies and shrink Cube tiles for valid shapes like D=48/96).

Adds `fla/utils/_ascend_align.py` (`npu_verify_last_dims`, `npu_require_last_dims`, `npu_verify_kv`, `npu_leftover_mask`, etc.) and wires masking, tile selection, and alignment checks through `triton_ascend` backends for common chunk ops, KDA, GLA, GDN2, attnres, gate, `wy_fast`, and related paths.

### Background

- Here is a precision issues if it is not aligned to 32 bytes. **`dim * itemsize % 32 == 0`** (e.g. for fp16, the last dim must be a multiple of 16 elements; 48/64 are OK, 60 is not).
- `npu_leftover_mask` applies only when **`T % BT != 0`** or **`varlen=True`**. It is not used to mask last-dim K leftovers when `boundary_check` already clears them—avoiding regressions on byte-aligned head dims.

### Main changes

| Area | What changed |
|------|----------------|
| **`fla/utils/_ascend_align.py`** | Shared NPU DMA alignment helpers and verifier utilities |
| **Backend `__init__.py` files** | Replace permissive verifiers (`return True, None`) with `npu_verify_kv` and related checks |
| **`chunk_o`, `chunk_delta_h`, …** | `MASK_LEFTOVER`, `boundary_check`; keep larger Cube tiles (e.g. 128) when T is aligned and leftover copies are unnecessary |
| **KDA / GLA / GDN2 / common** | Consistent masking and alignment in fwd/bwd chunk, gate, bwd intra, etc. |
| **Tests** | `tests/ops/conftest.py` auto-skips parametrized ops cases with unaligned K/V/D on NPU; unit tests for GDN/KDA verifiers and `npu_*` contract |

## Test plan

hardware: Atlas 800T A3(X86)
FLA_NPU_XDIST=1 ASCEND_RT_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 pytest --exitfirst -n 8 --dist load tests/modules tests/ops/utils tests/ops/test_gdn_kernels.py tests/ops/test_gdn.py tests/ops/test_kda.py tests/ops/test_attnres.py tests/ops/test_solve_tril.py tests/ops/test_gla.py
aa9edfdb912726cb88d8905e1b40f870

d3e7ae1d9856af41cfb45dd7723115d6
df96be584a5aea0046f1ba8fc80ee9fb

## Breaking changes

None

## Checklist


sunyi0505:
Supplement: Although the UT shows a pass, the passing condition for UT is error ratio <= 1e-3. On the Ascend NPU, if the last dimension is not aligned to 32 bytes, the compiler will automatically perform padding. This results in more computed data than the actual valid data, and the padded data on the last dimension will contaminate the UB. If the proportion of padded data is less than 0.3%, the UT will pass. However, if you check the max diff, you will find that the error is already extremely large.

sunyi0505:
@zhiyuan1i  This pr is ready for review. Could you please review this PR? Thanks!

.agents/skills/fla-ascend-performance/references/last-dim-dma-pad.md
```diff
@@ -0,0 +1,105 @@
+# Last-dim leftover DMA padding (`NPU_ALIGN=32`)
+
+Apply when a Triton-Ascend kernel uses `make_block_ptr` (or a last-dim stride equal to logical `K`/`V`) and `K`, `V`, or `T` is not a multiple of the tile — including varlen. Wrong last-dim padding is a **silent accuracy bug**: compile succeeds, aligned shapes pass, unaligned / leftover tiles do not.
+
+Helpers (do not re-implement): `npu_pad`, `npu_unpad`, `npu_leftover_mask` in `fla/utils/_ascend_align.py`, re-exported from `fla.utils`.
```
zhiyuan1i: this list is missing npu_pad_last_dim and npu_pad_state_h

fla/utils/_ascend_align.py
```diff
@@ -0,0 +1,88 @@
...
+    K: int | None = None,
+    BK: int | None = None,
+    V: int | None = None,
+    BV: int | None = None,
+    varlen: bool = False,
+) -> bool:
+    """True if a kernel must zero leftover lanes (partial last tile / varlen)."""
+    if varlen:
+        return True
+    return any(
```
zhiyuan1i, line 55: question: when n % tile == 0 but n % 32 != 0 (say n=48, tile=16) npu_pad still pads — does the hardware alias in that case too? if so this condition should be n % max(tile, 32) != 0

fla/ops/kda/backends/triton_ascend/fused_recurrent.py
```diff
@@ -45,6 +45,14 @@ def _get_bv(K: int, V: int) -> int:
     )
 
 
+def _pad_last_dim(t: torch.Tensor | None, padded: int) -> torch.Tensor | None:
```
zhiyuan1i: this duplicates npu_pad_last_dim from fla.utils verbatim, please import it instead

zhiyuan1i (commented):
the fix itself looks right to me, only small consistency nits inline. one follow-up for a later PR: hard max-diff assertions for a few unaligned shapes would help, since FLA_CI_ENV only warns.

fla/ops/gated_delta_rule/backends/triton_ascend/wy_fast.py
```diff
@@ -627,8 +674,16 @@ def recompute_w_u_fwd_npu(
...
+    mask_leftover = npu_leftover_mask(
+        T=T, BT=BT, K=K, BK=BK, V=V, BV=BV, varlen=cu_seqlens is not None,
+    )
+    if mask_leftover:
+        BK = min(64, BK)
+        BV = min(64, BV)
+
+    KP, VP = npu_pad(K, BK), npu_pad(V, BV)
+    u = v.new_zeros(B, T, HV, VP)
+    w = k.new_zeros(B, T, HV, KP)
```
zhiyuan1i: unconditional new_zeros costs a full memset on the aligned path; please keep new_empty unless leftover/pad applies (same for dk/dv below), like the kda counterpart

fla/ops/kda/backends/triton_ascend/gate.py
```diff
@@ -393,9 +409,12 @@ def kda_gate_fwd_npu(
     H, K = g.shape[-2:]
     T = g.numel() // (H * K)
     BT = _get_gate_fwd_bt(T, K)
-    yg = torch.empty_like(g, dtype=output_dtype)
+    DP = npu_pad(K, triton.next_power_of_2(K))
+    g_pad = npu_pad_last_dim(g, DP)
+    dt_bias = _pad_gate_bias(dt_bias, H, K, DP)
+    yg = g_pad.new_zeros(*g_pad.shape[:-1], DP, dtype=output_dtype)
```
zhiyuan1i: same here and for dg/o below: please stay on new_empty when the dims are already aligned

fla/ops/common/backends/triton_ascend/chunk_o.py
```diff
@@ -1402,18 +1599,39 @@ def chunk_bwd_dqkwg_npu(
...
+        hdh_bk = hdh_bv = None
+    KS, VS = npu_pad(K, state_bk), npu_pad(V, state_bv)
+    h = npu_pad_state_h(h, K, V, KS, VS, state_v_first)
+    dh = npu_pad_state_h(dh, K, V, KS, VS, state_v_first)
+    mask_leftover = npu_leftover_mask(
+        T=T, BT=BT, K=K, BK=BK, V=V, BV=BV, varlen=cu_seqlens is not None,
+    )
+    if not use_full:
+        dq_f32 = torch.zeros(B, T, HV, KP, dtype=torch.float32, device=q.device)
+        dk_f32 = torch.zeros(B, T, HV, KP, dtype=torch.float32, device=q.device)
```
zhiyuan1i: zeros is only needed when KP != K; please keep empty on the aligned path

fla/ops/common/backends/triton_ascend/chunk_h.py
```diff
@@ -311,26 +313,39 @@ def chunk_fwd_h_npu(
...
     BK, BV = _chunk_h_tile_size(K, V)
+    KS, VS = npu_pad(K, BK), npu_pad(V, BV)
+    state_pad = (VS, KS) if state_v_first else (KS, VS)
+    zero_ws = KS != K or VS != V or T % BT != 0
+    alloc = k.new_zeros if zero_ws else k.new_empty
+    # last-dim pad: leftover DMA of BK/BV cannot alias the next row
+    h = alloc(B, NS, H, *state_pad, dtype=torch.float if states_in_fp32 else k.dtype)
+    ht_pad = alloc(N, H, *state_pad, dtype=torch.float) if output_final_state else None
+    if h0 is not None:
+        h0_pad = k.new_zeros(h0.shape[0], H, *state_pad, dtype=h0.dtype)
```
zhiyuan1i: please use npu_pad_state_h here (and for dht in bwd), it already has the aligned no-op fast path

fla/ops/common/backends/triton_ascend/chunk_o.py
```diff
@@ -1012,6 +1177,11 @@ def chunk_bwd_kernel_dqkwg_full_npu(
             p_do = tl.make_block_ptr(do_ptr, (T_cur, V), (HV * V, 1), (i_t * BT, i_v * BV), (BT, BV), (1, 0))
             b_v = tl.load(p_v, boundary_check=(0, 1))
             b_do = tl.load(p_do, boundary_check=(0, 1))
+            if MASK_LEFTOVER:
+                o_v = i_v * BV + tl.arange(0, BV)
+                m_tv = m_t[:, None] & (o_v < V)[None, :]
+                b_v = tl.where(m_tv, b_v, 0)
+                b_do = tl.where(m_tv, b_do, 0)
```
zhiyuan1i, line 1185: the b_ds line right below is still exp2-then-mask, which the doc in this PR bans; please mask the diff before exp2 like the fwd kernel does

fla/ops/gla/backends/triton_ascend/chunk.py
```diff
@@ -378,42 +380,47 @@ def chunk_gla_fwd_kernel_o_npu(
...
             b_q = tl.load(p_q, boundary_check=(0, 1))
             b_g = tl.load(p_g, boundary_check=(0, 1)).to(tl.float32)
-            # fold scale into the operand: an elementwise op on the accumulator
-            # between the two dots forces a fixpipe round-trip through UB
-            b_qg = (b_q * exp2(b_g) * scale).to(b_q.dtype)
             b_h = tl.load(p_h, boundary_check=(0, 1))
+            if MASK_LEFTOVER:
+                b_q = tl.where(m_t[:, None], b_q, 0)
+                b_g = tl.where(m_t[:, None], b_g, 0)
+            b_qg = (b_q * exp2(b_g) * scale).to(b_q.dtype)
```
zhiyuan1i, line 409: please restore the deleted fixpipe why-comment here, the reason still holds

fla/ops/kda/backends/triton_ascend/wy_fast.py
```diff
@@ -203,34 +213,42 @@ def recompute_w_u_fwd_kda_kernel_npu(
...
             if STORE_KG:
-                o_k = i_k * BK + tl.arange(0, BK)
-                m_k = o_k < K
                 if GK_T_CONTIG:
                     b_gn = tl.load(gk_ptr + last_idx * K + o_k, mask=m_k, other=0.0).to(tl.float32)
                 else:
                     b_gn = tl.load(gk_ptr + last_idx * HV * K + o_k, mask=m_k, other=0.0).to(tl.float32)
-                b_kg = b_k * exp2(b_gn[None, :] - b_gk)
-                p_kg = tl.make_block_ptr(kg_ptr, (T, K), (HV * K, 1), (i_t * BT, i_k * BK), (BT, BK), (1, 0))
+                b_g_diff = tl.where(m_tk, b_gn[None, :] - b_gk, 0)
```
zhiyuan1i: this where runs unconditionally on the aligned hot path; please gate it behind MASK_LEFTOVER

fla/ops/gated_delta_rule/backends/triton_ascend/wy_fast.py
```diff
@@ -467,7 +495,10 @@ def prepare_wy_repr_bwd_da_gate_npu(
     )
     b_g = tl.load(p_g, boundary_check=(0,)).to(tl.float32)
     b_dA = tl.load(p_dA, boundary_check=(0, 1)).to(tl.float32)
-    b_prod = b_dA * exp2(b_g[:, None] - b_g[None, :])
+    o_t = i_t * BT + tl.arange(0, BT)
+    m_t = o_t < T
+    b_g_diff = tl.where(m_t[:, None] & m_t[None, :], b_g[:, None] - b_g[None, :], 0)
```
zhiyuan1i: same, please gate this where behind a constexpr leftover flag

zhiyuan1i (changes requested):
please address the inline nits before merge, all small.

sunyi0505:
> please address the inline nits before merge, all small.

Thanks for the review — addressed in the follow-up commit.

Helpers: documented npu_pad_last_dim / npu_pad_state_h; fused_recurrent now imports npu_pad_last_dim instead of duplicating it; chunk_h pads h0 / dht via npu_pad_state_h.
npu_leftover_mask stays n % tile != 0 (or varlen). n=48, tile=16 does alias without padding (npu_pad → 96), but that is absorbed by last-dim pad, not leftover-lane tl.where. Using n % max(tile, 32) would force leftover copies and smaller Cube tiles on complete-tile shapes; noted in the helper docstring and last-dim-dma-pad.md.
Aligned path: GDN wy_fast u/w/dk/dv, KDA gate yg/dg/o, and chunk_o dq_f32/dk_f32 keep new_empty unless leftover or last-dim pad applies.
Leftover tl.where on the hot path is now behind MASK_LEFTOVER (KDA wy_fast kg, GDN da_gate).
chunk_o bwd masks the gate diff before exp2 (full-tile and BC subtile), matching fwd.
Restored the GLA fixpipe why-comment; scale is still folded into the operand.
Local smoke: leftover T=75, gln=0.01, unaligned D=60, and aligned GDN chunk / chunk_bwd_dqkwg / KDA gate / fused recurrent.

hard max-diff assertions for a few unaligned shapes. We will discuss the plan internally for this task and then discuss the solution in issue #1251. It will be completed in the next PR.

zhiyuan1i (dismissed):

sunyi0505:
Based on the discussion in #1250, this release retains the npu_leftover_mask and corresponding kernel logic only for T leftover / varlen; fixes for gate before exp2; centralized guards such as npu_require_last_dims, with immediate errors thrown for unsupported last dims. Items removed in this release: the pad alignment mechanism for K/V/state, and the leftover-K masking strategy based on K % BK. Please review this when you have time. @zhiyuan1i

PR #1270, opened 2026-09-21
[Perf] Add Triton-Ascend backend for GDN2 fused recurrent inference

Ginray:
## Summary

Add a Triton-Ascend backend for GDN2 fused recurrent forward on Ascend NPU.

* Add `fused_recurrent_gdn2_fwd` dispatch and the Ascend implementation.
* Keep the existing Triton path as the fallback for non-NPU platforms.
* Add Ascend-specific UB-aware tiling, persistent vector-core scheduling, direct per-token row loads, V-first state layout, and int64 address arithmetic.
* Disable automatic multibuffering for the persistent kernel because its live state exceeds the available UB otherwise.
* Add `fused_recurrent_gdn2` to the benchmark runner and Ascend A2 CI.

## Test plan

```bash
pytest tests/ops/test_gdn2.py -k fused_recurrent -q
```

```text
21 passed, 42 deselected
```

Tested on Ascend910_9382, CANN 9.0.0, PyTorch 2.7.1, and Triton 3.2.0.

On Ascend, the fp32 and fp16 `chunk_gdn2` vs. `fused_recurrent_gdn2` cross-checks ran and passed.

Hardware: Ascend910_9382
Options: `use_qk_l2norm_in_kernel=True`, `output_final_state=True`

The baseline uses the original Triton implementation with backend dispatch disabled. Run the two commands sequentially on the same device:

```bash
FLA_DISABLE_BACKEND_DISPATCH=1 \
python -m benchmarks.ops.run \
  --op fused_recurrent_gdn2 \
  --no-base \
  --modes fwd
```

```bash
python -m benchmarks.ops.run \
  --op fused_recurrent_gdn2 \
  --no-base \
  --modes fwd
```

| Shape              | Before (ms) | After (ms) | Speedup |
| ------------------ | ----------: | ---------: | ------: |
| B1 T8192 H96 D128  |     107.039 |     50.178 |   2.13x |
| B2 T16384 H16 D128 |      80.184 |     50.124 |   1.60x |
| B4 T2048 H16 D128  |      20.036 |      9.411 |   2.13x |
| B4 T4096 H64 D128  |     146.904 |     68.989 |   2.13x |
| B8 T1024 H8 D64    |       3.580 |      1.959 |   1.83x |
| B8 T2048 H32 D256  |     219.241 |    153.212 |   1.43x |

Conclusion: **1.43x-2.13x speedup** across the six registered dense shapes, with a 1.85x geometric mean.

## Performance changes

* Persistent vector-core scheduling: launch one program per available vector core and process multiple independent state tiles inside the kernel.
* V-first state layout: use `[V, K]` internally for eligible large states (`K, V >= 256`) to make reductions more contiguous, then transpose back to the public layout. The isolated D256 experiment measured 1.302x plain and 1.213x with gate computation.
* UB-aware BV selection: use the calibrated `_RECUR_MEM_MULT=4.1` budget to keep the recurrent state tile compilable. This is a resource-safety measure, not an independent speedup claim.

Masked tail handling, int64 address arithmetic, dispatch, and `multibuffer=False` are compatibility or UB-safety changes rather than independent performance claims.

## Breaking changes

No public API signature changes.

The NPU backend requires contiguous `initial_state` and `out` buffers; direct calls reject non-contiguous buffers.

## Checklist


fla/ops/gdn2/backends/triton_ascend/fused_recurrent.py
```diff
@@ -0,0 +1,362 @@
...
+    if user_out is not None:
+        assert user_out.shape == v.shape
+        if not user_out.is_contiguous():
+            raise ValueError("`out` must be contiguous")
+
+    BV = _get_bv(K, V)
+    h0 = initial_state
+
+    # V-first speeds up large-state reductions; only transpose buffers allocated here.
+    internal_v_first = (
```
zheliuyu, line 287: This layout is new, and nothing in `tests/ops/test_gdn2.py` reaches it. `test_fused_recurrent` stops at `K=64, V=128`; `test_chunk_matches_fused_recurrent` is `K=V=64`. The D256 number in the PR is a benchmark only. Please add a `K=V=256`, `output_final_state=True` case (no `initial_state`) and check `o` and `ht` against `chunk_gdn2` or `naive_recurrent_gdn2`, in fp16 and bf16. The bench runner uses bf16 (`benchmarks/ops/run.py:341`), and the recurrent tests do not.

fla/ops/gdn2/backends/triton_ascend/__init__.py
```diff
@@ -104,3 +104,46 @@ def chunk_gdn2_bwd_wy_dqkg_fused_verifier(
...
+        out: torch.Tensor | None = None,
+        **kwargs,
+    ) -> tuple[bool, str | None]:
+        del scale, output_final_state, inplace_final_state, state_v_first
+        del use_qk_l2norm_in_kernel, use_gate_in_kernel, lower_bound, kwargs
+        required = {'q': q, 'k': k, 'v': v, 'g': g, 'b': b, 'w': w}
+        if any(not isinstance(t, torch.Tensor) for t in required.values()):
+            return False, 'GDN-2 Ascend recurrent requires tensor inputs'
+        optional = (A_log, dt_bias, initial_state, cu_seqlens, ssm_state_indices, num_accepted_tokens, out)
+        bad_device = [name for name, t in required.items() if t.device.type != 'npu']
```
zheliuyu, line 138: The chunk verifier in this class already rejects a bad dtype. This one only checks the device, so an unsupported dtype fails inside the kernel instead of falling back. Same allowlist is enough: fp16, bf16, fp32.

fla/ops/gdn2/backends/triton_ascend/fused_recurrent.py
```diff
@@ -0,0 +1,362 @@
...
+
+        if IS_VARLEN:
+            bos, eos = (
+                tl.load(cu_seqlens + i_n).to(tl.int64),
+                tl.load(cu_seqlens + i_n + 1).to(tl.int64),
+            )
+            T_cur = (eos - bos).to(tl.int32)
+        else:
+            bos = i_n * T
+            T_cur = T
```
zheliuyu, line 119: Suggestion, same pattern as the KDA recurrent kernel (`fla/ops/kda/backends/triton_ascend/fused_recurrent.py:122`, `:261`): `tl.cast(i_n, tl.int64)` before these multiplies. `task_id` is already int64 (`:104`). If `//` and `%` narrow `i_n` back to int32, `(i_n * HV + i_hv) * K * V` wraps at 2^31 elements and the state load/store hits the wrong batch. `HV=64, K=V=128` reaches that at `i_n >= 2048`.

fla/ops/gdn2/backends/triton_ascend/fused_recurrent.py
```diff
@@ -0,0 +1,362 @@
...
+        task_num=task_num,
+        num_core=num_core,
+        USE_QK_L2NORM_IN_KERNEL=use_qk_l2norm_in_kernel,
+        INPLACE_FINAL_STATE=inplace_final_state,
+        USE_GATE_IN_KERNEL=use_gate_in_kernel,
+        STATE_V_FIRST=kernel_v_first,
+    )
+    # Each program owns a core and processes independent state tiles in a loop.
+    # Persistent state liveness plus automatic multibuffering exceeds UB.
+    fused_recurrent_gdn2_fwd_kernel_npu[(num_core,)](**kernel_kwargs, multibuffer=False)
```
zheliuyu: Separately, `**ascend_compile_kwargs()` instead of a raw `multibuffer=False` (`fla/utils/_compat.py:34-42`). That helper stays empty off NPU.

zheliuyu (commented):
The recurrence, dispatch, and vector-core loop look right. One test gap should be filled before merge. The verifier note is small. The `i_n` cast is a suggestion.

fla/ops/gdn2/backends/triton_ascend/fused_recurrent.py
Ginray: Thanks. Added `K=V=256` coverage for both fp16 and bf16 with `output_final_state=True` and no initial state, so it exercises the automatic V-first path. Both `o` and `ht` are checked against `naive_recurrent_gdn2`.

fla/ops/gdn2/backends/triton_ascend/__init__.py
Ginray: Good point. Added the same fp16 / bf16 / fp32 allowlist to the recurrent verifier, including the optional floating-point tensors, and added verifier tests for the supported and unsupported dtype cases.

fla/ops/gdn2/backends/triton_ascend/fused_recurrent.py
Ginray: Thanks for the suggestion. Added an explicit `tl.int64` cast for `i_n` immediately after task decomposition, so the subsequent batch/state address arithmetic stays in int64.

fla/ops/gdn2/backends/triton_ascend/fused_recurrent.py
Ginray: Updated. The raw `multibuffer=False` has been replaced with `**ascend_compile_kwargs()` to keep the launch configuration consistent with the existing Ascend helper.

zheliuyu:
Thanks for adding the test cases.

@zhiyuan1i Looks like there are no other issues, and this PR is ready for review.

